PR #2, opened 2021-11-13
SQL-605: Add API stubs

pmeredit:


odbc/src/lib.rs
```diff
@@ -1,3 +1,6 @@
+pub mod api;
```
rychipman, line 1: I can't think of any reason for this to be `pub`
```suggestion
mod api;
```

odbc/src/api.rs
```diff
@@ -0,0 +1,1047 @@
...
+    SqlReturn, StatementAttribute, ULen, USmallInt, WChar,
+};
+
+#[no_mangle]
+pub extern "C" fn SQLAllocHandle(
+    handle_type: HandleType,
+    input_handle: Handle,
+    output_handle: *mut Handle,
+) -> SqlReturn {
+    SqlReturn::ERROR
```
rychipman, line 16:
```suggestion
    unimplemented!()
```
let's use `unimplemented!` for things that we haven't implemented yet, to differentiate from cases where we actually want to just return an error (like things we're going to intentionally not implement), with the idea being that all `unimplemented!` macro invocations have been replaced with something else by the time we close the epic

odbc/src/api.rs
```diff
@@ -0,0 +1,1047 @@
+#![allow(unused)]
+/// ^ TODO: remove that when implementations are complete
```
rychipman, line 2: I have a couple quibbles with this:
- I don't love that this is going to result in us not getting this lint at all until we've completed the entire API. Especially for things like remembering to use all the appropriate out params and length args, I think it is a useful lint for implementing a big C API like this
- Even once we _have_ completed the entire API, there will still be functions that don't use their arguments

As an alternative, I'd suggest that we prefix all of the function args with `_` in this PR. As we implement functions, we can remove the leading `_` from the params we use, and leave all others

rychipman (changes requested):
a few small comments. We also still need to do this part from the ticket description:

> We should add some sort of test that verifies that we have stubbed all the required API functions (perhaps by creating a basic DSN)

odbc/src/lib.rs
pmeredit: how would this work if it's not pub? I'm pretty sure it won't build the symbols

pmeredit: Oh right, `pub use`. I'm silly.

pmeredit:
> a few small comments. We also still need to do this part from the ticket description:
> 
> > We should add some sort of test that verifies that we have stubbed all the required API functions (perhaps by creating a basic DSN)

Creating a DSN isn't going to do that. I can use the output of `nm -g` on Linux.

rychipman:
> >  We should add some sort of test that verifies that we have stubbed all the required API functions (perhaps by creating a basic DSN)
> 
> Creating a DSN isn't going to do that. I can use the output of nm -g on Linux.

I was envisioning that this would be a functional test empirically validating that we did this by demonstrating a successful linker run. But I'm realizing now that that won't work, because the ODBC driver isn't linked, but rather dynamically loaded by the driver manager, and the only thing that gets linked into a program is the generic ODBC client libs

Given that fact, I don't think it actually makes sense to try to test this -- sorry about that!

pmeredit:
@rychipman yep! I see what you mean now. I can tell you when I did this before in the MVP I added functions as they didn't exist at runtime :(

rychipman (changes requested):
changes in here LGTM! just moving myself back to "review requested" per standup discussion, pending the addition of missing functions

rychipman (approved):
LGTM

pmeredit:
evergreen merge

PR #3, opened 2021-11-18
SQL-614: Return correct error code for unsupported functions

varsha804:
Explanations can be found in the "Functions" section of the [tech design](https://docs.google.com/document/d/1BiJZm6dZVYNlnkNh8nPHGHWOoXLWkzmiCuYtGKyqzu8/edit#)


rychipman (changes requested):
moving myself back to "changes requested" pending the SQLSTATE and GetDiagRec updates discussed in standup

pmeredit (changes requested):
Setting myself to changes requested until this is complete :)

odbc/src/util.rs
```diff
@@ -0,0 +1,51 @@
...
+      let desc = unsafe { (*(handle as *mut MongoHandle)).as_descriptor().ok_or(())? };
+      let mut desc_contents = desc.write().unwrap();
+      desc_contents.sql_states.push(sql_state.to_string());
+      desc_contents.error_messages.push(error_message.to_string());
+    }
+  };
+  Ok(())
+}
+
+pub fn set_sql_state(mut sql_state: String, output_ptr: *mut Char) -> () {
```
varsha804, line 37: Ideally, these would be macros, but the transmutation only worked properly when I made them functions.

odbc/src/api.rs
```diff
@@ -695,16 +861,108 @@ pub extern "C" fn SQLGetDiagFieldW(
...
+                    None => return SqlReturn::NO_DATA,
+                }
+            }
+            None => return SqlReturn::INVALID_HANDLE,
+        },
+    }
+    SqlReturn::SUCCESS
 }
 
 #[no_mangle]
```
varsha804, line 968: Do we also want to implement the corresponding unicode function `SQLGetDiagRecW()` (next line)?

nbagnard: I think we should focus on the W methods (unicode) first because this are the methods PowerBI uses.
As a general rule, the W and not W methods should be exactly the same (extract the code in another function?) except for the C to Rust (input and output buffers) conversions.

odbc/src/handles.rs
```diff
@@ -40,6 +48,8 @@ pub struct Env {
     // state of this Env
     pub state: EnvState,
     pub connections: HashSet<*mut MongoHandle>,
+    pub sql_states: Vec<String>,
```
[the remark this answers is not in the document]
varsha804: Good point, that seems like a more intuitive design -- I'll update the code

odbc/src/tests.rs
```diff
@@ -287,3 +294,259 @@ fn invalid_alloc() {
...
+            Ok(UNIMPLEMENTED_FUNC),
+            std::str::from_utf8(&*(sql_state as *const [u8; 5]))
+        );
+        assert_eq!(
+            Ok(ERROR_MESSAGE),
+            std::str::from_utf8(&*(message_text as *const [u8; 21]))
+        );
+        // Buffer is too small to hold the entire error message (0 < length < 21)
+        assert_eq!(
+            SqlReturn::SUCCESS,
```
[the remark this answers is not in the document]
varsha804: Fixed

odbc/src/util.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    };
+    Ok(())
+}
+
+/// set_sql_state writes the given sql state to the [`output_ptr`].
+pub fn set_sql_state(mut sql_state: String, output_ptr: *mut Char) {
+    unsafe {
+        let state = std::mem::transmute::<*mut u8, *mut Char>(sql_state.as_mut_ptr());
+        copy(state, output_ptr, 5);
```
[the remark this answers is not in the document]
varsha804: Hmm, so if the `message_text` buffer is too small, we're supposed to return `SUCCESS_WITH_INFO` and populate the buffer with the truncated error message. Should we do the same thing for sql states? Alternatively, we could just not populate the buffer and return `NO_DATA`, but the first option would provide more information to the user.

odbc/src/util.rs
```diff
@@ -0,0 +1,66 @@
...
+            let mut desc_contents = desc.write().unwrap();
+            desc_contents.sql_states.push(sql_state.to_string());
+            desc_contents.error_messages.push(error_message.to_string());
+        }
+    };
+    Ok(())
+}
+
+/// set_sql_state writes the given sql state to the [`output_ptr`].
+pub fn set_sql_state(mut sql_state: String, output_ptr: *mut Char) {
```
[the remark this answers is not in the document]
varsha804: Fixed

odbc/src/tests.rs
```diff
@@ -287,3 +294,259 @@ fn invalid_alloc() {
...
+                message_text,
+                21,
+                text_length_ptr
+            )
+        );
+        assert_eq!(
+            Ok(UNIMPLEMENTED_FUNC),
+            std::str::from_utf8(&*(sql_state as *const [u8; 5]))
+        );
+        assert_eq!(
```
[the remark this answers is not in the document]
varsha804: Done

odbc/src/api.rs
varsha804: I see -- I'll move this code into `SQLGetDiagRecW` then!

odbc/src/util.rs
```diff
@@ -0,0 +1,66 @@
...
+    mut error_message: String,
+    output_ptr: *mut Char,
+    buffer_len: usize,
+    text_length_ptr: *mut SmallInt,
+) {
+    unsafe {
+        let msg = std::mem::transmute::<*mut u8, *mut Char>(error_message.as_mut_ptr());
+        let num_chars = min(error_message.len(), buffer_len);
+        *text_length_ptr = num_chars as SmallInt;
+        copy(msg, output_ptr, num_chars);
```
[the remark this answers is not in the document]
varsha804: Oh, that's right. I'll look for a function that can copy unicode characters

odbc/src/util.rs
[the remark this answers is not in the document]
nbagnard: I checked the function again. The thing here is that we don't know the size of the buffer, we only get a pointer to it. I can see the copy function overriding memory which is right after the buffer if it too small, but I don't think we can avoid that.

odbc/src/util.rs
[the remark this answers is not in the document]
varsha804: Oh, that's right. @rychipman do we need to support all UTF-16 characters (looks like functions suffixed with `W` use UTF-16 while the other functions use UTF-8)?

odbc/src/util.rs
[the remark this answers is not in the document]
varsha804: We could try transmuting the pointer into a buffer of fixed length in order to assert its size before calling `copy()`?

odbc/src/util.rs
```diff
@@ -0,0 +1,66 @@
...
+        let state = std::mem::transmute::<*mut u8, *mut Char>(sql_state.as_mut_ptr());
+        copy(state, output_ptr, 5);
+    }
+}
+
+/// set_error_message writes [`error_message`] to the [`output_ptr`]. [`buffer_len`] is the
+/// length of the [`output_ptr`] buffer in characters; the error message should be truncated
+/// if it is longer than the buffer length. The number of characters written to [`output_ptr`]
+/// should be stored in [`text_length_ptr`].
+pub fn set_error_message(
```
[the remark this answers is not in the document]
varsha804: Done

odbc/src/api.rs
pmeredit: @nbagnard yes, in fact, I would say the extracted code could go in the core API while the unsafe ODBC part will convert the char/wchar to rust string and vice versa where necessary.

odbc/src/util.rs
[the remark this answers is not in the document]
pmeredit: Yea, there's no way we can check this. It's on the user to send a large enough buffer.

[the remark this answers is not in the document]
varsha804: Yeah, I wasn't able to find anything that could help us here

odbc/src/api.rs
```diff
@@ -709,16 +895,100 @@ pub extern "C" fn SQLGetDiagRec(
...
+    text_length_ptr: *mut SmallInt,
+) -> SqlReturn {
+    if rec_number < 1 || buffer_length < 0 {
+        return SqlReturn::ERROR;
+    }
+    let mongo_handle = handle as *mut MongoHandle;
+    // Make the record number zero-indexed
+    let rec_number = (rec_number - 1) as usize;
+    match handle_type {
+        HandleType::Env => match unsafe { (*mongo_handle).as_env() } {
```
pmeredit, line 914: nit: Each one of these match branches has quite a bit of code. I think it would be worthwhile to introduce helper functions for each handle type.

pmeredit (approved):
LGTM, feel free to address my nit or not.

odbc/src/util.rs
```diff
@@ -0,0 +1,78 @@
+use crate::handles::{MongoHandle, ODBCError};
+use odbc_sys::{Char, Handle, HandleType, SmallInt, SqlReturn};
+use std::{cmp::min, ptr::copy};
+
+/// set_handle_state writes the error code [`sql_state`] to the field `sql_state`
+/// in [`handle`].
+pub fn set_handle_state(
```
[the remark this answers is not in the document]
varsha804: Fair point! I'll change its name to `add_diagnostic_info`

odbc/src/util.rs
```diff
@@ -0,0 +1,81 @@
...
+    text_length_ptr: *mut SmallInt,
+) -> SqlReturn {
+    unsafe {
+        // Check if the entire error message plus a null terminator can fit in the buffer;
+        // we should truncate the error message if it's too long.
+        let num_chars = min(error_message.len() + 1, buffer_len);
+        let msg = &mut error_message[..num_chars - 1].to_string();
+        msg.push('\0');
+        *text_length_ptr = num_chars as SmallInt;
+        cp_str_to_buffer(msg.clone(), output_ptr, buffer_len);
```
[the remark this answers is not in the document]
varsha804: Yeah, I'll change the type of `cp_str_to_buffer`'s string parameter to `&String`

odbc/src/api.rs
```diff
@@ -709,16 +895,87 @@ pub extern "C" fn SQLGetDiagRec(
 
 #[no_mangle]
 pub extern "C" fn SQLGetDiagRecW(
```
nbagnard, line 897: I've closed the comment where I was mentioning this. The error message returned should follow ODBC specific syntax. See https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/diagnostic-messages for more details on the format.
Right now, I don't see where _[ vendor-identifier ][ ODBC-component-identifier ]_ is set. What do you think makes more sense? _[ vendor-identifier ]_ should be the same everywhere and be something like _[MongDB Inc.]_ or _[MongoDB]_. For _[ODBC-component-identifier]_, it could be part of the ODBCError struct, and be something like _[api],_ _[core]_, _[driver]_ (driver here meaning the error came from the Rust driver we'll be using in the core api) for example. But it should still needs to be provided and added to the error message I think because I don't see how it can be set by the DM. Did you test this out as an integration tests to see? If not, maybe we can create a separate ticket to investigate/address this after this PR is merged? WDYT?

nbagnard:
Aside from making really, really sure that the buffer content and reported text_length is correct, everything else is looking great! 👍 Almost there :)

odbc/src/util.rs
```diff
@@ -0,0 +1,81 @@
...
+pub fn set_error_message(
+    error_message: String,
+    output_ptr: *mut WChar,
+    buffer_len: usize,
+    text_length_ptr: *mut SmallInt,
+) -> SqlReturn {
+    unsafe {
+        // Check if the entire error message plus a null terminator can fit in the buffer;
+        // we should truncate the error message if it's too long.
+        let num_chars = min(error_message.len() + 1, buffer_len);
```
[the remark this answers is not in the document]
varsha804: Good catch. I've fixed some of the buffer logic and have added this as a test case!

odbc/src/tests.rs
```diff
@@ -287,3 +296,268 @@ fn invalid_alloc() {
...
+                text_length_ptr
+            )
+        );
+        // Native error pointer
+        assert_eq!(3, *native_err_ptr);
+    }
+}
+
+#[test]
+fn conn_diag_rec() {
```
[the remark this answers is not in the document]
varsha804: Done

odbc/src/api.rs
```diff
@@ -104,7 +108,17 @@ pub extern "C" fn SQLBindParameter(
...
+    match add_diag_info(
+        HandleType::Stmt,
+        hstmt as *mut _,
+        UNIMPLEMENTED_FUNC.to_string(),
+        "SQLBindParameter is unimplemented".to_string(),
+        0,
+        "api".to_string(),
+    ) {
+        Ok(_) => SqlReturn::ERROR,
+        Err(_) => SqlReturn::INVALID_HANDLE,
```
rychipman, line 120: instead of having every function do this mapping itself, would it instead make sense for `ODBCError` to have a `return_code` method that returns the appropriate return code for the error?

odbc/src/util.rs
```diff
@@ -0,0 +1,80 @@
...
+            let desc = unsafe { (*(handle as *mut MongoHandle)).as_descriptor().ok_or(())? };
+            let mut desc_contents = desc.write().unwrap();
+            desc_contents.errors.push(error);
+        }
+    };
+    Ok(())
+}
+
+/// set_sql_state writes the given sql state to the [`output_ptr`].
+pub fn set_sql_state(mut sql_state: String, output_ptr: *mut WChar) {
```
rychipman, line 47: nit: I don't think there's any need for this function to take ownership of `sql_string`, especially because you're just cloning the arg at the call site! Even better, this function should probably accept some sort of `SQLState` enum instead of just any string

odbc/src/handles.rs
```diff
@@ -143,6 +186,19 @@ impl Statement {
...
+pub struct Descriptor {
+    pub errors: Vec<ODBCError>,
+}
+
+impl Descriptor {
+    #[allow(dead_code)]
+    pub fn default() -> Descriptor {
+        Self { errors: vec![] }
+    }
+}
```
rychipman, line 204:
```suggestion
```
We should derive the `Default` trait here instead of implementing our own `default()` func

odbc/src/handles.rs
```diff
@@ -1,12 +1,15 @@
...
-use odbc_sys::Integer;
+use odbc_sys::{Integer, SmallInt, SqlReturn, WChar};
 
 #[derive(Debug)]
 pub enum MongoHandle {
     Env(RwLock<Env>),
     Connection(RwLock<Connection>),
     Statement(RwLock<Statement>),
+    #[allow(dead_code)]
+    Descriptor(RwLock<Descriptor>),
```
rychipman, line 12: what's the reason for adding this if it's just dead code? I was under the impression that we were intentionally omitting the descriptor handle type until we had evidence that it was needed (though perhaps my understanding is stale)

odbc/src/api.rs
rychipman: yeah, similar to what I mentioned in a comment elsewhere, I think this could just be a method on `MongoHandle`

rychipman (changes requested):
I still haven't gotten to review the tests in detail, but wanted to put my comments thus far up for review. Most of them are focused on Rust API design/factoring considerations.

nbagnard (changes requested):

odbc/src/api.rs
```diff
@@ -9,6 +10,9 @@ use odbc_sys::{
 };
 use std::sync::RwLock;
 
+pub const UNIMPLEMENTED_FUNC: &str = "HYC00";
+pub const VENDOR_IDENTIFIER: &str = "MongoDB";
```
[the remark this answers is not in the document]
varsha804: Removed this as part of the recent changes to error message strings!

odbc/src/handles.rs
```diff
@@ -30,6 +33,41 @@ impl MongoHandle {
...
+    }
+}
+
+#[derive(Debug)]
+pub struct ODBCError {
+    pub sql_state: String,
+    pub error_message: String,
+    pub native_err_code: i32,
+    pub component: String,
+}
```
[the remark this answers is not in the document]
varsha804: Done

odbc/src/util.rs
```diff
@@ -0,0 +1,80 @@
+use crate::handles::{MongoHandle, ODBCError};
+use odbc_sys::{Handle, HandleType, SmallInt, SqlReturn, WChar};
+use std::{cmp::min, ptr::copy};
+
+/// add_diag_info creates a new `ODBCError` object and appends it to the
+/// given handle's `errors` field.
+pub fn add_diag_info(
```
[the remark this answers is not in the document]
varsha804: Done

odbc/src/util.rs
varsha804: Done

odbc/src/handles.rs
varsha804: Done

odbc/src/util.rs
```diff
@@ -0,0 +1,80 @@
...
+) -> SqlReturn {
+    unsafe {
+        // Check if the entire error message plus a null terminator can fit in the buffer;
+        // we should truncate the error message if it's too long.
+        let msg_u16 = error_message.encode_utf16().collect::<Vec<u16>>();
+        let num_chars = min(msg_u16.len() + 1, buffer_len);
+        let mut msg = msg_u16[..num_chars - 1].to_vec();
+        msg.push('\u{0}' as u16);
+        copy(msg.as_ptr(), output_ptr, num_chars);
+        *text_length_ptr = num_chars as SmallInt;
```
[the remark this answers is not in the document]
varsha804: Done

odbc/src/handles.rs
varsha804: Some unimplemented functions, like `SQLSetDescField`,  specifically take descriptor handles as input, so we need a `Descriptor` variant in order to record errors for those functions

odbc/src/api.rs
varsha804: I actually changed the return type of `add_diag_info`, so now we can just return `SqlReturn::ERROR` in lieu of a match statement. The [docs](https://docs.microsoft.com/en-us/sql/odbc/reference/the-driver-manager?view=sql-server-ver15) say that the driver manager will check for null pointers, so the driver shouldn't need to return `SqlReturn::INVALID_HANDLE`.

odbc/src/api.rs
varsha804: I actually made those changes in the most recent commit -- I believe I've incorporated all of the formatting requirements. Do you think we still need integration tests?

odbc/src/handles.rs
```diff
@@ -1,12 +1,20 @@
...
+use crate::util::*;
+use odbc_sys::{Integer, SmallInt, SqlReturn, WChar};
+use std::{
+    collections::HashSet,
+    fmt::{Display, Formatter},
+    sync::RwLock,
+};
 
-use odbc_sys::Integer;
+pub const VENDOR_IDENTIFIER: &str = "MongoDB";
```
rychipman, line 9: A few questions:
- It looks like this is currently only used in one place. Do we plan to use it elsewhere in the future? If not, it might make sense to just inline it there
- Is there any need for this to be `pub`?

rychipman (changes requested):
haven't taken a close look at the tests yet, but wanted to get this feedback up before the weekend!

odbc/src/tests.rs
```diff
@@ -1,10 +1,16 @@
...
+    handles::{
+        Connection, ConnectionState, Descriptor, Env, EnvState, MongoHandle, ODBCError, Statement,
+        StatementState,
+    },
+    SQLAllocHandle, SQLFreeHandle, SQLGetDiagRecW,
 };
 use odbc_sys::{Handle, HandleType, SqlReturn};
 use std::sync::RwLock;
 
+const ERROR_MESSAGE_NULL: &str = "[MongoDB][API]SQLDrivers\0";
```
nbagnard, line 11: What does the suffix NULL mean for this 2 const variables?

odbc/src/api.rs
```diff
@@ -243,46 +267,64 @@ pub extern "C" fn SQLColumnsW(
...
+        SqlReturn::ERROR
+    }
 }
 
 #[no_mangle]
-pub extern "C" fn SQLCopyDesc(_source_desc_handle: HDesc, _target_desc_handle: HDesc) -> SqlReturn {
-    unimplemented!()
+pub extern "C" fn SQLCopyDesc(source_desc_handle: HDesc, target_desc_handle: HDesc) -> SqlReturn {
+    unsafe {
+        (*(source_desc_handle as *mut MongoHandle))
```
nbagnard, line 317: I don't think that the error should go in both handles.
From the spec, it should only be added to the TargetDescHandle. See [here](https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlcopydesc-function?view=sql-server-ver15#diagnostics)

> When SQLCopyDesc returns SQL_ERROR or SQL_SUCCESS_WITH_INFO, an associated SQLSTATE value can be obtained by calling SQLGetDiagRec with a HandleType of SQL_HANDLE_DESC and a Handle of TargetDescHandle. If an invalid SourceDescHandle was passed in the call, SQL_INVALID_HANDLE will be returned but no SQLSTATE will be returned.

odbc/src/util.rs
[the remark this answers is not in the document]
nbagnard: FYI
Looks like the helper was discarded. We add it in another PR, maybe one specifically for this because we will need this "functionality" in many other cases. I liked the "text.rs" file in adl_odbc_driver and would do something similar. I'll create JIRA for this.

nbagnard (changes requested):
This is looking good, we are almost there.
As we discussed, the descriptors need to be cleared between ODBC functions calls.
Otherwise, I only have a couple of comments and a few questions.

odbc/src/handles.rs
```diff
@@ -30,6 +38,89 @@ impl MongoHandle {
...
+                let mut desc_contents = (*d).write().unwrap();
+                desc_contents.errors.push(error);
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum SQLState {
+    HYC00,
```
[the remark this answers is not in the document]
varsha804: True, this does lead to some redundancy across `enum SQLState`, `impl Display`, and `ODBCError::get_sql_state()`. I've switched to using constants, and have moved the SQL state and `ODBCError` code to a new file called `errors.rs`

odbc/src/tests.rs
varsha804: "_NULL" means the string is null terminated. I previously had other constants for non-null terminated strings, which is why I introduced this suffix, but I can delete it now that I've removed those constants.

odbc/src/api.rs
```diff
@@ -394,12 +456,16 @@ pub extern "C" fn SQLDrivers(
     _drvr_attr_max: SmallInt,
     _out_drvr_attr: *mut SmallInt,
 ) -> SqlReturn {
-    unimplemented!()
+    unsafe {
+        (*(henv as *mut MongoHandle))
+            .add_diag_info(ODBCError::Unimplemented("SQLDrivers".to_string()));
+        SqlReturn::ERROR
+    }
```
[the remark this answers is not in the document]
varsha804: Good idea. I'd need to implement `From<HEnv> for &mut MongoHandle` in order to do that, and apparently need to create a type for `&mut MongoHandle` in order to call that `from` function.

Alternatively, I could write functions of the form
```
pub fn into_mongo_env(handle: HEnv) -> &mut MongoHandle {
    unsafe { (handle as *mut MongoHandle).borrow_mut() }
}
```
This is perhaps less idiomatic, but we would be able to get out of introducing a new, kind of awkward type. Not sure which option is better, but the latest commit implements the first one!

odbc/src/api.rs
```diff
@@ -243,46 +267,64 @@ pub extern "C" fn SQLColumnsW(
...
 }
 
 #[no_mangle]
-pub extern "C" fn SQLCopyDesc(_source_desc_handle: HDesc, _target_desc_handle: HDesc) -> SqlReturn {
-    unimplemented!()
+pub extern "C" fn SQLCopyDesc(source_desc_handle: HDesc, target_desc_handle: HDesc) -> SqlReturn {
+    unsafe {
+        (*(source_desc_handle as *mut MongoHandle))
+            .add_diag_info(ODBCError::Unimplemented("SQLCopyDesc".to_string()));
```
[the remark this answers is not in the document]
varsha804: Oh good catch! Not sure if we need a test though, since the `_source_desc_handle` parameter is now prefixed with an underscore, which makes it clear that only the target handle gets updated?

Update: I'm actually going to leave this function implementation as "unimplemented!()" now that I've filed [SQL-661](https://jira.mongodb.org/browse/SQL-661)

odbc/src/handles.rs
varsha804: Yeah, we'll need to reuse this in the future because it'll be used in every error message instance. I'll make it private!

odbc/src/api.rs
```diff
@@ -340,14 +380,17 @@ pub extern "C" fn SQLDescribeColW(
...
     _nullable_ptr: *mut SmallInt,
 ) -> SqlReturn {
-    unimplemented!()
+    let handle = MongoHandleRef::from(statement_handle);
+    handle.clear_diagnostics();
+    handle.add_diag_info(ODBCError::Unimplemented("SQLDescribeParam"));
+    SqlReturn::ERROR
 }
 
 #[no_mangle]
```
nbagnard, line 396: I just realized that some functions in api.rs still return unimplemented!(). Was there a reason why you skipped some functions? Or did you simply miss updating the code for them? All functions should return either SQL_ERROR or SQL_SUCCESS (or SUCCESS_WITH_INFO) for the driver to behave correctly. Keeping unimplemented!() means the driver will crash when the function is called.
Could you do another pass at it?

Now, on the same kind of topic, do you think a sql_unimplemented macro or helper function, which would take a MongoHandle and function name and return SQL_ERROR would help will the repetitive code of clearing diagnostics, add a new unimplemented error message and returning SQL_ERROR?

odbc/src/api.rs
```diff
@@ -408,39 +474,61 @@ pub extern "C" fn SQLDriversW(
...
 pub extern "C" fn SQLEndTran(
     _handle_type: HandleType,
-    _handle: Handle,
+    handle: Handle,
     _completion_type: CompletionType,
 ) -> SqlReturn {
-    unimplemented!()
+    unsafe {
+        (*(handle as *mut MongoHandle))
```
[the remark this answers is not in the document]
nbagnard: You should still return SQL_SUCCESS though 😅

nbagnard (changes requested):
One last comment for left over unimplemented!() and you should be good to go!

odbc/src/util.rs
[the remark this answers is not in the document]
nbagnard: Created SQL-662

odbc/src/api.rs
varsha804: This ticket is just supposed to cover unsupported functions, which is why I left those functions as-is. My impression was that we would remove `unimplemented!()` from supported functions as we implement them.

> Keeping unimplemented!() means the driver will crash when the function is called

Will we need to call unimplemented functions before they get properly implemented?

Re: a helper function, I've abstracted away the redundant code via a new function called `fn unsupported_function`

odbc/src/api.rs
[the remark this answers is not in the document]
varsha804: Ryan suggested removing the implementation for this function since it's not exactly unsupported! [Link](https://github.com/mongodb/mongo-odbc-driver/pull/3/files/10fb916131533dc8ab7d3333da5fac008d102cc0#r780487623) to comment

odbc/src/api.rs
nbagnard: Got it now! It makes sense :)
And I like the `unsupported_function` helper 👍

nbagnard (approved):
LGTM! Thank you Varsha for the ground work here! 
We now have a great foundation for error codes and all unsupported functions are taken care of, it's great 🥳

odbc/src/tests.rs
```diff
@@ -287,3 +291,163 @@ fn invalid_alloc() {
         );
     }
 }
+
```
rychipman, line 294: can we put all of these new tests into an inline `get_diag_rec` module to isolate them from the other tests in this file?

odbc/src/tests.rs
```diff
@@ -287,3 +291,163 @@ fn invalid_alloc() {
...
+fn validate_diag_rec(handle_type: HandleType, handle: *mut MongoHandle) {
```
rychipman, line 295: if this function is only used as a helper within the `simple_diag_rec` test, let's define it in that function instead of at the module level

odbc/src/tests.rs
```diff
@@ -1,10 +1,14 @@
...
+    errors::ODBCError,
     handles::{Connection, ConnectionState, Env, EnvState, MongoHandle, Statement, StatementState},
-    SQLAllocHandle, SQLFreeHandle,
+    SQLAllocHandle, SQLFreeHandle, SQLGetDiagRecW,
 };
 use odbc_sys::{Handle, HandleType, SqlReturn};
 use std::sync::RwLock;
 
+const ERROR_MESSAGE: &str = "[MongoDB][API] The feature SQLDrivers is not implemented\0";
+const UNIMPLEMENTED_FUNC: &str = "HYC00\0";
```
rychipman, line 10: it seems like these should both be local vars (or just inlined) in the `validate_diag_rec` helper function instead of being globals here

odbc/src/tests.rs
```diff
@@ -1,10 +1,14 @@
 use crate::{
+    errors::ODBCError,
```
rychipman, line 2: I think we need some integration tests as well to have complete coverage. Now that @nbagnard 's test framework PR has been merged, this is now possible!

In particular, I think we should test the following:
- Behavior of the newly implemented ODBC API functions
- Do error lists get concatenated or replaced when the same handle is used for consecutive function calls that return an error.

If you'd prefer to defer this to a separate ticket, that's fine by me as well

rychipman (changes requested):
just a few small things re: testing

odbc/src/tests.rs
varsha804: Filed [SQL-663](https://jira.mongodb.org/browse/SQL-663)

PR #4, opened 2021-11-18
SQL-608: Basic test framework for ODBC C API

nbagnard:
This is a draft PR to validate the test layout and framework.

Below is a description of the tests layout:
- tests : Test directory for general integration tests and PowerBi "replay" tests
- tests/common/mod.rs: Common test code. Things like creating a new Environment, connecting and validating results.
- tests/metadata_tests: Directory containing the yml tests definition files.
- tests/test_gen/mod.rs: Code for deserializing the yml test definition.
- tests/metadata_tests.rs: Metadata tests (sqlTables, sqlColumns, sqlGetInfo...)
- tests/power_bi_tests.rs: Tests replaying Power Bi flows to emulate testing with Power BI.

We might want to have tests generation code in our build.rs later on to be able to generate different tests for query tests.
Right now, I think having all sqlTqbles, sqlColumns, etc... tests "bundled" together in one test should be Ok.

It looks like odbc-api takes care of the clean-up, so I did not add common clean-up functions for disconnecting and freeing handles.

I used yml tests definition similar to the one for JDBC. Hopefully we can share tests.

We'll probably need something to convert String representation from the yml file to constant for things like SQL types names, etc. I did not dot anything there because we might have matching Enum or code to do that in the driver as well that we can use when the time comes.

Right now, test_sql_tables does nothing because it is marked as skipped and I added only the environment setup and clean-up in the Power Bi test.
We need a DSN setup and working connection to enable our tests.
Resultset validation is also not implemented yet.


odbc/tests/common/mod.rs
```diff
@@ -0,0 +1,113 @@
+extern crate odbc_api;
+
+use self::odbc_api::handles::StatementImpl;
+use self::odbc_api::{Connection, Cursor, CursorImpl, Environment};
+use lazy_static::lazy_static;
+use odbc_api::ResultSetMetadata;
+use std::env;
+
+#[derive(Debug, PartialEq)]
+pub struct ResultsetMetadata {
```
rychipman, line 10: In order to keep the scope of this ticket more constrained, I'd suggest that we limit this ticket just to getting a basic handwritten C API test working and automated via evergreen, and split out the YAML-specified result-set tests into a separate ticket

rychipman (changes requested):
moving myself to "requested changes" while waiting for the evergreen automation. I'll give a more thorough review at that point, but in the meantime I had one high-level comment re: scope for this ticket

odbc/tests/common/mod.rs
nbagnard: Done!
I tried to see how it would work with the yaml test definitions. I'll keep that on hand for later if needed.
I'm not too convince anyway still., because of how tests are generated in Rust.

evergreen.yml
```diff
@@ -138,11 +241,18 @@ tasks:
...
   - name: test-rust
+    depends_on:
+      - name: compile
     commands:
+      - func: "download artifact"
+      - func: "add Windows DLL path"
+      - func: "setup driver on Windows"
       - func: "install rust toolchain"
       - func: "run rust tests"
+      - func: "clean-up driver on Windows"
```
nbagnard, line 255: I did not find how to have a clean-up function/task which would run even if tests failed, but only after the tests ran, not after every task.

pmeredit: There are things called `pre` and `post`

pmeredit: https://github.com/10gen/sqlproxy/blob/master/.evg.yml#L661

pmeredit (changes requested):
Looking good so far, I would use `pre` and `post` for the cleanup so we don't need to worry about errors :)

evergreen.yml
nbagnard: I tried using pre and post, but this is not ideal because it runs before every task (clippy, etc...) and will fails for all tasks which are not test-rust unfortunately.
I went back to Evergreen wiki and I will try to use a task group, with only this task inside. It's a bit of a work-around, but if it works then I will be happy.

odbc/tests/power_bi_tests.rs
```diff
@@ -0,0 +1,80 @@
...
+    F: FnOnce() -> SqlReturn,
+{
+    let outcome: SqlReturn = func();
+    if validate_success {
+        assert_eq!(SqlReturn::SUCCESS, outcome)
+    } else if SqlReturn::ERROR == outcome {
+        panic!("ODBC call failed. Outcome is SQL_ERROR")
+        // TODO : Get the error information for the diagnostic records
+    }
+}
```
rychipman, line 19: this function seems to have some sharp edges, and doesn't seem like it will be cleanly extensible to testing other SQL return codes. Is there any reason we can't just use 
```
assert_eq!(
    SqlReturn::SUCCESS,
    SQLAllocHandle(...)
);
```
in the places where `execute_odbc_call` is currently used? It doesn't seem any less expressive/compact, and it's a good deal more explicit. You could even add convenience macros like `assert_success!` if that was a common pattern that you wanted to make more terse
```suggestion
```

rychipman (changes requested):

odbc/tests/metadata_tests.rs
```diff
@@ -0,0 +1,20 @@
+extern crate odbc_api;
+
+use crate::common::{connect, validate_rs};
+
+mod common;
+
+#[allow(dead_code)]
+//#[test]
+fn test_sql_tables_no_filter() {
```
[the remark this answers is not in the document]
nbagnard: I've change the test file to use one that just connecting. Maybe this will make more sense to everyone. I also trimmed common.rs accordingly. The headers for test_default_connection still need to be flipped until connection is implemented on the driver.

evergreen.yml
```diff
@@ -104,17 +132,92 @@ functions:
...
+        build_variants:
+          - windows-32
+          - windows-64
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        remote_file: mongosql-odbc-driver/artifacts/test/${version_id}/${build_variant}/mongoodbc.dll
+        local_file: mongosql-odbc-driver/mongoodbc.dll
+        bucket: mciuploads
+
+  "setup driver on Windows":
```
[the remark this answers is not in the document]
nbagnard: Unfortunately, because we don't have the setup routines (the UI for the window that is supposed to pop-up when doing "configure" on ODBC Admin), the command fails with `Add-OdbcDsn : The setup routines for the %s ODBC driver could not be accessed.  Please reinstall the driver.`.

pmeredit (approved):
LGTM.

I like the reg files, I always did that work manually in the shell. I think the reg file is nicer.

evergreen.yml
```diff
@@ -43,10 +43,27 @@ functions:
...
+      params:
+        shell: cmd
+        working_dir: mongosql-odbc-driver
+        script: |
+          set WIN_DRIVER_DLL_PATH=%cd%
+          set WIN_DRIVER_DLL_PATH=%WIN_DRIVER_DLL_PATH:\=\\\\\\\\%
+          echo   export DRIVER_DLL_PATH="%WIN_DRIVER_DLL_PATH%" >> expansions.yml
+    - command: expansions.update
+      params:
+        file: mongosql-odbc-driver/expansions.yml
```
rychipman, line 66: This seems like a very brittle way of getting the `DRIVER_DLL_PATH` export into the `prepare_shell` expansion, since it assumes that `expansions.yml` is formatted in a specific way. Is there any reason we can't just do this in the existing `generate expansions` function above?

odbc/tests/power_bi_tests.rs
```diff
@@ -0,0 +1,67 @@
...
+    unsafe {
+        // Verify that freeing the handle is working as expected
+        assert_eq!(
+            SqlReturn::SUCCESS,
+            SQLFreeHandle(HandleType::Env, env_handle as Handle)
+        );
+    }
+}
+
+// TODO : Add the other flows [SQL-639], [SQL-640], [SQL-641] [SQL-642]
```
rychipman, line 67:
```suggestion
```
Since these tasks are captured in JIRA, I think we can remove this TODO comment. Sometimes we will leave comments referencing JIRA tickets in our code when there's something that needs to be explained (e.g. "here's why we're skipping this test" or "here's why we're doing this weird workaround"), but since these tickets are all strictly additive, I think we can just omit the reference

evergreen.yml
```diff
@@ -104,17 +132,92 @@ functions:
...
+          reg query "HKEY_LOCAL_MACHINE\SOFTWARE\ODBC\ODBCINST.INI\ODBC Drivers" -s
+          if [ 'windows-32' == '${build_variant}' ]; then
+            reg query "HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\ODBC\ODBCINST.INI\ADL_ODBC_DRIVER" -s
+            reg query "HKEY_LOCAL_MACHINE\SOFTWARE\ODBC\ODBCINST.INI\ADL_ODBC_DRIVER" -s
+          else
+            reg query "HKEY_LOCAL_MACHINE\SOFTWARE\ODBC\ODBCINST.INI\ADL_ODBC_DRIVER" -s
+            reg query "HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\ODBC\ODBCINST.INI\ADL_ODBC_DRIVER" -s
+          fi
+          echo "-------------------------"
+        continue_on_err: true
```
rychipman, line 189: What are the situations where we'd expect this to fail but would still want to continue? I'd have expected that if driver setup failed, we'd want the task to fail

rychipman (changes requested):
looking good! just a couple tiny things left for me

evergreen.yml
nbagnard: I was doing it this way because I wanted to get the Windows path and also thought that I needed DRIVER_DLL_PATH between tasks. I found the cygpath command and also realized that DRIVER_DLL_PATH is only used in the driver setup task so it doesn't have to be here.

evergreen.yml
nbagnard: This was because if a registry entry doesn't exist, the task fails. But a couple of registry entries are expected to be missing. The idea is to make sure the setup did not create unwanted entries and/or that they are no lingering entries. I changed this to suppress errors when we expect the entry to be missing.

rychipman (approved):
LGTM! great work, and awesome to have this to build on now!

nbagnard:
evergreen merge

PR #5, opened 2021-11-19
SQL-616: Add initial support for SQLAllocHandle and SQLFreeHandle

pmeredit:
I'm sure `Descriptor` will change in the future, and we will definitely be adding more attributes, but `SQLAllocHandle` and `SQLFreeHandle` will be mostly unchanged (maybe a bit of change for `Descriptors`.


odbc/src/api.rs
```diff
@@ -7,11 +11,49 @@ use odbc_sys::{
...
+) -> SqlReturn {
+    match handle_type {
+        HandleType::Env => {
+            let env = Box::new(EnvHandle::new());
+            env.env.write().unwrap().state = EnvState::Allocated;
+            unsafe { *output_handle = Box::into_raw(env) as *mut _ }
+        }
+        HandleType::Dbc => {
+            let env = input_handle as *mut _;
+            let conn = Box::new(ConnectionHandle::new(env));
```
pmeredit, line 26: We may, here, wish to return an error if we try to allocate a connection on an unallocated environment, but we can leave that for when we have error support.

odbc/src/api.rs
```diff
@@ -7,11 +11,49 @@ use odbc_sys::{
...
+            conn.connection.write().unwrap().state =
+                ConnectionState::AllocatedEnvAllocatedConnection;
+            unsafe {
+                let conn_ptr = Box::into_raw(conn) as *mut _;
+                (*env).env.write().unwrap().connections.insert(conn_ptr);
```
pmeredit, line 31: Note these `unwraps` are safe and what is commonly done with locks. They can only be errors if another thread holding the lock panics.

pmeredit: In the MVP I foolishly did busy waiting on these, but they actually already sleep until the lock is available.

odbc/src/api.rs
```diff
@@ -7,11 +11,49 @@ use odbc_sys::{
...
-) -> SqlReturn {
-    unimplemented!()
+    handle_type: HandleType,
+    input_handle: Handle,
+    output_handle: *mut Handle,
+) -> SqlReturn {
+    match handle_type {
+        HandleType::Env => {
+            let env = Box::new(EnvHandle::new());
+            env.env.write().unwrap().state = EnvState::Allocated;
```
rychipman, line 21: what is the scenario in which we can actually get a handle in the `Unallocated` state? In the Rust code you get this when you create one with `EnvHandle::new()`, but I'm wondering whether it's possible to get to this state via the ODBC C API

odbc/src/handles.rs
```diff
@@ -0,0 +1,202 @@
...
+    collections::HashSet,
+    sync::{Mutex, RwLock},
+};
+
+use odbc_sys::Integer;
+
+#[derive(Debug)]
+pub struct EnvHandle {
+    pub env: RwLock<Env>,
+}
```
rychipman, line 11: question for all the `XHandle` structs:
Do we ever expect these structs to contain anything other than the single `RwLock` field? (I _think_ the answer is no, at least as long as we stick with our whole-handle locking strategy)

If not, I wonder whether it might make the code that uses the handles a bit cleaner to eliminate the need for the intermediate `env` field access by making the `EnvHandle` into either a type alias for `RwLock<Env>` or by making the `env` field private and providing methods directly on `EnvHandle`

These are all half-baked API ergonomics ideas, so just wanted to throw them out there -- happy to discuss further in person, or just feel free to ignore.

odbc/src/handles.rs
```diff
@@ -0,0 +1,202 @@
...
+            state: ConnectionState::AllocatedEnvUnallocatedConnection,
+            _descriptors: HashSet::new(),
+            statements: HashSet::new(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct StatementHandle {
+    pub stmt: Mutex<Statement>,
```
rychipman, line 116: I see that you called this out in the design doc, but asking here instead because this is where other discussion threads already exist:

Why does this need to be a mutex? I'm not sure I understand why a `Cursor` is different wrt what kind of locking we need

odbc/src/api.rs
```diff
@@ -431,8 +473,54 @@ pub extern "C" fn SQLForeignKeysW(
...
 #[no_mangle]
-pub extern "C" fn SQLFreeHandle(_handle_type: HandleType, _handle: Handle) -> SqlReturn {
-    unimplemented!()
+pub extern "C" fn SQLFreeHandle(handle_type: HandleType, handle: Handle) -> SqlReturn {
+    unsafe {
+        match handle_type {
+            // By making Boxes to the types and letting them go out of
+            // scope, they will be dropped.
+            HandleType::Env => {
+                let env = Box::from_raw(handle as *mut EnvHandle);
```
rychipman, line 482: IIUC, this will result in UB if we interpret the handle as the wrong type (which could happen if the caller lies about `handle_type`). Is that allowed, or do we need to explicitly return `SQL_INVALID_HANDLE`? The state transition tables seem to suggest it is the latter, but also possibly that it's handled by the DM?

odbc/src/handles.rs
```diff
@@ -0,0 +1,202 @@
...
+                connections: HashSet::new(),
+            }),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Env {
+    // attributes for this Env
+    pub _attributes: EnvAttributes,
```
rychipman, line 28: here and elsewhere in this patch, I'd suggest that we just omit things that aren't used yet, and instead add things in as we actually implement the things that use them

odbc/src/api.rs
```diff
@@ -431,8 +473,54 @@ pub extern "C" fn SQLForeignKeysW(
...
+                    .connection
+                    .write()
+                    .unwrap()
+                    .statements
+                    .remove(&handle);
+            }
+            HandleType::Desc => {
+                let desc = Box::from_raw(handle as *mut DescriptorHandle);
+                // Actually reading this value would make ASAN fail, but this
+                // is what the ODBC standard expects.
```
rychipman, line 518: can you elaborate? I don't really know what this comment means.

rychipman (changes requested):
some high-level comments, and a whole bunch of questions!

odbc/src/api.rs
pmeredit: sgtm

odbc/src/api.rs
pmeredit: A program reading this state value after `SQLFreeHandle` is reading unallocated memory. ASAN will fail if that happens.

pmeredit: I'm arguing that I'm writing a value that should never be read, but it is what is implied by the ODBC standard

odbc/src/api.rs
pmeredit: It seems we should handle it. I know how to make this work. Thanks for pointing this out.

odbc/src/handles.rs
```diff
@@ -0,0 +1,202 @@
...
+impl Default for EnvAttributes {
+    fn default() -> Self {
+        Self { odbc_ver: 3 }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum EnvState {
+    Unallocated,
+    Allocated,
```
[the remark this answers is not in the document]
pmeredit: Whoops, I have that in the Design Doc, too.

odbc/src/api.rs
pmeredit: After `SQLFreeHandle`, but as mentioned, it is an address violation  for anything to actually read that state after `SQLFreeHandle` returns!

odbc/src/handles.rs
pmeredit: Let me see how that goes. I like the alias idea.

odbc/src/handles.rs
pmeredit: Reading a cursor is a write, also. Maybe this is unnecessary, just trying to be safe as possible.

pmeredit: Actually, I think it makes sense to change to `RwLock` and we'll just have to use `write` to access the cursor.

rychipman: yeah, I think the `Cursor` API already prevents this -- you need a mutable borrow to iterate the cursor!

odbc/src/handles.rs
pmeredit: We absolutely know we'll need attributes though. I'd be fine with removing the other things that aren't attributes/the handle state names.

odbc/src/api.rs
pmeredit: Ok, this finally works! It also returns `INVALID_HANDLE` if "super handles" are not passed for `Statements` and `Connections`.

odbc/src/api.rs
```diff
@@ -7,11 +11,49 @@ use odbc_sys::{
...
+                (*conn)
+                    .connection
+                    .write()
+                    .unwrap()
+                    .statements
+                    .insert(stmt_ptr);
+                *output_handle = stmt_ptr as *mut _
+            }
+        }
+        HandleType::Desc => {
```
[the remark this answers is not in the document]
rychipman: Keeping descriptor stuff in a separate ticket SGTM. If we're going to go that route, let's do the following in this PR:
- File the jira ticket to implement alloc/free for descriptor handles
- Replace the `HandleType::Desc` branches here and in `SQLFreeHandle` with `unimplemented!("see SQL-XXX")`

odbc/src/tests.rs
```diff
@@ -0,0 +1,111 @@
...
+        assert_eq!(
+            SqlReturn::SUCCESS,
+            SQLAllocHandle(
+                HandleType::Env,
+                std::ptr::null_mut(),
+                std::mem::transmute::<*mut *mut EnvHandle, *mut Handle>(handle_ptr),
+            )
+        );
+        assert_eq!(EnvState::Allocated, (*handle).read().unwrap().state);
+        assert_eq!(
```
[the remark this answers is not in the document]
pmeredit: That won't work because it will make the address sanitizer fail.

odbc/src/api.rs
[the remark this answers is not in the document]
pmeredit: sgtm!

odbc/src/api.rs
```diff
@@ -1,17 +1,83 @@
...
+            let conn = RwLock::new(Connection::with_state(
+                input_handle,
+                ConnectionState::Allocated,
+            ));
+            // input handle cannot be NULL
+            if input_handle == std::ptr::null_mut() {
+                return Err(());
+            }
+            // input handle must be an Env
+            let env = unsafe { (*input_handle).as_env()? };
```
nbagnard, line 48: If the check of the input type was the first thing done in the block, instead of creating a new connection which will be discarded when the input handle check fails, would it save us some time by not having to allocate and free memory? This is more of a general question, so that I can have an idea of the impact of this kind of things. I generally try to fail as early as possible to avoid unnecessary work because in Java for example creating the new Connection instance would take (some minimal) time and have to be garbage collected soon after (which also take some time and if too many useless instance of Connection are created by a dumb programm, the GB could be running more cycles than expected), but I am not sure of the incidence in Rust.

odbc/src/tests.rs
```diff
@@ -0,0 +1,244 @@
...
+                .read()
+                .unwrap()
+                .statements
+                .len()
+        );
+    }
+}
+
+#[test]
+fn invalid_free() {
```
nbagnard, line 155: Not a big deal, but is it possible to test INVALID_HANDLE for HandleType::Env for the sake of test coverage? Would it work with passing it a null pointer somehow?

nbagnard (approved):

odbc/src/api.rs
pmeredit: Oh, I agree. I'll check that out. This was kind of silly.

odbc/src/tests.rs
pmeredit: null pointer would segfault. I’ll think of something

odbc/src/handles.rs
```diff
@@ -0,0 +1,148 @@
...
+
+#[derive(Debug)]
+pub enum MongoHandle {
+    Env(RwLock<Env>),
+    Connection(RwLock<Connection>),
+    Statement(RwLock<Statement>),
+}
+
+impl MongoHandle {
+    pub fn as_env(&self) -> Result<&RwLock<Env>, ()> {
```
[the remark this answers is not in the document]
pmeredit: I'm not sure, but being consistent with `bson::Bson` makes sense

odbc/src/api.rs
```diff
@@ -431,8 +497,51 @@ pub extern "C" fn SQLForeignKeysW(
...
+                if conn_contents.statements.is_empty() {
+                    conn_contents.state = ConnectionState::Connected;
+                }
+            }
+            HandleType::Desc => {
+                // TODO: SQL-618
+                unimplemented!();
+            }
+        }
+        // create the Box at the end to ensure Drop is not reordered?
```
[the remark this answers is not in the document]
pmeredit: I was worried about the drop being reordered, but I think such actually violates the semantics of rust.

odbc/src/api.rs
```diff
@@ -431,8 +497,51 @@ pub extern "C" fn SQLForeignKeysW(
...
+                    .as_connection()?
+                    .write()
+                    .unwrap();
+                conn_contents.statements.remove(&handle);
+                if conn_contents.statements.is_empty() {
+                    conn_contents.state = ConnectionState::Connected;
+                }
+            }
+            HandleType::Desc => {
+                // TODO: SQL-618
```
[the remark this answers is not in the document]
pmeredit: yep!

odbc/src/api.rs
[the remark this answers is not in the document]
pmeredit: Oh, I remember now, the problem is you'll drop the box even if `sql_free_handle` fails because of wrong handle type!

odbc/src/api.rs
```diff
@@ -1,17 +1,83 @@
...
+            }
+            // input handle must be an Env
+            let env = unsafe { (*input_handle).as_env()? };
+            let conn = RwLock::new(Connection::with_state(
+                input_handle,
+                ConnectionState::Allocated,
+            ));
+            let mut env_contents = (*env).write().unwrap();
+            let mh = Box::new(MongoHandle::Connection(conn));
+            let mh_ptr = Box::into_raw(mh) as *mut _;
```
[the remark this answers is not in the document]
pmeredit: You are correct here. I think I needed that in prior code and didn't check to see if it was necessary after I changed the code (it was probably a `Handle` instead of a `MongoHandle` at some point).

pmeredit:
evergreen merge

PR #6, opened 2021-11-23
SQL-619: Add sanitizer tests

pmeredit:
This is based off SQL-616 Because I needed code to test. It won't be merged until 616 is merged, so that we can have a clean gitlog. So the only changes that matter here are in evergreen.yml

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -118,6 +118,48 @@ functions:
           ${prepare_shell}
           cargo test
 
+  "install unix odbc":
+    - command: shell.exec
+      type: system
+      params:
+        shell: bash
+        working_dir: mongo-odbc-driver
+        script: |
+          ${prepare_shell}
+          dir="$PWD"
+          mkdir -p unixodbc/install
+          cd unixodbc
+          echo "downloading unixODBC"
+          unixODBC_dir=unixODBC-2.3.6
+          curl -O "http://noexpire.s3.amazonaws.com/sqlproxy/binary/linux/unixODBC-2.3.6.tar.gz" \
+            --silent \
+            --fail \
+            --max-time 60 \
+            --retry 5 \
+            --retry-delay 0
+          tar xf "$unixODBC_dir.tar.gz"
+          cd "$unixODBC_dir"
+          ./configure --prefix="$dir/unixodbc/install" --with-pic
+          make
+          make install
+
+
+  "run asan tests":
+    - command: shell.exec
+      type: test
+      params:
+        shell: bash
+        working_dir: mongo-odbc-driver
+        script: |
+          ${prepare_shell}
+          dir="$PWD"
+          ~/.cargo/bin/rustup default nightly
+          ~/.cargo/bin/rustup target add x86_64-unknown-linux-gnu
+          export LD_LIBRARY_PATH="$dir/unixodbc/install/lib"
+          export LIBRARY_PATH="$dir/unixodbc/install/lib"
+          export RUSTFLAGS="-Z sanitizer=address"
+          cargo test --target x86_64-unknown-linux-gnu
+
 pre:
   - func: "fetch source"
   - func: "generate expansions"
@@ -144,6 +186,12 @@ tasks:
       - func: "install rust toolchain"
       - func: "run rust tests"
 
+  - name: asan
+    commands:
+      - func: "install rust toolchain"
+      - func: "install unix odbc"
+      - func: "run asan tests"
+
 buildvariants:
 
   - name: static-analysis
@@ -166,3 +214,9 @@ buildvariants:
     tasks:
       - name: compile
       - name: test-rust
+
+  - name: ubuntu1804
+    display_name: Ubuntu Sanitizers
+    run_on: [ ubuntu1804-test ]
+    tasks:
+      - name: asan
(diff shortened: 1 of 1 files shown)


pmeredit:
Note that the tests properly fail in the patch before the last one here.

evergreen.yml
```diff
@@ -118,6 +118,48 @@ functions:
...
+            --retry 5 \
+            --retry-delay 0
+          tar xf "$unixODBC_dir.tar.gz"
+          cd "$unixODBC_dir"
+          ./configure --prefix="$dir/unixodbc/install" --with-pic
+          make
+          make install
+
+
+  "run asan tests":
```
nbagnard, line 147: Once we get past just allocating and freeing handle, we will need to build and then setup the driver using odbcinst.ini and odbc.ini config files, right?

evergreen.yml
```diff
@@ -118,6 +118,48 @@ functions:
...
+        shell: bash
+        working_dir: mongo-odbc-driver
+        script: |
+          ${prepare_shell}
+          dir="$PWD"
+          ~/.cargo/bin/rustup default nightly
+          ~/.cargo/bin/rustup target add x86_64-unknown-linux-gnu
+          export LD_LIBRARY_PATH="$dir/unixodbc/install/lib"
+          export LIBRARY_PATH="$dir/unixodbc/install/lib"
+          export RUSTFLAGS="-Z sanitizer=address"
```
nbagnard, line 160: What about leak and memory sanitizers? Can we run them too? Looking at the examples from rust-san readme, it seems that we could quite easily have leaks (for a newbie with the language like me at least).

pmeredit:
@rychipman @nbagnard this has been rebased on master, so now it clearly reflects that only the evergreen.yml file needed to change

evergreen.yml
pmeredit: I don't think that is necessary for unit tests. We will definitely want that for integration tests.

evergreen.yml
pmeredit: `address` actually catches `leaks`, I believe `leak` is just a faster version that catches only `leaks`. It seems faster and uses slightly different output for leaks, but I don't see a reason to run both when `address` catches leaks. What is `memory`, isn't that just `address`?

nbagnard: oh, I was looking at the doc for the sanitizer here : https://github.com/japaric/rust-san
Seems like I should have look here : https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html
Then, yes, address is enough :)

nbagnard (approved):
LGTM!

evergreen.yml
```diff
@@ -166,3 +214,9 @@ buildvariants:
     tasks:
       - name: compile
       - name: test-rust
+
+  - name: ubuntu1804
+    display_name: Ubuntu Sanitizers
+    run_on: [ ubuntu1804-test ]
```
rychipman, line 220:
```suggestion
  - name: ubuntu2004
    display_name: Ubuntu 20.04
    run_on: [ ubuntu2004-large ]
```
a couple comments:
- Is there any reason not to use the newest ubuntu?
- I don't think there's any need to have a sanitizer-specific variant -- that's a property of the task, not the variant -- if we start building the driver on Linux in the future, it can be on this variant too

evergreen.yml
```diff
@@ -118,6 +118,48 @@ functions:
...
+        working_dir: mongo-odbc-driver
+        script: |
+          ${prepare_shell}
+          dir="$PWD"
+          ~/.cargo/bin/rustup default nightly
+          ~/.cargo/bin/rustup target add x86_64-unknown-linux-gnu
+          export LD_LIBRARY_PATH="$dir/unixodbc/install/lib"
+          export LIBRARY_PATH="$dir/unixodbc/install/lib"
+          export RUSTFLAGS="-Z sanitizer=address"
+          cargo test --target x86_64-unknown-linux-gnu
```
rychipman, line 161: why is this explicit `--target` flag needed? Wouldn't this be the target by default?

evergreen.yml
```diff
@@ -118,6 +118,48 @@ functions:
...
+      params:
+        shell: bash
+        working_dir: mongo-odbc-driver
+        script: |
+          ${prepare_shell}
+          dir="$PWD"
+          ~/.cargo/bin/rustup default nightly
+          ~/.cargo/bin/rustup target add x86_64-unknown-linux-gnu
+          export LD_LIBRARY_PATH="$dir/unixodbc/install/lib"
+          export LIBRARY_PATH="$dir/unixodbc/install/lib"
```
rychipman, line 159: instead of hard-coding this path into this function and the "install unix odbc" function, let's modify the "generate expansions" task to add a new `UNIX_ODBC_PATH` var to the env set up by `prepare_shell` and use that var from both functions

rychipman (changes requested):

evergreen.yml
pmeredit: No, it doesn't work without the `--target` flag, and I don't actually know why.

rychipman (approved):
LGTM

pmeredit:
evergreen merge

PR #12, opened 2022-07-29
SQL-634: Implement connect

pmeredit:
I wouldn't consider this a complete PR yet because there are a few things I'm still mulling.

* I'm thinking of leaving out the URI handling for a separate PR. I think we have enough to discuss in this PR without it,
I also haven't decided how it should work on linux (though it's pretty straight forward on Windows):
  * Should we just assume INIs are in the right place, or is there some way we can get the output of `odbcinst -j` to know where the INIs are?

* I'm not currently using SUCCESS_WITH_INFO, I will need to see exactly where that comes up. As well as the output_buffer (is it only used in cases where we prompt for more info? I need to check)

tell me what you think about the error handling with respect to referring mongodb errors.


core/src/conn.rs
```diff
@@ -22,14 +22,24 @@ impl MongoConnection {
...
-        _current_db: Option<&str>,
-        _operation_timeout: Option<i32>,
-        _login_timeout: Option<i32>,
+        uri: &str,
+        current_db: Option<String>,
+        operation_timeout: Option<i32>,
+        login_timeout: Option<i32>,
     ) -> Result<Self> {
-        unimplemented!()
+        println!("uri = {:?}", uri);
```
pmeredit, line 33: This is just until I figure out how to handle looking up DSN

core/src/conn.rs
```diff
@@ -22,14 +22,23 @@ impl MongoConnection {
     // The operation will timeout if it takes more than loginTimeout seconds.
     // The initial current database if provided should come from SQL_ATTR_CURRENT_CATALOG
```
pmeredit, line 23: @nbagnard any idea where we should read the value of `SQL_ATTR_CURRENT_CATALOG`?

pmeredit:
This is working fine locally. Seems we need to setup some stuff.

odbc/src/api/functions.rs
```diff
@@ -380,16 +389,53 @@ pub extern "C" fn SQLDriverConnect(
...
     _driver_completion: DriverConnectOption,
 ) -> SqlReturn {
-    unimplemented!()
+    return SqlReturn::SUCCESS;
+    println!("______");
+    let conn_handle = MongoHandleRef::from(connection_handle);
+    let conn = (*conn_handle).as_connection();
+    if conn.is_none() {
+        conn_handle.add_diag_info(ODBCError::InvalidHandleType("Connection"));
+        return SqlReturn::ERROR;
```
rychipman, line 407: According to the [ODBC Docs](https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqldriverconnect-function?view=sql-server-ver16) it appears that both `SQLDriverConnect` and `SQLDisconnect` have more error cases than the ones we've implemented here

Let's make sure we cover the stuff that's been documented and also unit/integration test them as needed to get coverage!

odbc/src/api/functions.rs
```diff
@@ -1655,6 +1701,30 @@ mod util {
     use odbc_sys::{Integer, SmallInt, SqlReturn, WChar};
     use std::{cmp::min, ptr::copy_nonoverlapping};
 
+    /// input_wtext_to_string converts an input cstring to a rust String.
+    /// It assumes nul termination if the supplied length is negative.
+    #[allow(clippy::uninit_vec)]
```
rychipman, line 1706: do we need to just allow the lint here? A quick google suggests that perhaps there's another solution by using `MaybeUninit`?

rychipman (changes requested):
I know this is WIP, but just putting up some comments that I have so far -- feel free to re-tag me whenever you want me to take a look, but also no hurry if you want to just spend some time fixing stuff and getting it fleshed out first

odbc/src/api/functions.rs
pmeredit: Definitely. I want to make sure to handle the MoreInfo cases, too.

odbc/src/api/functions.rs
```diff
@@ -380,16 +389,42 @@ pub extern "C" fn SQLDriverConnect(
...
     _string_length_2: *mut SmallInt,
     _driver_completion: DriverConnectOption,
 ) -> SqlReturn {
-    unimplemented!()
+    let conn_handle = MongoHandleRef::from(connection_handle);
+    let conn = (*conn_handle).as_connection();
+    if conn.is_none() {
+        conn_handle.add_diag_info(ODBCError::InvalidHandleType("Connection"));
+        return SqlReturn::ERROR;
+    }
```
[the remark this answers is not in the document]
pmeredit: I did not intend to change error handling at all. I'll clean that up.

odbc/tests/connection_tests.rs
```diff
@@ -13,7 +13,7 @@ fn test_invalid_connection() {
     );
 }
 #[allow(dead_code)]
-//#[test]
+#[test]
```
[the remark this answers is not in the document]
pmeredit: I don't want to merge this until I get it working :)

It's working for me locally now (finally), but I think we are missing the necessary test infra to connect (we'll see in a second).

Locally, I'm connecting directly to a running mongod (since no real sql specific infra is necessary to check connecting).

odbc/src/api/functions.rs
pmeredit: I was looking at `MaybeUninit` and couldn't figure out how to get it to work with `copy_non_overlapping`. It told me I couldn't use the pointer to that vec in `copy_non_overlapping`, and a quick look didn't reveal any examples working with both `copy_non_overlapping` and `MaybeUninit`. I can definitely look again.

Anyway, this PR is definitely still very much a work in progress :)

core/src/lib.rs
```diff
@@ -1,11 +1,11 @@
 #![allow(dead_code)]
 mod collections;
 pub use collections::MongoCollections;
-mod conn;
+pub mod conn;
```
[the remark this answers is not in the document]
pmeredit: It complains `conn` is private in the `odbc` crate.

core/src/conn.rs
```diff
@@ -17,19 +17,136 @@ pub struct MongoConnection {
...
+    //
+    // The operation will timeout if it takes more than loginTimeout seconds. This timeout is
+    // delegated to the mongo rust driver.
+    //
     // The initial current database if provided should come from SQL_ATTR_CURRENT_CATALOG
     // and will take precedence over the database setting specified in the uri if any.
-    // The initial operation time if provided should come from  and will take precedence over the
+    // The initial operation time if provided should come from and will take precedence over the
     // setting specified in the uri if any.
     pub fn connect(
```
nbagnard, line 30: I am not sure about the separation between the odbc and core layers here. I thought that we wanted the core package to be a thin wrapper around the Rust driver and not hold much ODBC specific knowledge.
Because of that I was expecting that `connect` here could be used by all the SQLxxConnect methods when/if we add more.
Also, I think that dealing with loading DSN info from registry or INI files is definitely specific to ODBC, and I would put it in the odbc package.
On way to do this would be:
- In the odbc package, do the gymnastic you have here to transform ODBC connection setting info (either DSN info or direct connection strings plus any SQLxxConnect arguments like UID/PWD for example) to a MongoDB uri. The code can/should be in helpers so that all SQLxxConnect functions can use it.
- Keep the code you have from line 58 onward, and add the update of the value for database, connectTimeoutMS using current_db and login_timeout.

This way the code would be ready for us if/when we had more SQLxxConnect methods.
What do you think?

core/src/conn.rs
```diff
@@ -17,19 +17,136 @@ pub struct MongoConnection {
...
+        if uri.is_empty() {
+            return Err(Error::UriFormatError("uri must not be empty".to_string()));
+        }
+        // TODO: handle DSN expansion here (i.e., lookup the DSN and add the resolved attributes).
+        // split the uri attributes on ';'
+        uri.split(';')
+            .filter(|attr| !attr.is_empty())
+            .map(|attr| {
+                // now split each attribute pair on '='
+                let mut sp = attr.split('=').map(String::from).collect::<Vec<_>>();
```
nbagnard, line 103: Similarly with `=`. It could be that `=` is part of a value.

core/src/conn.rs
```diff
@@ -17,19 +17,136 @@ pub struct MongoConnection {
...
+    );
+
+    assert_eq!(
+        expected,
+        MongoConnection::get_attributes("    Driver  =  Foo   ;    SERVER  =   bAr  ").unwrap()
+    );
+
+    assert_eq!(
+        expected,
+        MongoConnection::get_attributes("    Driver  =  Foo   ;    SERVER  =   bAr  ;").unwrap()
```
nbagnard, line 150: [nit]Why using a connection string with extra blank on this test case? If I understand it correctly, this is to check that a trailing `;` is supported, right?
What about using a short connection string, like `driver=Foo;server=Bar;`? It makes the purpose of the test easier to see.

nbagnard (changes requested):
First pass, I only reviewed the changes in the core package for now.

odbc/src/api/functions.rs
```diff
@@ -375,21 +409,57 @@ pub extern "C" fn SQLDriverConnect(
...
+    dbg!("!!!!");
+    let conn_handle = MongoHandleRef::from(connection_handle);
+    let conn = (*conn_handle).as_connection();
+    if conn.is_none() {
+        conn_handle.add_diag_info(ODBCError::InvalidHandleType("Connection"));
+        return SqlReturn::ERROR;
+    }
+    let conn = conn.unwrap();
+    let uri = input_wtext_to_string(in_connection_string, string_length_1 as usize);
+    let database = env::var("SQL_ATTR_CURRENT_CATALOG").ok();
```
nbagnard, line 437: You should be able to retrieve the current catalog if set using ConnectionAttributes::current_db which should be present in the`attributes` struct of conn.
Now, because sqlSetConnAttr is not implemented, there is no way for a user to set it yet but `MongoConnection::connect` is an Option, so it should be alright.

odbc/src/api/functions.rs
```diff
@@ -375,21 +409,57 @@ pub extern "C" fn SQLDriverConnect(
...
+    let connect_result = mongo_odbc_core::MongoConnection::connect(
+        &uri,     // uri
+        database, // current_db
+        None,     // op timeout i32
+        None,     // login timeout i32
+    );
+    match connect_result {
+        Ok(mc) => {
+            let mut conn_writer = conn.write().unwrap();
+            conn_writer.attributes.current_db = mc.current_db.clone();
```
nbagnard, line 447: [Just a comment]
I didn't find anything in the spec which says that we should set SQL_ATTR_CURRENT_CATALOG attribute outside of a SQLSetConnAttr call but I think it makes sense here.

odbc/src/api/functions.rs
```diff
@@ -375,21 +409,57 @@ pub extern "C" fn SQLDriverConnect(
...
 #[no_mangle]
 pub extern "C" fn SQLDriverConnectW(
-    _connection_handle: HDbc,
+    connection_handle: HDbc,
     _window_handle: HWnd,
-    _in_connection_string: *const WChar,
-    _string_length_1: SmallInt,
+    in_connection_string: *const WChar,
+    string_length_1: SmallInt,
     _out_connection_string: *mut WChar,
```
nbagnard, line 422: I looked into the spec and it seems that we should copy *InConnectionString to *OutConnectionString because we are not planning to support SQL_DRIVER_PROMPT (see [Driver Guidelines](https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqldriverconnect-function?view=sql-server-ver16#driver-guidelines)).

odbc/src/api/functions.rs
```diff
@@ -1259,17 +1408,20 @@ pub extern "C" fn SQLSetPos(
     _operation: USmallInt,
     _lock_type: USmallInt,
 ) -> SqlReturn {
+    dbg!();
     unsupported_function(MongoHandleRef::from(statement_handle), "SQLSetPos")
 }
 
 #[no_mangle]
 pub extern "C" fn SQLSetEnvAttr(
```
nbagnard, line 1416: I thought we decided to just provide a Unicode driver and only implemented the Unicode (W) functions. Did you add it because it is very simple? And should we add it, because it could make things a bit confusing if we sometimes support ANSI functions, not other times.

core/src/conn.rs
pmeredit: Not sure where I got the idea to pull `SQL_ATTR_CURRENT_CATALOG` from the shell environment!

core/src/conn.rs
pmeredit: Ok, this is interesting to me, because my idea for the core api was too be an ODBC driver except for the parts that are specific to C (i.e., do as much as possible in the core api). I'm not sure on the best way to go.

core/src/conn.rs
```diff
@@ -17,19 +17,135 @@ pub struct MongoConnection {
...
-        _current_db: Option<&str>,
-        _operation_timeout: Option<i32>,
-        _login_timeout: Option<i32>,
+        uri: &str,
+        current_db: Option<String>,
+        operation_timeout: Option<i32>,
+        login_timeout: Option<i32>,
     ) -> Result<Self> {
-        unimplemented!()
+        dbg!(uri);
```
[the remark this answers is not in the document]
pmeredit: These `dbg` statements will be removed before merge!

pmeredit (commented):
@nbagnard @rychipman I've applied most of the comments, but I don't think there is a need to rereview until the local ADF changes are merged.

I changed list_databases to ping because it should take fewer permissions

The println that is still there will be removed after I am sure that the local ADF test is working

core/src/conn.rs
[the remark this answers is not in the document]
nbagnard: I see. Do we have special plan for extra debug logging or we will just rely on SQLTrace to start with?
I think SQLTrace is fine to start with. For adding more logs, I am not sure what would be best. Maybe something to look into in the design?

odbc/src/api/functions.rs
pmeredit: Filled SQL-991 to handle this so that we can start parallelizing the other work!

odbc/src/api/functions.rs
```diff
@@ -375,21 +409,57 @@ pub extern "C" fn SQLDriverConnect(
...
+    string_length_1: SmallInt,
     _out_connection_string: *mut WChar,
     _buffer_length: SmallInt,
     _string_length_2: *mut SmallInt,
     _driver_completion: DriverConnectOption,
 ) -> SqlReturn {
-    unimplemented!()
+    dbg!();
+    dbg!("!!!!");
+    let conn_handle = MongoHandleRef::from(connection_handle);
```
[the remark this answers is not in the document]
pmeredit: This change actually simplifies a lot of our functions quite a bit. Good suggestion!

odbc/src/api/functions.rs
```diff
@@ -769,57 +769,51 @@ pub extern "C" fn SQLGetDiagRecW(
...
-                        buffer_length,
-                        text_length_ptr,
-                        native_error_ptr,
-                    ),
-                    None => SqlReturn::NO_DATA,
-                }
+        HandleType::Env => {
+            let env = check_invalid! {unsafe { (*mongo_handle).as_env() }};
+            let env_contents = (*env).read().unwrap();
+            match env_contents.errors.get(rec_number) {
```
pmeredit, line 775: I think we should actually put add a HandlesContents trait or some such name with an `errors` method so we can drastically simplify this function too.

pmeredit: Unfortunately, because of the way LockGuards work there is no way to make this much better. We can use a macro, but the macro invocation is almost as big as the code already, so it only adds complexity.

core/src/conn.rs
nbagnard: Removing leading and trailing blanks is part of the previous test already and I don't think it's related to the trailing semicolon handling, so I thought you could remove them in this test.
You are right, I couldn't find where in the spec it says if leading and trailing blanks should be trimmed or not.
It might not because it says that `the value of the DSN keyword cannot consist only of blanks and should not contain leading blanks` and because for ADO.net this is [explicitly described](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ms722656(v=vs.85)#setting-values-that-use-spaces).
But really, using leading and trailing blanks is asking for troubles.

core/src/conn.rs
```diff
@@ -17,19 +17,135 @@ pub struct MongoConnection {
...
+        client_options.connect_timeout = login_timeout.map(|to| Duration::new(to as u64, 0));
+        // set application name, note that users can set their own application name, or we default
+        // to mongo-odbc-driver.
+        let application_name = attributes.remove("application_name");
+        if application_name.is_some() {
+            client_options.app_name = application_name;
+        } else {
+            client_options.app_name = Some("mongo-odbc-driver".to_string());
+        }
+        let client = Client::with_options(client_options)?;
```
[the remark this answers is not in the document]
nbagnard: I'm not sure I follow where SERVER goes into play in this?

core/src/conn.rs
```diff
@@ -17,19 +17,135 @@ pub struct MongoConnection {
...
+        }
     }
+
+    fn get_attributes(uri: &str) -> Result<BTreeMap<String, String>> {
+        if uri.is_empty() {
+            return Err(Error::UriFormatError("uri must not be empty".to_string()));
+        }
+        // TODO: handle DSN expansion here (i.e., lookup the DSN and add the resolved attributes).
+        // split the uri attributes on ';'
+        uri.split(';')
```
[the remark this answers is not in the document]
nbagnard: Do you mind if I add that we need to support UID on top or in conjunction with user in SQL-990?

core/src/conn.rs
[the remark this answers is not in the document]
pmeredit: I think SQLTrace is fine to start. We can  consider adding a logging library, but there is always the question of where it will log to (especially on windows).

core/src/conn.rs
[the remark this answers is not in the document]
pmeredit: Discussed in one on one. SERVER appears in the integration test I was running.

core/src/conn.rs
pmeredit: We discussed, the parsing for uris will be added to odbc crate and the name of `uri` in `connect` (in core) will be changed to `mongo_uri` to make things more clear

pmeredit:
Everything is actually working now 🎉

pmeredit:
@nbagnard @rychipman Note that I put all our tests in mods called either "unit" or "integration" because there seems to be no other way to split tests up in cargo. If you know a better way @rychipman, tell me!

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,191 @@
+use crate::errors::{ODBCError, Result};
```
pmeredit, line 1: This gives us a useful interface that should continue to work no matter what we do with the actually parsing in the `new` associated function come SQL-990. Note that this is a rather strange interface where all the methods are destructive, but this suits all of our use cases and this is a crate level interface. By writing it this way we can avoid copying some strings. Not a big deal, ultimately, but why waste memory when this is pretty easy?

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,191 @@
...
+const EQUAL_ERROR: &str = "all uri atttributes must be of the form key=value";
+
+// TODO SQL-990: Audit these mandatory attributes
+const USER: &[&str] = &["user", "uid"];
+const PWD: &[&str] = &["pwd", "password"];
+const SERVER: &[&str] = &["server"];
+const SSL: &[&str] = &["ssl"];
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct ODBCUri<'a>(BTreeMap<String, &'a str>);
```
pmeredit, line 15: sadly, the lowercasing of the keys means that the ODBCUri must own the keys :(

If we really care we could use ` Cow` here so that it only owns the memory when lowercasing actually occurs.

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    // remove_or_else is the same as remove but with a default thunk.
+    pub fn remove_or_else(&mut self, f: impl Fn() -> &'a str, names: &[&str]) -> &'a str {
+        self.remove(names).unwrap_or_else(f)
+    }
+
+    // remove_mandatory_attribute will find an attribute that must exist and transfer ownership to
+    // the caller. It accepts a slice of names that will be checked in order for names that are
+    // synonyms (e.g., uid and user are both viable attribute names for a user). If both names
+    // exist it will only find the first.
```
pmeredit, line 61:
```suggestion
    // exist, it will only find the first.
```

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    // remove_or_else is the same as remove but with a default thunk.
+    pub fn remove_or_else(&mut self, f: impl Fn() -> &'a str, names: &[&str]) -> &'a str {
+        self.remove(names).unwrap_or_else(f)
+    }
+
+    // remove_mandatory_attribute will find an attribute that must exist and transfer ownership to
+    // the caller. It accepts a slice of names that will be checked in order for names that are
+    // synonyms (e.g., uid and user are both viable attribute names for a user). If both names
+    // exist, it will only find the first.
```
pmeredit, line 61: This is private because I only foresee its ever being used in `remove_mongo_uri`

README.md
```diff
@@ -19,9 +19,9 @@ Note: users can utilize the built-in driver manager.
```
pmeredit, line 19: ADL => ADF, ignore this file

evergreen.yml
```diff
@@ -81,6 +82,7 @@ functions:
           # rustup installs into C:\Users\$USER instead of
           # C:\home\$USER, so we symlink both .rustup and .cargo
           if [ "Windows_NT" == "$OS" ]; then
+              echo "sym linking rustup"
```
pmeredit, line 85:
```suggestion

```

odbc/src/api/definitions.rs
```diff
@@ -1,39 +1,28 @@
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, FromPrimitive)]
 pub enum SqlBool {
     False = 0,
     True,
 }
 
 // Environment attributes
 
-#[derive(Copy, Clone, Debug, PartialEq, FromPrimitive)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, FromPrimitive)]
```
pmeredit, line 11: Clippy suggested we add `Eq`, and I agree!

odbc/src/api/errors.rs
```diff
@@ -1,49 +1,57 @@
...
+
 const VENDOR_IDENTIFIER: &str = "MongoDB";
 
 // SQL states
 pub const HYC00: &str = "HYC00";
 pub const HY024: &str = "HY024";
 pub const _01S02: &str = "01S02";
 
-#[derive(Debug)]
+#[derive(Debug, Error)]
```
pmeredit, line 10: Switched to using `thiserror` to be consistent with our other projects

odbc/src/api/errors.rs
```diff
@@ -1,49 +1,57 @@
...
         match self {
             ODBCError::Unimplemented(_) => HYC00,
-            ODBCError::InvalidAttrValue(_) => HY024,
+            ODBCError::CoreError(_) => HYC00,
+            ODBCError::InvalidUriFormat(_) | ODBCError::InvalidAttrValue(_) => HY024,
+            ODBCError::InvalidHandleType(_) => HYC00,
             ODBCError::OptionValueChanged(_, _) => _01S02,
         }
     }
     pub fn get_error_message(&self) -> String {
```
pmeredit, line 41: Not sure we actually need the `get_error_message` method since all it does is call format.

core/src/conn.rs
```diff
@@ -16,20 +21,44 @@ pub struct MongoConnection {
...
+        // to mongo-odbc-driver.
+        client_options.app_name = application_name
+            .map(String::from)
+            .or_else(|| Some(MONGODB_ODBC_DRIVER.to_string()));
+        let client = Client::with_options(client_options)?;
+        // run the "ping" command on the `auth_src` database. We assume this requires the
+        // fewest permissions of anything we can do to verify a connection.
+        client
+            .database(auth_src)
+            .run_command(doc! {"ping": 1}, None)?;
```
nbagnard, line 57: This is different from the JDBC driver where we are doing "SELECT 1 from DUAL". Is there a reason? Is one better than the other and in this case, should we update the JDBC driver at some point?

nbagnard (changes requested):
It's looking good!
I only have a few suggestions and mainly a bunch of questions.

resources/run_adf.sh
```diff
@@ -6,13 +6,13 @@
 # This script will start a local mongod and Atlas Data Federation instance, used for integration testing.
 # The supported platforms are windows, macos, ubuntu1804, and rhel7.
 #
-# - To skip the download of ADL, set the environment variable HAVE_LOCAL_MONGOHOUSE to 1
+# - To skip the download of ADF, set the environment variable HAVE_LOCAL_MONGOHOUSE to 1
```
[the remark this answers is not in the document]
pmeredit: I did not.

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,187 @@
...
+        assert!(ODBCUri::new("").is_err());
+        assert!(ODBCUri::new("Foo").is_err());
+        assert!(ODBCUri::new("driver=Foo;Bar").is_err());
+
+        let expected = ODBCUri(map! {"driver".to_string() => "Foo"});
+        assert_eq!(expected, ODBCUri::new("Driver=Foo").unwrap());
+
+        let expected = ODBCUri(map! {"driver".to_string() => "Foo", "server".to_string() => "bAr"});
+        assert_eq!(expected, ODBCUri::new("Driver=Foo;SERVER=bAr").unwrap());
+        assert_eq!(expected, ODBCUri::new("Driver=Foo;SERVER=bAr;").unwrap());
```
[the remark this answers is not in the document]
pmeredit: sgtm

odbc/src/api/functions.rs
```diff
@@ -1674,12 +1769,23 @@ mod util {
         buffer_len: usize,
         text_length_ptr: *mut SmallInt,
     ) -> SqlReturn {
+        assert!(!error_message.is_empty());
```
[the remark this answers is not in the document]
pmeredit: `error_message` being empty would point to a coder error. It should never happen!

evergreen.yml
```diff
@@ -43,10 +43,13 @@ functions:
             export CARGO_NET_GIT_FETCH_WITH_CLI="$CARGO_NET_GIT_FETCH_WITH_CLI"
             export UNIX_ODBC_PATH="$PWD/unixodbc/install"
             git config --global url."ssh://git@github.com/".insteadOf "https://github.com/"
-            export ADL_TEST_USER=test_user
-            export ADL_TEST_PWD=test_pwd
-            export ADL_TEST_HOST=test_host
-            export ADL_TEST_DB=test_db
+            # TODO: remove these with local ADF. This isn't a real security issue because
```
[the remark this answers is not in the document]
pmeredit: THANKS!

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,187 @@
...
+    assert!(ODBCUri::new("").is_err());
+    assert!(ODBCUri::new("Foo").is_err());
+    assert!(ODBCUri::new("driver=Foo;Bar").is_err());
+
+    let expected = ODBCUri(map! {"driver".to_string() => "Foo"});
+    assert_eq!(expected, ODBCUri::new("Driver=Foo").unwrap());
+
+    let expected = ODBCUri(map! {"driver".to_string() => "Foo", "server".to_string() => "bAr"});
+    assert_eq!(expected, ODBCUri::new("Driver=Foo;SERVER=bAr").unwrap());
+    assert_eq!(expected, ODBCUri::new("Driver=Foo;SERVER=bAr;").unwrap());
```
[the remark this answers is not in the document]
pmeredit: I added comments to all the tests

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,187 @@
...
+        for (i, name) in names.iter().enumerate() {
+            len = i;
+            let result = self.0.remove(&name.to_string());
+            if let Some(ret) = result {
+                return Ok(ret);
+            }
+        }
+        if len == 1 {
+            Err(ODBCError::InvalidUriFormat(format!(
+                "{} is required for a valid Mongo ODBC Uri",
```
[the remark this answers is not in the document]
pmeredit: Yes, we can leave that off this review though!

odbc/src/api/errors.rs
```diff
@@ -1,49 +1,57 @@
...
 }
 
+pub type Result<T> = std::result::Result<T, ODBCError>;
+
 impl ODBCError {
     pub fn get_sql_state(&self) -> &str {
         match self {
             ODBCError::Unimplemented(_) => HYC00,
-            ODBCError::InvalidAttrValue(_) => HY024,
+            ODBCError::CoreError(_) => HYC00,
```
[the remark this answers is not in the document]
pmeredit: We can inspect the Core error :) I'll add a TODO here.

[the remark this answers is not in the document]
pmeredit: Actually, we'll do this as a method on the CoreError, which I've added a skeleton for. We might want to make a separate crate for sql_state so that we don't repeat a subset of the sql_states in the dll, but I'm not _too_ worried about it.

odbc/src/api/errors.rs
nbagnard: If we don't, then, let's just get rid of it. I'm fine with it.

pmeredit:
@rychipman @nbagnard 
I decided it bothered me too much that some static strings would appear twice in the DLL, so now we have a constants crate.

pmeredit:
@rychipman @nbagnard I'm unsure if we should add Varsha to the authors in the toml files. She is an author, but I'm not sure if she would want someone to contact her in the future about some bug.

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,187 @@
...
+    // remove_mongo_uri converts this ODBCUri to a mongo_uri String. It will
+    // remove all the attributes necessary to make a mongo_uri. This is destructive!
+    pub fn remove_mongo_uri(&mut self) -> Result<String> {
+        let user = self.remove_mandatory_attribute(USER)?;
+        let pwd = self.remove_mandatory_attribute(PWD)?;
+        let server = self.remove_mandatory_attribute(SERVER)?;
+        let ssl = self.remove(SSL);
+        // TODO: SQL-990: we may wish to support more attributes as options.
+        // If we do, add more tests to cover them.
+        let ssl_string =
```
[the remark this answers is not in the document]
pmeredit: It would be api breaking for any of the driver to change the default to true.

odbc/src/api/functions.rs
```diff
@@ -692,14 +803,14 @@ pub extern "C" fn SQLGetDiagField(
...
-    _handle: Handle,
+    handle: Handle,
     _record_rumber: SmallInt,
     _diag_identifier: SmallInt,
     _diag_info_ptr: Pointer,
     _buffer_length: SmallInt,
     _string_length_ptr: *mut SmallInt,
 ) -> SqlReturn {
-    unimplemented!()
+    unsupported_function(MongoHandleRef::from(handle), "SQLGetDiagFieldW")
```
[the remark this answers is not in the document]
pmeredit: I think the idea was `unimplemented` would be implemented eventually. Perhaps I should set this back to `unimplemented`

core/src/conn.rs
pmeredit: I think we should change the JDBC driver.

nbagnard (approved):
Looks like we have some funkiness going on with GitHub sync of the changes.
But remaining comments are very minor.
LGTM!
Connecting to MongoDB - check! 🎉

nbagnard:
> @rychipman @nbagnard I'm unsure if we should add Varsha to the authors in the toml files. She is an author, but I'm not sure if she would want someone to contact her in the future about some bug.

You can add her name and deactivated Mongodb email address maybe?

core/src/conn.rs
nbagnard: Sounds good. I create [SQL-998](https://jira.mongodb.org/browse/SQL-998) for this.

rychipman:
Switched myself out for @mattChiaravalloti. For whatever reason, it seems I can't actually remove myself, since I had posted a set of comments in the past, but I'll just permanently leave myself in "review requested", and you should feel free to merge without me!

odbc/src/api/errors.rs
```diff
@@ -1,49 +1,50 @@
...
+    InvalidHandleType(&'static str),
+    #[error("[{}][API] Invalid value for attribute {0}", VENDOR_IDENTIFIER)]
     InvalidAttrValue(&'static str),
+    #[error(
+        "[{}][API] Invalid value for attribute {0}, changed to {1}",
+        VENDOR_IDENTIFIER
+    )]
     OptionValueChanged(&'static str, &'static str),
+    #[error("[{}][Core] {0}", VENDOR_IDENTIFIER)]
+    CoreError(#[from] mongo_odbc_core::Error),
```
mattChiaravalloti, line 20:
```suggestion
    Core(#[from] mongo_odbc_core::Error),
```
[nit] Should we omit the suffix "Error" from this variant name since "Error" is implied by the enum name itself?

odbc/src/api/errors.rs
```diff
@@ -1,49 +1,50 @@
...
 }
 
+pub type Result<T> = std::result::Result<T, ODBCError>;
+
 impl ODBCError {
     pub fn get_sql_state(&self) -> &str {
         match self {
             ODBCError::Unimplemented(_) => HYC00,
-            ODBCError::InvalidAttrValue(_) => HY024,
+            // TODO: set the state based on the type of CoreError since they may differ
```
mattChiaravalloti, line 29: This TODO seems addressed given the case on line 30. Is this still a TODO that is being tracked by a future ticket? If yes, should we note the ticket number as part of this TODO?

core/src/err.rs
```diff
@@ -1,9 +1,40 @@
...
-    MongoDriver(#[from] mongodb::error::Error), // Source and Display delegate to mongodb::Error
+    MongoError(#[from] mongodb::error::Error),
+}
+
+impl Error {
+    pub fn get_sql_state(&self) -> &'static str {
+        match self {
+            // TODO: for now we just return HY024 for all Mongo Errors.
+            // In the future this will change based on the type of error.
+            Error::MongoError(_) => HYC00,
```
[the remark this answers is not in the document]
mattChiaravalloti: btw should we associate a ticket number with this TODO?

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    #[test]
+    fn test_remove_to_mongo_uri() {
+        use super::*;
+        // test missing SERVER
+        assert!(ODBCUri::new("USER=foo;PWD=bar")
+            .unwrap()
+            .remove_to_mongo_uri()
+            .is_err());
```
mattChiaravalloti, line 133: Should we be asserting specific errors here? I guess the specific thing we'd be looking for is the presence of the expected missing key in the error message. It is possible we edit `remove_to_mongo_uri` in the future such that `"USER=foo;PWD=bar" still encounters an error but a different error than "SERVER required...", which means we'll lose the coverage we think we currently have, right?

odbc/src/api/odbc_uri.rs
```diff
@@ -0,0 +1,201 @@
...
+                ""
+            };
+        Ok(format!(
+            "mongodb://{}:{}@{}{}",
+            user, pwd, server, ssl_string
+        ))
+    }
+}
+
+mod unit {
```
mattChiaravalloti, line 103: I'm wondering if individually named tests would be better than two functions with many assertions each. I know line numbers will help us sort out what's failing, but doesn't this mean we'll need to run the tests sequentially to see all failures? By that, I mean if an assertion on line 112 fails, we won't see if the assertions after that pass or fail since the test will terminate. That might not lend itself to easiest maintenance in the future.

odbc/src/api/functions.rs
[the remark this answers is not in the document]
mattChiaravalloti: I think changing `unimplemented!` to `unsupported_function(...)` is cleaner, but ultimately that would just be work we'd undo eventually anyway so I'm not sure it's worth it.

odbc/src/api/functions.rs
```diff
@@ -882,134 +964,132 @@ pub extern "C" fn SQLGetStmtAttrW(
...
-                    },
-                    StatementAttribute::AsyncStmtEvent => unsafe {
-                        *(value_ptr as *mut _) = stmt_contents.attributes.async_stmt_event;
-                    },
-                }
-            }
-            SqlReturn::SUCCESS
+    let stmt = must_be_valid!(stmt_handle.as_statement());
+    if value_ptr.is_null() {
+        return SqlReturn::SUCCESS;
```
mattChiaravalloti, line 969: Can we add a comment explaining why this is the return value in this case?

odbc/src/api/functions.rs
```diff
@@ -1265,11 +1345,12 @@ pub extern "C" fn SQLSetPos(
...
 pub extern "C" fn SQLSetEnvAttr(
     environment_handle: HEnv,
-    _attribute: EnvironmentAttribute,
-    _value: Pointer,
+    attribute: EnvironmentAttribute,
+    value: Pointer,
     _string_length: Integer,
 ) -> SqlReturn {
-    unsupported_function(MongoHandleRef::from(environment_handle), "SQLSetEnvAttr")
+    //unsupported_function(MongoHandleRef::from(environment_handle), "SQLSetEnvAttr")
```
mattChiaravalloti, line 1352:
```suggestion
```

odbc/src/api/functions.rs
```diff
@@ -1655,8 +1723,35 @@ mod util {
...
+    pub fn input_wtext_to_string(text: *const WChar, len: usize) -> String {
+        if (len as isize) < 0 {
+            let mut dst = Vec::new();
+            let mut itr = text;
+            unsafe {
+                while *itr != 0 {
+                    dst.push(*itr);
+                    itr = itr.offset(1);
+                }
+            }
```
mattChiaravalloti, line 1738: For my own understanding, why does this need to be in an `unsafe` block?

mattChiaravalloti (changes requested):
This is a lot to take in. I think I understood an adequate amount of it. There is ample code-commenting, which is great! I have a few suggestions below that range from minor clean-up to structural changes (particularly around tests) to questions for my own understanding. Hopefully this comments won't prolong this PR too much longer!

core/src/err.rs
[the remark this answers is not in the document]
nbagnard: Oh, darn, I missed it. The `C00` looked too much like a `000`.
I think HY000 (General error) is the best choice.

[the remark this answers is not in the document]
pmeredit: I think it will need to be readdressed through multiple tickets, as we implement functionality. Not sure on the best way to organize that.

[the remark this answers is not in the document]
pmeredit: HY000 | General driver defined error
I think is what we want for now.

odbc/src/api/errors.rs
pmeredit: Good idea!

odbc/src/api/errors.rs
pmeredit: TODO is stale.

odbc/src/api/odbc_uri.rs
pmeredit: I think the only thing that sort of works is checking the formatted error message. So I added that...

core/src/conn.rs
```diff
@@ -16,20 +21,44 @@ pub struct MongoConnection {
     pub operation_timeout: Option<Duration>,
 }
 
+const MONGODB_ODBC_DRIVER: &str = "mongo_odbc_driver";
+
 impl MongoConnection {
-    // Creates a new MongoConnection with the given settings and execute a dummy aggregation to
-    // validate it works.
-    // The operation will timeout if it takes more than loginTimeout seconds.
+    // Creates a new MongoConnection with the given settings and lists databases to make sure the
```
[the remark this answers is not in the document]
pmeredit: Updated to just say "runs a command". That way we can change what we want to do later.

odbc/src/api/odbc_uri.rs
pmeredit: Couldn't hurt.

odbc/src/api/functions.rs
pmeredit: I think it should actually be `ERROR`. That was left over from fixing segfaults on win64.

odbc/src/api/functions.rs
pmeredit: You can only dereference raw pointers in unsafe blocks. In fact, this is generally the only place you need unsafe (there are a few others, but that's the most common). You can do whatever you want to a pointer whenever you want as long as it's never dereferenced (adding to a pointer is safe, so long as you do not dereference it).

odbc/src/api/odbc_uri.rs
pmeredit: I also changed the name of `tests.rs` to `unit.rs` and removed the sub `unit` module. All the tests in there are unit tests, it was redundant. 
For clarity:
We need `integration` and `unit` in every test name because we now use `cargo test unit` and `cargo test integration` to run those separately in different evergreen tasks.

It's something we might want to consider for mongosql-rs at some point, too. But we already have the tests that run separately because they require go.

mattChiaravalloti (approved):
lgtm 💯

pmeredit:
evergreen merge

PR #14, opened 2022-08-17
SQL-983: Implement query and function runner

bucaojit:
Implemented test runner that gets a connection using a connection string.  It loads yaml files from the `resources/integration_test/tests` directory and can run a query or functions.  

Many of the functions that I intended to run, based on the ones we will support in the design doc and return a result set, I found are not supported in odbc-sys so they are included but commented out.

This test runner will be extended with subsequent tickets to validate the rows and add test cases.


integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+/// parse_test_file_yaml deserializes the given YAML file into a
+/// IntegrationTest struct.
+pub fn parse_test_file_yaml(path: &str) -> Result<IntegrationTest, Error> {
+    let mut f = fs::File::open(path).map_err(|e| Error::InvalidFile(format!("{:?}", e)))?;
+    let mut contents = String::new();
+    f.read_to_string(&mut contents)
+        .map_err(|e| Error::CannotReadFileToString(format!("{:?}", e)))?;
+    let yaml: IntegrationTest = serde_yaml::from_str(&contents)
+        .map_err(|e| Error::CannotDeserializeYaml(format!("{:?}", e)))?;
+    Ok(yaml)
```
mattChiaravalloti, line 96:
```suggestion
    let f = fs::File::open(path).map_err(|e| Error::InvalidFile(format!("{:?}", e)))?;
    let integration_test: IntegrationTest = serde_yaml::from_reader(f).map_err(|e| Error::CannotDeserializeYaml(...))?;
    Ok(integration_test)
```
I've been using `serder_yaml::from_reader` directly on the `File` instead of going to `String` as an intermediate step. Is there a reason you have the `File` -> `String` step before going from `String` -> data structure?

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+}
+
+/// connect obtains a connection to the local Atlas Data Federation instance and returns Connection
+/// it fetches environment variables to create the connection string
+fn connect() -> Result<Connection<'static>, Error> {
+    let user_name = env::var("ADL_TEST_USER").expect("ADL_TEST_USER is not set");
+    let password = env::var("ADL_TEST_PWD").expect("ADL_TEST_PWD is not set");
+    let host = env::var("ADL_TEST_HOST").expect("ADL_TEST_HOST is not set");
+    let auth_db = match env::var("ADL_TEST_AUTH_DB") {
```
mattChiaravalloti, line 105: In #12 , Patrick is changing these to the format `ADF_TEST_LOCAL_xyz`, if you want to jump ahead onto that format instead of this one!

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+#[test]
+#[ignore]
+pub fn integration_test() -> Result<(), Error> {
+    let paths = load_file_paths(PathBuf::from(TEST_FILE_DIR)).unwrap();
+    for path in paths {
+        let yaml = parse_test_file_yaml(&path).unwrap();
+
+        for test in yaml.tests {
+            match test.skip_reason {
+                Some(_) => continue,
```
mattChiaravalloti, line 140: Should we print/log some indication of the test being skipped, similar to how we do for the mongosql-rs spec tests?

resources/run_adf.sh
```diff
@@ -6,13 +6,13 @@
 # This script will start a local mongod and Atlas Data Federation instance, used for integration testing.
```
bucaojit, line 6: Most of these changes are from PR 12 to change how `go` is called.  I added a few similar changes for the `jq` command as the evergreen host has trouble finding it in the path.

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    value.as_i64().expect("Unable to cast value as i64") as i16
+}
+
+fn check_array_length(array: &Vec<Value>, length: usize) {
+    if array.len() < length {
+        panic!(
+            "not enough values in array expected: {}, actual: {}",
+            length,
+            array.len()
+        )
```
mattChiaravalloti, line 190: Should this instead return an error which can be propagated all the way up to `integration_test` and return like any other error, or panicking the behavior we prefer in this case?

mattChiaravalloti (changes requested):
This is really nice! 💯 I just left a few small suggestions, otherwise this is already in really good shape.

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+                    to_i16(&function[6]),
+                ))
+            }
+        }
+         */
+        _ => Err(format!("unknown function {}", function_name)),
+    };
+
+    let sql_return_val = sql_return.unwrap();
+    if (sql_return_val != SqlReturn::SUCCESS) && (sql_return_val != SqlReturn::SUCCESS_WITH_INFO) {
```
nbagnard, line 288: Do we expect SUCCESS_WITH_INFO at any point?
SUCCESS_WITH_INFO means something is not quite right.

nbagnard (changes requested):
A couple of minor comments, but generally, it's looking good :)

integration_test/src/test_runner/mod.rs
bucaojit: I tried the suggested method and it worked as expected.  Good to know the intermediate `String` step is not needed.
I've made this change 👍

integration_test/src/test_runner/mod.rs
bucaojit: Sounds good, I've updated these to match what is in Patrick's PR.

integration_test/src/test_runner/mod.rs
bucaojit: Sure, I added a print of the skip reason.

integration_test/src/test_runner/mod.rs
bucaojit: Okay, made the change to return an error instead.

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+    let f = fs::File::open(path).map_err(|e| Error::InvalidFile(format!("{:?}", e)))?;
+    let integration_test: IntegrationTest =
+        serde_yaml::from_reader(f).map_err(|e| Error::CannotDeserializeYaml(format!("{:?}", e)))?;
+    Ok(integration_test)
+}
+
+/// connect obtains a connection to the local Atlas Data Federation instance and returns Connection
+/// it fetches environment variables to create the connection string
+fn connect() -> Result<Connection<'static>, Error> {
+    let user_name = env::var("ADL_TEST_LOCAL_USER").expect("ADL_TEST_LOCAL_USER is not set");
```
mattChiaravalloti, line 99:
```suggestion
    let user_name = env::var("ADF_TEST_LOCAL_USER").expect("ADL_TEST_LOCAL_USER is not set");
```
watch out: these still have `L` instead of `F` throughout this file!

mattChiaravalloti (approved):
looks good other than the `L` -> `F` issue ✅

I am curious to see if Natacha's idea for the `TestDef` enum works! That could really help make this even cleaner 🎉  no need to retag me after addressing her comments, though!

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    let user_name = env::var("ADL_TEST_USER").expect("ADL_TEST_USER is not set");
+    let password = env::var("ADL_TEST_PWD").expect("ADL_TEST_PWD is not set");
+    let host = env::var("ADL_TEST_HOST").expect("ADL_TEST_HOST is not set");
+    let auth_db = match env::var("ADL_TEST_AUTH_DB") {
+        Ok(val) => val,
+        Err(_) => "admin".to_string(), //Default auth db
+    };
+    let db = match env::var("ADL_TEST_DB") {
+        Ok(val) => val,
+        Err(_) => "INTEGRATION_TEST".to_string(), //Default driver name
```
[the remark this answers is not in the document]
bucaojit: Good catch, made the change.

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+        .map_err(|e| Error::CannotReadFileToString(format!("{:?}", e)))?;
+    let yaml: IntegrationTest = serde_yaml::from_str(&contents)
+        .map_err(|e| Error::CannotDeserializeYaml(format!("{:?}", e)))?;
+    Ok(yaml)
+}
+
+/// connect obtains a connection to the local Atlas Data Federation instance and returns Connection
+/// it fetches environment variables to create the connection string
+fn connect() -> Result<Connection<'static>, Error> {
+    let user_name = env::var("ADL_TEST_USER").expect("ADL_TEST_USER is not set");
```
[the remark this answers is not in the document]
bucaojit: Sounds good, made the change.

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    let paths = load_file_paths(PathBuf::from(TEST_FILE_DIR)).unwrap();
+    for path in paths {
+        let yaml = parse_test_file_yaml(&path).unwrap();
+
+        for test in yaml.tests {
+            match test.skip_reason {
+                Some(_) => continue,
+                None => {
+                    let connection = connect().unwrap();
+                    let test_result = match (&test.query, &test.function) {
```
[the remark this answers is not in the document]
bucaojit: This worked, I added the ` pub test_definition: TestDef` entry in the yaml definition.  
Took some time to figure out I needed the `#[serde(untagged)]` attribute on top of the enum to work as expected.

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    if value.is_null() {
+        null_mut()
+    } else {
+        odbc_api::handles::SqlText::new(value.as_str().expect("Unable to cast value as string"))
+            .ptr()
+    }
+}
+
+fn to_i16(value: &Value) -> i16 {
+    value.as_i64().expect("Unable to cast value as i64") as i16
```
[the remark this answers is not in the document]
bucaojit: I added a try_from() that will error on overflow.

integration_test/src/test_runner/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+                    to_i16(&function[6]),
+                    str_or_null(&function[7]),
+                    to_i16(&function[8]),
+                ))
+            }
+        }
+        /*
+        "sqlprimarykeys" => {
+            unsafe {
+                Ok(odbc_sys::SQLPrimaryKeys(
```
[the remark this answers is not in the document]
bucaojit: I'll bring up the missing `SQLPrimaryKeys()` function during standup to see how to handle testing this. 

I added the `SQLForeignKeysW()` function which is in odbc-sys, although the W version in case we want to test it.

integration_test/src/test_runner/mod.rs
bucaojit: I changed this to only accept `SqlReturn::SUCCESS`.  I see now that `SUCCESS_WITH_INFO` is success with a warning message, that's a bit confusing.  I haven't seen a `SUCCESS_WITH_INFO` in my limited testing so I don't think we should expect it as success.

integration_test/src/test_runner/mod.rs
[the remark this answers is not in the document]
bucaojit: I created an investigation ticket to address this ([SQL-1015](https://jira.mongodb.org/browse/SQL-1015))

evergreen.yml
```diff
@@ -247,6 +247,21 @@ functions:
           # we only run asan on the unit tests for now
           cargo test --target x86_64-unknown-linux-gnu unit
 
+  "run result set test":
```
mattChiaravalloti, line 250: Since the data loader didn't have any test tasks to run, I left evergreen changes out of #15 . Can you update the relevant evergreen task(s) to include `cargo run data_loader` to run the data loader after ADF starts?

mattChiaravalloti (approved):
Still looks good, but we should update this with data loader changes before merging it!

resources/integration_test/tests/functions.yaml
```diff
@@ -0,0 +1,5 @@
+tests:
+  - description: stub function test
+    skip_reason: "SQL-987: Create ODBC Result Set function tests"
+    test_definition: [  ]
+    db: integration_test
```
nbagnard, line 5: FYI: There is a missing new line at the end of the file

nbagnard (approved):

evergreen.yml
```diff
@@ -292,6 +315,18 @@ task_groups:
     tasks:
       - test-rust
 
+  - name: test-result-set-group
```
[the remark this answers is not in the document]
bucaojit: I renamed it because Patrick's connection PR added a `test-integration-group` [here](https://github.com/mongodb/mongo-odbc-driver/blob/e8dcba32f8afe87ee9ddc521a3522bcc2a2347c4/evergreen.yml#L296), which was the conflict I mentioned during standup.   
One idea could be to run them together in the same task.  For now it might be good to run them separately, as that may help with troubleshooting them.  `test-result-set-group` was the first descriptive name that came to mind, how about`test-harness-group`, `test-query-function-group`?

resources/integration_test/tests/functions.yaml
bucaojit: Added newline 👍

evergreen.yml
bucaojit: Added, thanks for the suggestion.  Also, it looks like the run command has a --bin option: `cargo run --bin data_loader`

evergreen.yml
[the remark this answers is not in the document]
bucaojit: I'll keep the current name and we can modify it in a future PR if necessary.

bucaojit:
evergreen merge

PR #21, opened 2022-09-24
SQL-990: Properly handle the ODBC URI format

pmeredit:
The spec can be found here: https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqldriverconnect-function?view=sql-server-ver16

the grammar suggests that {} is only allowed for the driver, but the text suggests it is allowed for any value, so that's what I went with.

({} is so that you can have special characters in a value... except obviously }).

### DIFF
--- a/odbc/src/api/driver_connect_tests.rs
+++ b/odbc/src/api/driver_connect_tests.rs
@@ -94,7 +94,7 @@ mod unit {
             driver_completion = DriverConnectOption::NoPrompt,
             expected_sql_state = UNABLE_TO_CONNECT,
             expected_sql_return = SqlReturn::ERROR,
-            expected_error_message = "[MongoDB][API] Invalid Uri: One of [\"user\", \"uid\"] is required for a valid Mongo ODBC Uri"
+            expected_error_message = "[MongoDB][API] Invalid Uri: One of [\"uid\", \"user\"] is required for a valid Mongo ODBC Uri"
         );
     test_connection_diagnostics! (
             missing_pwd_in_connection_string,
@@ -102,7 +102,7 @@ mod unit {
             driver_completion = DriverConnectOption::NoPrompt,
             expected_sql_state = UNABLE_TO_CONNECT,
             expected_sql_return = SqlReturn::ERROR,
-            expected_error_message = "[MongoDB][API] Invalid Uri: One of [\"pwd\", \"password\"] is required for a valid Mongo ODBC Uri"
+            expected_error_message = "[MongoDB][API] Invalid Uri: One of [\"password\", \"pwd\"] is required for a valid Mongo ODBC Uri"
         );
     test_connection_diagnostics!(
         missing_driver_in_connection_string,
--- a/odbc/src/api/odbc_uri.rs
+++ b/odbc/src/api/odbc_uri.rs
@@ -36,7 +36,7 @@ pub struct ODBCUri<'a>(HashMap<String, &'a str>);
 impl<'a> ODBCUri<'a> {
     pub fn new(odbc_uri: &'a str) -> Result<ODBCUri<'a>> {
         if odbc_uri.is_empty() {
-            return Err(ODBCError::InvalidUriFormat(NOT_EMPTY_ERROR.to_string()));
+            return Err(ODBCError::InvalidUriFormat(EMPTY_URI_ERROR.to_string()));
         }
         let mut input = odbc_uri;
         let mut ret = ODBCUri(HashMap::new());
--- a/odbc/src/api/odbc_uri.rs
+++ b/odbc/src/api/odbc_uri.rs
@@ -90,7 +90,9 @@ impl<'a> ODBCUri<'a> {
             })?);
         match rest.len() {
             0 => unreachable!(),
+            // Just the "}" remaining
             1 => return Ok((value, None)),
+            // Just "};" remaining
             2 if rest.chars().nth(1).unwrap() == ';' => return Ok((value, None)),
             _ => (),
         }
--- a/odbc/src/api/odbc_uri.rs
+++ b/odbc/src/api/odbc_uri.rs
@@ -42,10 +42,9 @@ impl<'a> ODBCUri<'a> {
         let mut ret = ODBCUri(HashMap::new());
         while let Some((keyword, value, rest)) = ODBCUri::get_next_attribute(input)? {
             // if attributes are repeated, the first is the one that is kept.
-            if ret.0.contains_key(&keyword) {
-                continue;
+            if !ret.0.contains_key(&keyword) {
+                ret.0.insert(keyword, value);
             }
-            ret.0.insert(keyword, value);
             if rest.is_none() {
                 return Ok(ret);
             }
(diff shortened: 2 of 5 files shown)


odbc/src/api/driver_connect_tests.rs
```diff
@@ -94,15 +94,15 @@ mod unit {
             driver_completion = DriverConnectOption::NoPrompt,
             expected_sql_state = UNABLE_TO_CONNECT,
             expected_sql_return = SqlReturn::ERROR,
-            expected_error_message = "[MongoDB][API] Invalid Uri: One of [\"user\", \"uid\"] is required for a valid Mongo ODBC Uri"
```
pmeredit, line 97: I changed the order here to always be alphabetic.

odbc/src/api/odbc_uri.rs
```diff
@@ -19,23 +38,79 @@ impl<'a> ODBCUri<'a> {
...
+                ODBCError::InvalidUriFormat(MISSING_CLOSING_BRACE_ERROR.to_string())
+            })?;
+            ODBCUri::handle_braced_value(rest)?
+        } else {
+            ODBCUri::handle_unbraced_value(rest)?
+        };
+        Ok(Some((keyword.to_lowercase(), value, rest)))
+    }
+
+    fn handle_braced_value(input: &'a str) -> Result<(&'a str, Option<&'a str>)> {
```
pmeredit, line 86: I'm actually not sure now if we should, perhaps, allow nested {}. What do you guys think? It wouldn't be that hard to support since the braced_value is its own function, but it isn't at all clear from the spec if we should.

@nbagnard, what do you think?

odbc/src/api/odbc_uri.rs
```diff
@@ -80,13 +154,8 @@ impl<'a> ODBCUri<'a> {
     pub fn remove_to_mongo_uri(&mut self) -> Result<String> {
         let user = self.remove_mandatory_attribute(USER)?;
         let pwd = self.remove_mandatory_attribute(PWD)?;
-        // TODO SQL-990: Support the PORT attribute, right now the only way to specify PORT is as
```
pmeredit, line 83: Decided not to support PORT for now. We can add that later if anyone asks for it.

odbc/src/api/odbc_uri.rs
```diff
@@ -19,23 +38,75 @@ impl<'a> ODBCUri<'a> {
...
+
+    fn handle_braced_value(input: &'a str) -> Result<(&'a str, Option<&'a str>)> {
+        let (value, rest) =
+            input.split_at(input.find('}').ok_or_else(|| {
+                ODBCError::InvalidUriFormat(MISSING_CLOSING_BRACE_ERROR.to_string())
+            })?);
+        match rest.len() {
+            0 => unreachable!(),
+            1 => return Ok((value, None)),
+            2 if rest.chars().nth(1).unwrap() == ';' => return Ok((value, None)),
```
bucaojit, line 94: This got me wondering whether we need to support white space in the connection string, such as if the user had an attribute like `UID={User} ;`.  If it's not something powerBI does then that would not need to be done but something to consider.  It doesn't specify in the SQLDriverConnect spec, however here's one entry for OLE DB that indicates it is supported: 
https://learn.microsoft.com/en-us/openspecs/sql_server_protocols/ms-oledbstr/880aa5bf-33c8-4a90-9dbf-e738880cad98

odbc/src/api/odbc_uri.rs
```diff
@@ -19,23 +38,75 @@ impl<'a> ODBCUri<'a> {
...
-                }
-                // ODBC attribute keys are case insensitive, so we lowercase the keys
-                Ok((sp.remove(0).to_lowercase(), sp.remove(0)))
-            })
-            .collect::<Result<HashMap<_, _>>>()
-            .map(ODBCUri)
+        let mut input = odbc_uri;
+        let mut ret = ODBCUri(HashMap::new());
+        while let Some((keyword, value, rest)) = ODBCUri::get_next_attribute(input)? {
+            ret.0.insert(keyword, value);
```
bucaojit, line 44: The spec outlines that if keywords are repeated, then the value for the first occurrence of the keyword will be used.  We may want to add a check to only insert if the keywords doesn't already exist here.

bucaojit (changes requested):
It was really interesting seeing how you handled iterating through the attributes 👍.

odbc/src/api/odbc_uri.rs
bucaojit: I did not see it specified in the spec, I vote for not supporting it unless a customer asks.

mattChiaravalloti: > A DSN or connection string value enclosed with braces ({}) containing any of the characters []{}(),;?*=!@ is passed intact to the driver. However, when using these characters in a keyword, the Driver Manager returns an error when working with file DSNs but passes the connection string to the driver for regular connection strings. Avoid using embedded braces in a keyword value.

To me, this implies attribute values _can_ include `{}` so we should be supporting them. Maybe I'm misinterpreting this spec, though.

odbc/src/api/odbc_uri.rs
mattChiaravalloti, line 94: These two took me a reread to understand. Maybe we can add a comment like we have in the code above.

```suggestion
            // Just the "}" remaining
            1 => return Ok((value, None)),
            // Just "};" remaining
            2 if rest.chars().nth(1).unwrap() == ';' => return Ok((value, None)),
```
something like that ^

mattChiaravalloti (changes requested):
nice changes, easy to read and follow

odbc/src/api/odbc_uri.rs
pmeredit: I think you are right!

odbc/src/api/odbc_uri.rs
pmeredit: It can still be supported using `{}`, so I'm not super worried about it.

odbc/src/api/odbc_uri.rs
```diff
@@ -1,15 +1,34 @@
 use crate::errors::{ODBCError, Result};
+use lazy_static::lazy_static;
+use regex::{RegexSet, RegexSetBuilder};
 use std::collections::HashMap;
 
-// TODO SQL-990: These errors will probably change.
-const NOT_EMPTY_ERROR: &str = "uri must not be empty";
-const EQUAL_ERROR: &str = "all uri atttributes must be of the form key=value";
+const NOT_EMPTY_ERROR: &str = "URI must not be empty";
```
pmeredit, line 6:
```suggestion
const EMPTY_URI_ERROR: &str = "URI must not be empty";
```

odbc/src/api/odbc_uri.rs
```diff
@@ -19,23 +38,77 @@ impl<'a> ODBCUri<'a> {
         if odbc_uri.is_empty() {
             return Err(ODBCError::InvalidUriFormat(NOT_EMPTY_ERROR.to_string()));
```
pmeredit, line 39:
```suggestion
            return Err(ODBCError::InvalidUriFormat(EMPTY_URI_ERROR.to_string()));
```

odbc/src/api/odbc_uri.rs
```diff
@@ -1,41 +1,129 @@
 use crate::errors::{ODBCError, Result};
+use lazy_static::lazy_static;
+use regex::{RegexSet, RegexSetBuilder};
 use std::collections::HashMap;
 
-// TODO SQL-990: These errors will probably change.
-const NOT_EMPTY_ERROR: &str = "uri must not be empty";
-const EQUAL_ERROR: &str = "all uri atttributes must be of the form key=value";
+const EMPTY_URI_ERROR: &str = "URI must not be empty";
+const EQUAL_ERROR: &str = "all URI attributes must be of the form keyword=value";
```
mattChiaravalloti, line 7:
```suggestion
const INVALID_ATTR_FORMAT_ERROR: &str = "all URI attributes must be of the form keyword=value";
```
A rename for this error could be useful, too.

bucaojit:
One thing to note: for the failing integration-test if you update [this GO_VERSION ](https://github.com/mongodb/mongo-odbc-driver/blob/832a7d73acd82afc3f5183148c04de143253411e/resources/run_adf.sh#L27)to 1.18, that should get the test to succeed again.

pmeredit:
evergreen merge

PR #27, opened 2022-10-11
SQL-654: Implement SQLGetConnectAttr and SQLSetConnectAttr

mattChiaravalloti:
This PR implements `SQLGetConnectAttr` and `SQLSetConnectAttr`. This ticket mentioned that PowerBi _only uses_ `SQLSetConnectAttr(SQL_ATTR_LOGIN_TIMEOUT)`, so for the `Set` function I only implemented that attribute. It would be trivial to implement the other two, so let me know if that should be done. I implemented all three attributes for `Get`.

I'm still a little unsure if what I did here is what is expected/intended. Also, how/where could/should I test this? Or are we leaving testing to future tickets?

### DIFF
--- a/odbc/src/api/util.rs
+++ b/odbc/src/api/util.rs
@@ -1,0 +1,28 @@
+use odbc_sys::ConnectionAttribute;
+
+pub(crate) fn connection_attribute_to_string(attr: ConnectionAttribute) -> String {
+    match attr {
+        ConnectionAttribute::AsyncEnable => "ASYNC_ENABLE".to_string(),
+        ConnectionAttribute::AccessMode => "ACCESS_MODE".to_string(),
+        ConnectionAttribute::AutoCommit => "AUTO_COMMIT".to_string(),
+        ConnectionAttribute::LoginTimeout => "LOGIN_TIMEOUT".to_string(),
+        ConnectionAttribute::Trace => "TRACE".to_string(),
+        ConnectionAttribute::TraceFile => "TRACE_FILE".to_string(),
+        ConnectionAttribute::TranslateLib => "TRANSLATE_LIB".to_string(),
+        ConnectionAttribute::TranslateOption => "TRANSLATE_OPTION".to_string(),
+        ConnectionAttribute::TxnIsolation => "TXN_ISOLATION".to_string(),
+        ConnectionAttribute::CurrentCatalog => "CURRENT_CATALOG".to_string(),
+        ConnectionAttribute::OdbcCursors => "ODBC_CURSORS".to_string(),
+        ConnectionAttribute::QuietMode => "QUIET_MODE".to_string(),
+        ConnectionAttribute::PacketSize => "PACKET_SIZE".to_string(),
+        ConnectionAttribute::ConnectionTimeout => "CONNECTION_TIMEOUT".to_string(),
+        ConnectionAttribute::DisconnectBehaviour => "DISCONNECT_BEHAVIOUR".to_string(),
+        ConnectionAttribute::AsyncDbcFunctionsEnable => "ASYNC_DBC_FUNCTIONS_ENABLE".to_string(),
+        ConnectionAttribute::AsyncDbcEvent => "ASYNC_DBC_EVENT".to_string(),
+        ConnectionAttribute::EnlistInDtc => "ENLIST_IN_DTC".to_string(),
+        ConnectionAttribute::EnlistInXa => "ENLIST_IN_XA".to_string(),
+        ConnectionAttribute::ConnectionDead => "CONNECTION_DEAD".to_string(),
+        ConnectionAttribute::AutoIpd => "AUTO_IPD".to_string(),
+        ConnectionAttribute::MetadataId => "METADATA_ID".to_string(),
+    }
+}


odbc/src/api/functions.rs
```diff
@@ -1948,16 +1991,39 @@ pub unsafe extern "C" fn SQLRowCount(
 /// [`SQLSetConnectAttr`]: https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/SQLSetConnectAttr-function
 ///
 /// # Safety
-/// Because this is a C-infereface, this is necessarily unsafe
+/// Because this is a C-interface, this is necessarily unsafe
 ///
 #[no_mangle]
 pub unsafe extern "C" fn SQLSetConnectAttr(
```
pmeredit, line 1997: I think we might as well support setting the current catalog since it could prove useful for testing (and eventually we'll want it anyway).

odbc/src/api/functions.rs
```diff
@@ -1095,17 +1096,59 @@ pub unsafe extern "C" fn SQLFreeStmt(_statement_handle: HStmt, _option: SmallInt
 /// [`SQLGetConnectAttr`]: https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/SQLGetConnectAttr-function
 ///
 /// # Safety
-/// Because this is a C-infereface, this is necessarily unsafe
+/// Because this is a C-interface, this is necessarily unsafe
 ///
 #[no_mangle]
 pub unsafe extern "C" fn SQLGetConnectAttr(
```
pmeredit, line 1102: Leave `SQLGetConnectAttr` unimplemented and move this code to `SQLGetConnectAttrW` since you are outputting *WChar

odbc/src/api/functions.rs
pmeredit: Also move this code to `SQLSetConnectAttrW` for reasons as above. It will matter once you support setting the current catalog since that is going to be passed as a `*WChar`.

My latest patch also adds support for ASCII strings, but we don't need to support _accepting_ ASCII strings yet, imo.

mattChiaravalloti:
Marked this as a draft since it is not currently ready for review. I have a few more questions that came up I think Patrick will be able to help me with tomorrow and then this should move forward.

mattChiaravalloti:
Ok, this is ready now! Code has been cleaned up and tests have been added. I'll rebase this if Patrick's #26 gets merged first since I copied some of his utility methods.

odbc/src/api/functions.rs
```diff
@@ -1966,23 +2013,62 @@ pub unsafe extern "C" fn SQLSetConnectAttr(
...
+    // all operating systems, and the docs say it can panic.
+    let sql_return = {
+        let conn = must_be_valid!((*conn_handle).as_connection());
+        let mut conn_guard = conn.write().unwrap();
+
+        match attribute {
+            ConnectionAttribute::CurrentCatalog => {
+                let current_catalog =
+                    input_wtext_to_string(value_ptr as *const WChar, str_length as usize);
+                conn_guard.attributes.current_catalog = Some(current_catalog);
```
nbagnard, line 2103: If the catalog name is a not a database the user has access to, then this should return SQL_ERROR and 3D000-Invalid catalog name.

odbc/src/api/connect_attr_tests.rs
```diff
@@ -0,0 +1,281 @@
...
+        util::input_wtext_to_string,
+        SQLGetConnectAttrW, SQLSetConnectAttrW,
+    };
+    use odbc_sys::{ConnectionAttribute, Integer, Pointer, SqlReturn, UInteger};
+    use std::sync::RwLock;
+
+    // Test getting CurrentCatalog attribute from the Connection
+    // when none of the attributes are set.
+    #[test]
+    fn get_string_attrs_default() {
```
nbagnard, line 14: [note] I am not a big fan of the code duplication here. I think Oliver's work on the helpers could help with refactoring it later.

odbc/src/api/functions.rs
```diff
@@ -2032,23 +2077,62 @@ pub unsafe extern "C" fn SQLSetConnectAttr(
...
+                SqlReturn::SUCCESS
+            }
+            ConnectionAttribute::LoginTimeout => match FromPrimitive::from_u32(value_ptr as u32) {
+                Some(login_timeout) => {
+                    conn_guard.attributes.login_timeout = Some(login_timeout);
+                    SqlReturn::SUCCESS
+                }
+                None => {
+                    err = Some(ODBCError::InvalidAttrValue("SQL_ATTR_LOGIN_TIMEOUT"));
```
nbagnard, line 2112:
```suggestion
                    err = Some(ODBCError::InvalidAttrValue(attribute));
```

nbagnard (changes requested):
Overall it looks good and will do what it should! 
The things I've commented on are mostly related to spec details or the functionality associated with the attribute value change.
I tried to go through the spec and check all the details, but I may still have forgotten some corner cases. I'll give it another thorough look next round.

odbc/src/api/functions.rs
```diff
@@ -1966,23 +2013,62 @@ pub unsafe extern "C" fn SQLSetConnectAttr(
...
+        match attribute {
+            ConnectionAttribute::CurrentCatalog => {
+                let current_catalog =
+                    input_wtext_to_string(value_ptr as *const WChar, str_length as usize);
+                conn_guard.attributes.current_catalog = Some(current_catalog);
+                SqlReturn::SUCCESS
+            }
+            ConnectionAttribute::LoginTimeout => match FromPrimitive::from_u32(value_ptr as u32) {
+                Some(login_timeout) => {
+                    conn_guard.attributes.login_timeout = Some(login_timeout);
```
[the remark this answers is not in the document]
mattChiaravalloti: Oh wait, the spec says:
> An SQLUINTEGER value corresponding to the number of seconds to wait for a login request to complete before returning to the application. The default is driver-dependent. If ValuePtr is 0, the timeout is disabled and a connection attempt will wait indefinitely.
>
> If the specified timeout exceeds the maximum login timeout in the data source, the driver substitutes that value and returns SQLSTATE 01S02 (Option value changed).

So the value argued to the function is supposed to be unsigned. Given that, isn't it not possible to get a negative number? If a user passes in a negative i32 value, isn't that their mistake for using the api incorrectly?

odbc/src/api/functions.rs
```diff
@@ -1966,23 +2013,62 @@ pub unsafe extern "C" fn SQLSetConnectAttr(
...
+
+    // This scope is introduced to make the RWLock Guard expire before we write
+    // any error values via add_diag_info as RWLock::write is not reentrant on
+    // all operating systems, and the docs say it can panic.
+    let sql_return = {
+        let conn = must_be_valid!((*conn_handle).as_connection());
+        let mut conn_guard = conn.write().unwrap();
+
+        match attribute {
+            ConnectionAttribute::CurrentCatalog => {
```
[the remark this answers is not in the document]
mattChiaravalloti: discussed offline -- deferring this work until after the epic since it is not required to get PowerBI working. Followup ticket: https://jira.mongodb.org/browse/SQL-1075

odbc/src/api/functions.rs
mattChiaravalloti: discussed offline -- deferring this work until after the epic since it is not required to get PowerBI working. Followup ticket: https://jira.mongodb.org/browse/SQL-1075

odbc/src/api/functions.rs
```diff
@@ -1966,23 +2013,62 @@ pub unsafe extern "C" fn SQLSetConnectAttr(
...
 ///
 #[no_mangle]
 pub unsafe extern "C" fn SQLSetConnectAttrW(
-    _hdbc: HDbc,
-    _attr: ConnectionAttribute,
-    _value: Pointer,
-    _str_length: Integer,
+    connection_handle: HDbc,
+    attribute: ConnectionAttribute,
+    value_ptr: Pointer,
```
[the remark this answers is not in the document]
mattChiaravalloti: Since we deferred setting the CURRENT_CATALOG for now and that is the only string valued attribute, this issue has gone away for now.

odbc/src/api/functions.rs
```diff
@@ -1114,24 +1116,69 @@ pub unsafe extern "C" fn SQLGetConnectAttr(
...
+                        cc,
+                        value_ptr as *mut WChar,
+                        buffer_length as usize,
+                        string_length_ptr as *mut SmallInt,
+                    ),
+                }
+            }
+            ConnectionAttribute::LoginTimeout => {
+                let login_timeout = attributes.login_timeout.unwrap_or(0);
+                set_output_fixed_data(&login_timeout, value_ptr, string_length_ptr as *mut Len)
```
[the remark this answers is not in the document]
mattChiaravalloti: Good point. I'm not sure what we should do here since the spec doesn't say what to do if the value_ptr is null and the value is not a string. What do you think we should do in this case?

odbc/src/api/connect_attr_tests.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    // Test setting CurrentCatalog attribute.
+    #[test]
+    fn set_string_attrs() {
+        unsafe {
+            let conn = Connection::with_state(std::ptr::null_mut(), ConnectionState::Connected);
+            let mongo_handle: *mut _ = &mut MongoHandle::Connection(RwLock::new(conn));
+
+            let mut value = "test".encode_utf16().collect::<Vec<u16>>();
+            value.push('\u{0}' as u16);
```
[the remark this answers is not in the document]
mattChiaravalloti: This is also gone away since we are not setting strings at this time.

odbc/src/api/connect_attr_tests.rs
mattChiaravalloti: To be clear, are you suggesting using the `verify_sql_diagnostics` helper method?

odbc/src/api/functions.rs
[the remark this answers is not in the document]
pmeredit: This seems like bizarre behavior to me. The current catalog should have no affect on outstanding cursors. Do we really have to support that behavior? It seems fine to me to just not error and have it work.

pmeredit (approved):
LGTM outside of the edge cases @nbagnard brought up

odbc/src/api/connect_attr_tests.rs
mattChiaravalloti: I investigated how to use that function but it required quite a bit of overhead to even make it available in the tests, and even then it was as cleanly applicable as I would have expected. I did do a bunch of major refactoring on the tests to cut down on the duplication. It abstracts away a lot of the nitty-gritty details from the tests themselves, which _should_ be more useful? lmk what you think!

nbagnard: uhm, yes, it looks like Oliver's helper are meant more for the integration tests than the unit tests. 
I was more thinking of a generic function for initializing and handling the recurrent code for both get and set which is exactly what you did 👍

odbc/src/api/functions.rs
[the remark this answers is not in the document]
nbagnard: I always try to give them credit for doing things for a reason. I can't find something that makes me want to fight for this behavior.
Maybe if you are in prepare/execute flow and the change happen in between them? Let's you prepare on DB1, you get the resultset metadata, bind your resultset column or just simply prepare your buffer for retrieving the data, switch the catalog to DB2 which has the same table but with a different schema, then you execute and then when retrieving everything is either looking weird/incorrect or failing?
Let talk more about this when we get to SQL-1075.

odbc/src/api/functions.rs
nbagnard: What about `ODBCError::UnsupportedConnectionAttribute(attribute)` below then? Does it mean that the error message is going to be "The feature {Attribute index} is not implemented"?
I don't see a test checking the error message itself, but this is what I'd expect.

nbagnard (changes requested):
Can you please just check the error message for unsupported attribute and decide+implement what to do for Null pointer for the login timeout?
Otherwise, it looks good to me :)
👍

odbc/src/api/connect_attr_tests.rs
```diff
@@ -0,0 +1,254 @@
...
+        expected_value = 42u32,
+        actual_value_modifier = modify_numeric_attr,
+    );
+
+    // Test getting unsupported attributes.
+    #[test]
+    fn get_unsupported_attr() {
+        unsafe {
+            for attr in [
+                ConnectionAttribute::AsyncEnable,
```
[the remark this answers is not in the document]
mattChiaravalloti: The lists are different since `Set`'s list contains `CurrentCatalog` and `ConnectionTimeout`.

I can make the rest of the list a constant except for those two, and then for the `Set` test I can append the extra ones, so this will remove the duplication between parts of the two lists but waring that it won't be a wholesale replacement!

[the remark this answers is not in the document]
mattChiaravalloti: done

mattChiaravalloti:
Still trying to debug the asan failure 😞 I narrowed it down to the `get_attr` tests, but I've tried several different things on separate patches (including manipulating the initial buffer sizes, the initial length values, the types of the arguments to `set_output_wstring` and `set_output_fixed_data` to `Integer` instead of `Len`, and ensuring I use the same `let _ = Box::from_raw(<input buffer>);` like we do in other unit tests). Nothing has worked so far and I'm not quite sure what to do next to get to the bottom of this.

mattChiaravalloti:
Ok, actually, I think I've gotten to the bottom of it with this patch: https://spruce.mongodb.com/task/mongosql_odbc_driver_ubuntu2004_asan_patch_ac4701e11a668fec75e3da515a3422f5a351a661_63517bac5623433bdd4b9e88_22_10_20_16_47_52/logs?execution=0

This is actually something I had already tried but I only just understood it this time. This patch shows we have the following memory leaks:
2 of 200 bytes,
3 of 40 bytes,
5 of 10 bytes

These line up with test tests in `data_tests.rs`.

Before these changes, the problems lined up with the tests in `connect_attr_tests.rs`.

The issue _seems_ to be that `connect_attr_tests.rs` need to use `*mut Integer` lengths, whereas `data_tests.rs` need to use `*mut Len` lengths, and I tried forcing them to _both_ use the `i32_len` module. I'm going to change `i32_len` to actually use `Integer` (an i32) and create another new module called `isize_len` that will be used by the `SQLGetData` codepaths. I _think_ this should fix this. Will test and resubmit if it does the trick.

mattChiaravalloti:
confirmed 🎉  that did the trick, this is ready for another look @nbagnard ! The last two remaining threads are about invalid values for non-string attributes. I left responses/questions on both threads.

odbc/src/api/functions.rs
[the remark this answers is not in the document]
nbagnard: The trick to know what the DM, is to try it out 😅 You don't have to use our driver though, any ODBC driver would work, but you need to install one and call the ODBC API directly, you can't use unit test for this.
Otherwise, yes, the spec is not clear, so we should be able to pick whichever solution, since no one can tell us that we are "wrong" 😜

odbc/src/api/functions.rs
nbagnard: You are right, ODBCError::UnsupportedConnectionAttribute is the next line. Both are linked, but it explains the confusion!
Since  you said that when using ODBCError::InvalidAttrValue(attribute) it would use the int32 value from the enum and this was why you had to use the string value, I was asking if this was not also be a problem with ODBCError::UnsupportedConnectionAttribute(attribue).
This was originally a question because I did not know if there was something specific done for UnsupportedConnectionAttribute that I missed or not. You just answered this now. Thank you :) 
Can we make it so that it takes a string then?

odbc/src/api/functions.rs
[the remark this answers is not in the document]
nbagnard: This goes away with the new code too anyway 🎉

nbagnard (approved):
LGTM after you update the code to pass in the String representation of the invalid attribute instead of its int32 "code" 💯 
And I'm glad you figured out the issue with ASAN! 👍

mattChiaravalloti:
evergreen merge

mattChiaravalloti:
evergreen merge

PR #33, opened 2022-11-08
SQL-984: Test Harness: Implement Result Set validator

bucaojit:
Implemented the query and function result set validation.  
There is still work to validate the column metadata, I will create a ticket to cover that. 
I have tested retrieving null and string data and added the query and function tests in `resources/integration_test/tests/queries.yaml|functions.yaml`.  I will test the numeric types before I merge. 
One area that could be improved is if there is a way to run the tests as a separate `#[test]` like in the connection tests [here](https://github.com/mongodb/mongo-odbc-driver/blob/master/odbc/tests/connection_tests.rs).  The issue is that the tests are run from integration_test() function, open to any suggestions.

### DIFF
--- a/integration_test/src/test_runner/mod.rs
+++ b/integration_test/src/test_runner/mod.rs
@@ -193,15 +193,12 @@ fn run_query_test(
             to_wstr_ptr(query),
             query.len() as i32,
         ) {
-            SqlReturn::SUCCESS => {}
-            sql_return => {
-                return Err(Error::OdbcFunctionFailed(
-                    "SQLExecDirectW".to_string(),
-                    format!("{:?}", sql_return),
-                ))
-            }
+            SqlReturn::SUCCESS => validate_result_set(entry, stmt),
+            sql_return => Err(Error::OdbcFunctionFailed(
+                "SQLExecDirectW".to_string(),
+                format!("{:?}", sql_return),
+            )),
         }
-        validate_result_set(entry, stmt)
     }
 }
 
@@ -371,8 +368,8 @@ fn validate_result_set(
                     });
                 }
 
-                for i in 1..(columns + 1) {
-                    let expected_field = expected_row.get(i - 1).unwrap();
+                for i in 0..(columns) {
+                    let expected_field = expected_row.get(i).unwrap();
                     let expected_data_type = if expected_field.is_number() {
                         CDataType::SLong
                     } else {
@@ -414,7 +411,8 @@ fn get_data(
     unsafe {
         match odbc_sys::SQLGetData(
             stmt.handle() as *mut _,
-            column,
+            // Result set columns start at 1, the column input parameter is 0-indexed
+            column + 1,
             data_type,
             buffer as *mut _,
             BUFFER_LENGTH as isize,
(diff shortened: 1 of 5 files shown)


pmeredit (approved):
LGTM!

integration_test/src/test_runner/mod.rs
```diff
@@ -207,13 +181,36 @@ fn check_array_length(array: &Vec<Value>, length: usize) -> Result<(), Error> {
...
+        ) {
+            SqlReturn::SUCCESS => {}
+            sql_return => {
+                return Err(Error::OdbcFunctionFailed(
+                    "SQLExecDirectW".to_string(),
+                    format!("{:?}", sql_return),
+                ))
+            }
+        }
+        validate_result_set(entry, stmt)
```
mattChiaravalloti, line 204:
```suggestion
            SqlReturn::SUCCESS => validate_result_set(entry, stmt)
            sql_return => {
                return Err(Error::OdbcFunctionFailed(
                    "SQLExecDirectW".to_string(),
                    format!("{:?}", sql_return),
                ))
            }
        }
```
I may be misunderstanding this function, but it seems we only validate the result set if we get `SUCCESS`, so why not have this function live within the match statement instead of after it?

integration_test/src/test_runner/mod.rs
```diff
@@ -322,47 +339,134 @@ fn run_function_test(
...
+            if let Some(expected_row) = expected_row_check {
+                if expected_row.len() != columns {
+                    return Err(Error::ColumnCount {
+                        test: entry.description.clone(),
+                        expected: expected_row.len(),
+                        actual: columns,
+                    });
+                }
 
+                for i in 1..(columns + 1) {
```
mattChiaravalloti, line 374:
```suggestion
                // Recall that get_data is 1-indexed, but of course the expected_row is 0-indexed since
                // it is a Rust vec. That is why we iterate from 1 to columns+1 and why we must subtract 1
                // when getting from expected_row but not when calling get_data.
                for i in 1..(columns + 1) {
```
Not sure this comment is the best phrasing, but I think a comment could be helpful here just to clarify why we have to do `i-1` when getting expected and not when getting actual.

mattChiaravalloti: Arguably, we _could_ make `get_data` effectively 0-indexed by having that method handle the `+1` and letting this loop just go from 0 to columns and use `i` directly. The only benefit of doing it that way is that we would be able to just use `i` when retrieving both actual and expected.

integration_test/src/test_runner/mod.rs
bucaojit: That's a good point, thanks for catching that.  I moved the function into the match, and it looks cleaner 👍 .  
I also modified the other functions that had the same pattern to move the functionality into the no-op, I must have written this before I had my coffee ☕

integration_test/src/test_runner/mod.rs
bucaojit: Sounds good.  I updated `get_data` to handle the `+1` and added a comment to explain why.

bucaojit:
evergreen merge

PR #43, opened 2022-11-18
SQL-1109: Finish SQLGetInfoW implementation

mattChiaravalloti:
This PR uncomments the last `SQLGetInfoW` match case. As we know, this results in integration test memory failures, but these seem to be unrelated to this code directly so we could just override dependencies and get this into master anyway.


nbagnard (approved):
LGTM!

mattChiaravalloti:
evergreen merge

PR #49, opened 2022-11-30
SQL-661: Initial support for descriptor handles

jchemburkar:
Added initial infrastructure for descriptor handles. There are some pieces in the handle struct that aren't strictly necessary, but felt right given how other handles function and potential future additions (specifically, a reference to the dbc and a `state`).


pmeredit (approved):
This looks great! As a side note, we may actually be removing the state stuff, but it's good to have it for now (we think it's not something we actually have the track because it seems like the driver manager may actually do it, but we still aren't 100% sure).

pmeredit:
evergreen merge

PR #51, opened 2022-12-05
SQL-639: PowerBi connection test

nbagnard:


odbc/src/api/definitions.rs
```diff
@@ -111,131 +111,194 @@ pub enum UseBookmarks {
...
+    SQL_DROP_TRANSLATION = 142,
+    SQL_DROP_VIEW = 143,
+    SQL_DYNAMIC_CURSOR_ATTRIBUTES1 = 144,
+    SQL_DYNAMIC_CURSOR_ATTRIBUTES2 = 145,
+    SQL_FORWARD_ONLY_CURSOR_ATTRIBUTES1 = 146,
+    SQL_FORWARD_ONLY_CURSOR_ATTRIBUTES2 = 147,
+    SQL_INDEX_KEYWORDS = 148,
+    SQL_INFO_SCHEMA_VIEWS = 149,
+    SQL_KEYSET_CURSOR_ATTRIBUTES1 = 150,
+    SQL_KEYSET_CURSOR_ATTRIBUTES2 = 151,
```
pmeredit, line 274: Also a good change.

odbc/src/api/functions.rs
```diff
@@ -1840,448 +1840,425 @@ pub unsafe extern "C" fn SQLGetInfoW(
...
+                InfoType::SQL_SYSTEM_FUNCTIONS => {
+                    // MongoSQL does not support any of the ODBC system functions.
+                    i16_len::set_output_fixed_data(&SQL_U32_ZERO, info_value_ptr, string_length_ptr)
+                }
+                InfoType::SQL_TIMEDATE_FUNCTIONS => {
+                    // MongoSQL supports the following timedate functions.
+                    const TIMEDATE_FUNCTIONS: u32 = SQL_FN_TD_CURRENT_TIMESTAMP
+                        | SQL_FN_TD_EXTRACT;
+                    i16_len::set_output_fixed_data(&TIMEDATE_FUNCTIONS, info_value_ptr, string_length_ptr)
+                }
```
pmeredit, line 2017: [nit] We don't need these comments anymore after the enum name change :D

odbc/src/handles/definitions.rs
```diff
@@ -237,8 +237,8 @@ pub struct Statement {
 
 #[derive(Debug)]
 pub struct StatementAttributes {
-    pub app_row_desc: Pointer,
-    pub app_param_desc: Pointer,
+    pub app_row_desc: *mut MongoHandle,
```
pmeredit, line 240: this definitely makes this easier to work with.

pmeredit (approved):
LGTM other than the small nit :)

evergreen.yml
```diff
@@ -202,7 +246,7 @@ functions:
           export RUST_BACKTRACE=1
           ./resources/run_adf.sh start &&
           cargo run --bin data_loader &&
-          cargo test integration
+          cargo test integration -- --nocapture
```
mattChiaravalloti, line 249: thank you 🙏

integration_test/tests/power_bi_tests.rs
```diff
@@ -0,0 +1,231 @@
...
+                SqlReturn::SUCCESS,
+                SQLGetInfoW(
+                    conn_handle as HDbc,
+                    InfoType::DbmsVer,
+                    output_buffer as Pointer,
+                    BUFFER_LENGTH,
+                    str_len_ptr
+                )
+            );
+            println!(
```
mattChiaravalloti, line 221: Do we want to assert anything about these values (DBMS_NAME and VER) instead of just printing them? Or since they could vary (specifically version) are we ok just printing them and manually checking the values are correct?

mattChiaravalloti (approved):
lgtm 👍

odbc/src/api/functions.rs
nbagnard: You are right, I'll remove them!

integration_test/tests/power_bi_tests.rs
nbagnard: The idea is that we already have/will have unit tests which are validating the values so there is no need to re-check it again.
I print them out to make sure that using the built lib and the DM we get something out and that the buffer length and data match and that in case something looks wrong, we have more trace rather than less to go with when investigating.

odbc/src/api/functions.rs
nbagnard: Done

nbagnard:
I've pushed the changes Oliver to run_adf.sh and evg.yml so that there is not more tasks hanging. The issue was that ADF started OK but Ubuntu 20.04 does not have 'netstat' so the check would fail, and then Evergreen would hang because ADF was still running in the background since we did not know it was, we were not stopping it. Oliver also added a task to copy the log so that we get more context.

The other change here is that I added functions to convert the sql return code to a string and get the diagnostic message when there is a failure and add that to the error the Integration test returns when something does not work as expected. This way instead of 
```
thread 'test_runner::integration_test' panicked at 'assertion failed: `(left == right)`
left: `Ok(())`,
right: `Err(OdbcFunctionFailed("sqltablesw", "SqlReturn(-1)"))`'
```
we get something a little more detailed like:
```
thread 'test_runner::integration_test' panicked at 'assertion failed: `(left == right)`
left: `Ok(())`,
right: `Err(OdbcFunctionFailed("sqltablesw", "ERROR", "[MongoDB][API] The feature SQLGetDiagFieldW is not implemented"))`', integration_test\src\test_runner\mod.rs:166:21

```
The changes are in `integration_test/tests/common/mod.rs` and used in `integration_test/src/test_runner/mod.rs` and `integration_test/tests/power_bi_tests.rs`.

integration_test/tests/common/mod.rs
```diff
@@ -33,3 +34,45 @@ pub fn generate_default_connection_str() -> String {
 
     connection_string
 }
+
+#[allow(dead_code)]
```
mattChiaravalloti, line 38:
```suggestion
#[test]
```
would `#[test]` be more appropriate than `#[allow(dead_code)]` since this could isn't "dead" but instead is used in codepaths guarded by `#[test]`? or maybe i'm losing the thread of how this function is being used!

integration_test/tests/common/mod.rs
```diff
@@ -33,3 +34,45 @@ pub fn generate_default_connection_str() -> String {
...
+    };
+    unsafe {
+        String::from_utf16_lossy(slice::from_raw_parts(
+            actual_message_text as *const WChar,
+            *text_length_ptr as usize,
+        ))
+    }
+}
+
+#[allow(dead_code)]
```
mattChiaravalloti, line 66: same question/answer would apply here! ignore if `#[test]` is not correct.

mattChiaravalloti (approved):
The run_adf changes look good 👍

integration_test/tests/common/mod.rs
nbagnard: ah, this one was a mystery to me.
In lib.rs, this is flagged as code used for tests.
```
#[path = "../tests/common/mod.rs"]
#[cfg(test)]
mod common;
```
 It is also used similarly to generate_default_connection_str.
 So, I don't know what I missed, but it seemed more like a false positive that Clippy was flagging so I put the tag to allow dead code.
 If I missed something and I can clean this up, let me know.
 #[test] does not work because this is not a test, but a helper function used by multiple tests.

mattChiaravalloti: Oh right right, I meant `#[cfg(test)]` 🤦‍♀️  But if the entire module is already marked with that, then I'm not sure adding it at the function level will do anything anyway!

pmeredit (approved):
LGTM!

PR #52, opened 2022-12-07
SQL-1058: Add unit tests for SQLGetInfoW

mattChiaravalloti:
This PR (finally) adds tests for `SQLGetInfoW`. This was deferred from the initial SQL-1058 PR (#32) since that one had already dragged out so long due to the memory issue. This unit tests all InfoTypes added in #32 except for `DbmsVer` since that needs to be an integration test since it requires connecting to ADF to get the version info. ~That integration test is still a work in progress.~ That integration test has been deferred to [SQL-1142](https://jira.mongodb.org/browse/SQL-1142) since it is blocked by the same memory issues as so much other ODBC work.


odbc/src/api/get_info_tests.rs
```diff
@@ -0,0 +1,634 @@
...
+        driver_ver,
+        info_type = InfoType::DriverVer,
+        expected_sql_return = SqlReturn::SUCCESS,
+        buffer_length = 11,
+        expected_length = 10,
+        expected_value = "00.01.0000",
+        actual_value_modifier = modify_string_value,
+    );
+
+    // TODO: change the InfoType variant to correct name and value
```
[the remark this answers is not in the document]
mattChiaravalloti: Semi-discussed offline: not going to do this since there are larger changes related to all of this coming up in a PR soon addressing segfaults.

odbc/src/api/get_info_tests.rs
```diff
@@ -0,0 +1,633 @@
...
+mod unit {
+    use super::*;
+
+    test_get_info!(
+        driver_name,
+        info_type = InfoType::DriverName,
+        expected_sql_return = SqlReturn::SUCCESS,
+        buffer_length = 40,
+        expected_length = 39,
+        expected_value = "MongoDB Atlas SQL interface ODBC Driver",
```
bucaojit, line 84: This string is a good candidate for adding to our `constants` package.

odbc/src/api/get_info_tests.rs
```diff
@@ -0,0 +1,633 @@
...
+        actual_value_modifier = modify_string_value,
+    );
+
+    test_get_info!(
+        dbms_name,
+        info_type = InfoType::DbmsName,
+        expected_sql_return = SqlReturn::SUCCESS,
+        buffer_length = 14,
+        expected_length = 13,
+        expected_value = "MongoDB Atlas",
```
bucaojit, line 124: This one also for the `constants` package.

odbc/src/api/get_info_tests.rs
```diff
@@ -0,0 +1,633 @@
...
+        info_type = InfoType::TimedateFunctions,
+        expected_sql_return = SqlReturn::SUCCESS,
+        expected_length = std::mem::size_of::<u32>() as i16,
+        expected_value = 1572864u32,
+        actual_value_modifier = modify_u32_value,
+    );
+
+    test_get_info!(
+        convert_big_int,
+        info_type = InfoType::ConvertBigInt,
```
bucaojit, line 226: Since this test and the following until `InfoType::ConvertGuid` have the same expected values, I think it would help with readability if we define another macro for these `SQL_U32_ZERO` expected values.  Something like:
```
test_get_info_expect_u32_zero! (convert_big_int, info_type = InfoType::ConvertBigInt);
test_get_info_expect_u32_zero! (convert_decimal, info_type = InfoType::ConvertDecimal);
test_get_info_expect_u32_zero! (convert_double, info_type = info_type = InfoType::ConvertDouble);
...
```

odbc/src/api/get_info_tests.rs
```diff
@@ -0,0 +1,633 @@
...
+use odbc_sys::{Pointer, SmallInt, SqlReturn, UInteger, USmallInt};
+
+macro_rules! test_get_info {
+    ($func_name:ident,
+    info_type = $info_type:expr,
+    expected_sql_return = $expected_sql_return:expr,
+    $(buffer_length = $buffer_length:expr,)?
+    $(expected_length = $expected_length:expr,)?
+    $(expected_value = $expected_value:expr,)?
+    $(actual_value_modifier = $actual_value_modifier:ident,)?
```
bucaojit, line 16: That's a nice solution to get the appropriate value type.

bucaojit (approved):
Tests look good, and elegant solution for converting the output value to match the expected. 

I had a few suggestions that you can implement if you think it will be helpful.

odbc/src/api/get_info_tests.rs
```diff
@@ -0,0 +1,633 @@
...
+        expected_value = "database",
+        actual_value_modifier = modify_string_value,
+    );
+
+    test_get_info!(
+        convert_functions,
+        info_type = InfoType::ConvertFunctions,
+        expected_sql_return = SqlReturn::SUCCESS,
+        expected_length = std::mem::size_of::<u32>() as i16,
+        expected_value = 2u32,
```
nbagnard, line 184: Not a real big deal, but it's hard to tell if the bit flags are correct this way. I'm not saying the value returned itself, I fully trust you checked it, but more the values used to get to this.
Could you express them similarly to how they are defined in sql_get_info?

nbagnard (approved):
Great job checking all the info type. I agree with Oliver, the macro is very neat.
I just have a small suggestion regarding the values tied to bitmasks and how to see what they mean.

odbc/src/api/get_info_tests.rs
mattChiaravalloti: done

odbc/src/api/get_info_tests.rs
mattChiaravalloti: done

odbc/src/api/get_info_tests.rs
mattChiaravalloti: good idea -- done

odbc/src/api/get_info_tests.rs
```diff
@@ -0,0 +1,633 @@
...
+        actual_value_modifier = modify_string_value,
+    );
+
+    test_get_info!(
+        driver_ver,
+        info_type = InfoType::DriverVer,
+        expected_sql_return = SqlReturn::SUCCESS,
+        buffer_length = 11,
+        expected_length = 10,
+        expected_value = "00.01.0000",
```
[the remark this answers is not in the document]
mattChiaravalloti: Yes we will need to change this whenever we release a new version. But that is useful! It will ensure we are actually returning the correct value whenever a new version is released. We could hypothetically calculate the version value at test time using the same code to calculate it in production, but then we wouldn't be testing anything really.

odbc/src/api/get_info_tests.rs
mattChiaravalloti: done

mattChiaravalloti:
evergreen merge

mattChiaravalloti:
evergreen merge

PR #53, opened 2022-12-14
SQL-1056: Add SQLColumns support with tests

pmeredit:
I'm surprised to see all my iteration code actually worked as I wrote it. Desk checking actually is possible, haha. What did not work, however, was decoding the schema result because I forgot our Resultset Schemas are shaped slightly different than our Collection schemas!


core/src/col_metadata.rs
```diff
@@ -104,7 +104,7 @@ pub struct VersionedJsonSchema {
 }
 
 impl SqlGetSchemaResponse {
-    /// Converts a sqlGetResultSchema command response into a list of column
```
pmeredit, line 107: These comments were a result of regex replace, but I like them better this way slightly, hah!

core/src/fields.rs
```diff
@@ -512,15 +512,22 @@ impl MongoFields {
...
+                    let get_schema_cmd = doc! {"sqlGetSchema": collection_name.clone()};
 
                     let db = mongo_connection.client.database(&self.current_db_name);
-                    let current_col_metadata_response: SqlGetSchemaResponse =
+                    let current_col_metadata_response: Result<SqlGetSchemaResponse> =
                         bson::from_document(db.run_command(get_schema_cmd, None).unwrap())
-                            .map_err(Error::BsonDeserialization)
-                            .unwrap();
+                            .map_err(Error::BsonDeserialization);
+                    if current_col_metadata_response.is_err() {
```
pmeredit, line 522: I think it's best to just skip a collection if we fail to deserialize the response, rather than killing iteration.

pmeredit:
Note that the new integration tests pass locally.

core/src/col_metadata.rs
```diff
@@ -87,14 +87,14 @@ impl MongoColMetadata {
     }
 }
 
-// Struct representing the response for a sqlGetResultSchema command.
+// Struct representing the result for a sqlGetResultSchema command.
```
pmeredit, line 90:
```suggestion
// Struct representing the response for a sqlGetResultSchema command.
```

core/src/col_metadata.rs
```diff
@@ -87,14 +87,14 @@ impl MongoColMetadata {
...
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone, Default)]
 pub struct SqlGetSchemaResponse {
     pub ok: i32,
     pub schema: VersionedJsonSchema,
 }
 
-// Auxiliary struct representing part of the response for a sqlGetResultSchema
+// Auxiliary struct representing part of the result for a sqlGetResultSchema
```
pmeredit, line 97:
```suggestion
// Auxiliary struct representing part of the response for a sqlGetResultSchema
```

core/src/col_metadata.rs
```diff
@@ -104,7 +104,7 @@ pub struct VersionedJsonSchema {
 }
 
 impl SqlGetSchemaResponse {
-    /// Converts a sqlGetResultSchema command response into a list of column
+    /// Converts a sqlGetResultSchema command result into a list of column
```
pmeredit, line 107:
```suggestion
    /// Converts a sqlGetResultSchema command response into a list of column
```

core/src/col_metadata.rs
```diff
@@ -179,6 +182,55 @@ impl SqlGetSchemaResponse {
             // 4. Collect as a Vec.
             .collect::<Result<Vec<MongoColMetadata>>>()
     }
+
+    /// Converts a sqlGetSchema command result into a list of column
```
pmeredit, line 186:
```suggestion
    /// Converts a sqlGetSchema command response into a list of column
```

integration_test/tests/query_tests.rs
```diff
@@ -29,4 +29,70 @@ mod integration {
...
+    }
+
+    #[test]
+    fn columns_with_catalog() {
+        let expected = [
+            ["integration_test_2", "example_2", "_id"],
+            ["integration_test_2", "example_2", "b"],
+        ];
+        columns_test(&expected, ("integration_test_2", "", ""));
+    }
```
mattChiaravalloti, line 97: Do you think there would be any value added by including a test that combines filters? Just by reading the code, I have no reason to think this would not work, so the independent filter tests provide sufficient coverage. Not only that, but also we might add more coverage for this and all other functions using the yaml test generator, like in the JDBC repo... The only reason I can think to add a test that has multiple filters is so we don't accidentally regress one day by altering the list_columns code. But still, that can be covered by a generated test.. What do you think -- worth adding `columns_test(&expected, ("integration_test_2", "example_2", "b")` or something like that, or is there no need at this time?

mattChiaravalloti (changes requested):
No big suggestions or questions -- mostly looks great! Marking myself as "requested changes" only because I ended up with 5 comments and that seems like enough to warrant a second look after they're all addressed.

core/src/fields.rs
pmeredit: It would fail silently, but I think that's the best thing to do if somehow there is a junk response from sqlGetSchema.

odbc/src/api/functions.rs
```diff
@@ -572,16 +572,59 @@ pub unsafe extern "C" fn SQLColumnPrivilegesW(
...
+    panic_safe_exec!(
+        || {
+            let mongo_handle = MongoHandleRef::from(statement_handle);
+            let stmt = must_be_valid!((*mongo_handle).as_statement());
+            let catalog_string = input_text_to_string(catalog_name, catalog_name_length as usize);
+            let catalog = if catalog_name.is_null() {
+                None
+            } else {
+                Some(catalog_string.as_str())
+            };
```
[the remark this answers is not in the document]
pmeredit: I don't think this will work, let me try, though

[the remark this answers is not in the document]
pmeredit: Yea, rust is stupid on this. You can't do that because it doesn't live long enough.

integration_test/tests/query_tests.rs
pmeredit: I was thinking it wasn't necessary, but why not. But only one for a filter at every level not some power set :D

core/src/fields.rs
mattChiaravalloti: An alternative to skipping could be to mark the collection as having the "Any" schema. Although that would just defer the problem to other parts of the code, unless we handle this case all the way to the end and report columns from such problematic collections as having "BSON" (unknown) as the datatype.

Yea, I'm convinced skipping (for now) is the best move.

odbc/src/api/functions.rs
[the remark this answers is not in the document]
mattChiaravalloti: I should have realized there would be a problem with the lifetime here. Oh well. The `input_text_to_string` function handles null just fine and we won't "waste" a lot of time on the call either, so this isn't a problem.

mattChiaravalloti (approved):
lgtm 💯

bucaojit (approved):
LGTM, changes look good. 

I had one comment on the integration test.

core/src/fields.rs
pmeredit: Right now it will only fail to deserialize if the schema isn't an object. If it's not an object we don't have columns to report anyway.

core/src/col_metadata.rs
```diff
@@ -179,6 +182,55 @@ impl SqlGetSchemaResponse {
...
+            self.schema.json_schema.clone().try_into()?;
+        let collection_object_schema = collection_schema.assert_datasource_schema()?;
+
+        collection_object_schema
+            // 1. Access collection_schema.properties and sort alphabetically.
+            //    This means we are sorting by field name.
+            .properties
+            .clone()
+            .into_iter()
+            .sorted_by(|a, b| Ord::cmp(&a.0, &b.0))
```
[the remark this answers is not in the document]
pmeredit: Our ordinal positions are defined as a lexicographic sort on the keys, so we actually need this sort to conform to the spec :D

core/src/col_metadata.rs
```diff
@@ -179,6 +182,55 @@ impl SqlGetSchemaResponse {
...
+    ///       "foo": {
+    ///         bsonType: "int",
+    ///       },
+    ///       "bar": {
+    ///         bsonType: "double",
+    ///       }
+    ///   }
+    ///
+    /// produces a list of metadata with the order: "bar", "foo".
+    pub(crate) fn process_collection_metadata(
```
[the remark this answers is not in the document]
pmeredit: Ok, this took a couple hours, but I think I found the best solution to reduce duplicated code, maintain performance, and improve naming.

odbc/src/api/functions.rs
```diff
@@ -595,16 +638,76 @@ pub unsafe extern "C" fn SQLColumns(
...
+        || {
+            let mongo_handle = MongoHandleRef::from(statement_handle);
+            let stmt = must_be_valid!((*mongo_handle).as_statement());
+            let catalog_string = input_wtext_to_string(catalog_name, catalog_name_length as usize);
+            let catalog = if catalog_name.is_null() {
+                None
+            } else {
+                Some(catalog_string.as_str())
+            };
+            // ignore schema
```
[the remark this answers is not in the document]
pmeredit: Oh, great idea!

core/src/fields.rs
```diff
@@ -512,15 +512,22 @@ impl MongoFields {
         loop {
```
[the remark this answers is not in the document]
pmeredit: MongoDB already sorts dbs and collections lexicographically, only fields are not sorted.

core/src/col_metadata.rs
```diff
@@ -124,60 +124,95 @@ impl SqlGetSchemaResponse {
...
-                Ok((datasource_name, obj_schema.clone()))
+                Ok::<std::vec::IntoIter<MongoColMetadata>, Error>(
+                    Self::schema_to_col_metadata(&datasource_schema, current_db, &datasource_name)?
+                        .into_iter(),
+                )
             })
-            .collect::<Result<Vec<(String, ObjectSchema)>>>()?;
+            // 3. flatten each Ok(inner_iterator) into the top iterator, will short circuit if an
+            //    Err is hit and just return the first Error.
+            .flatten_ok()
```
pmeredit, line 151: ^^ this was the best to figure out

nbagnard (commented):
Things look good!
Nice refactoring on col_metadata and also thanks for adding the different ascii flavor of sqlstate and sqlgetDiagRec :)
The only question I have is on a test where the error message seem to be expected to be empty and I find that odd, if you can simply show me what I miss, that'd be great. I expected the error message to be "[MongoDB][API] A schema pattern was specified, and the driver does not support schemas" since you are passing a schema and it is not supported.

integration_test/tests/query_tests.rs
```diff
@@ -101,4 +101,26 @@ mod integration {
...
+        conn_string.push_str("DATABASE=integration_test");
+
+        let env = Environment::new().unwrap();
+        let conn = env.connect_with_connection_string(&conn_string).unwrap();
+
+        let res = conn.columns("", "foo", "", "");
+        assert!(res.is_err());
+        unsafe {
+            assert_eq!(
+                "ODBC emitted an error calling 'SQLColumns':\nState: HYC00, Native error: 0, Message: ".to_string(),
```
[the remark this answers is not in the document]
pmeredit: I haven't been able to figure out yet if this is an us problem or and odbc_api problem. I'll look into it.

pmeredit:
evergreen merge

PR #55, opened 2022-12-16
SQL-985: Implement test file generator

mattChiaravalloti:
This PR adds a test file generator executable which runs the integration tests in `resources/integration_test/tests` and writes the results to yaml files. This is heavily influenced by the JDBC test generator (shout out to @bucaojit 💯 ).

Caveats:
- `serde_yaml` is criminally underfeatured and does not support any formatting options. I looked into a few other libraries for formatting but those required quite a bit of overhead to get going and I didn't find any that could natively integrate with serde_yaml. I'm thinking when people generate tests, they can use a command line utility or built-in IDE/text-editor utility to format the generated yaml as desired.
- The `SqlDateType` to `CDataType` mapping is imperfect and possibly incorrect in places, so please take a close look at that!
- I only tested this locally so far and always skipped actually running the integration tests (since I am running on a Mac). I am going to check on a windows host that this runs end-to-end as expected but anticipate only minor bug fixes (if any) to come from that, so this is ready for review 👍

### DIFF
--- a/integration_test/src/lib.rs
+++ b/integration_test/src/lib.rs
@@ -1,5 +1,4 @@
-#[cfg(test)]
-mod test_runner;
+pub mod test_runner;
 
 #[path = "../tests/common/mod.rs"]
 #[cfg(test)]
--- a/integration_test/src/lib.rs
+++ b/integration_test/src/lib.rs
@@ -1,5 +1,4 @@
 pub mod test_runner;
 
 #[path = "../tests/common/mod.rs"]
-#[cfg(test)]
 mod common;
--- a/integration_test/src/test_runner/mod.rs
+++ b/integration_test/src/test_runner/mod.rs
@@ -16,7 +16,7 @@ use crate::{
 };
 use thiserror::Error;
 
-const TEST_FILE_DIR: &str = "./resources/integration_test/tests";
+const TEST_FILE_DIR: &str = "../resources/integration_test/tests";
 const SQL_NULL_DATA: isize = -1;
 const BUFFER_LENGTH: usize = 200;
 
--- a/integration_test/src/test_runner/mod.rs
+++ b/integration_test/src/test_runner/mod.rs
@@ -16,7 +16,7 @@ use crate::{
 };
 use thiserror::Error;
 
-const TEST_FILE_DIR: &str = "../resources/integration_test/tests";
+const TEST_FILE_DIR: &str = "./resources/integration_test/tests";
 const SQL_NULL_DATA: isize = -1;
 const BUFFER_LENGTH: usize = 200;
 
--- a/integration_test/src/test_runner/mod.rs
+++ b/integration_test/src/test_runner/mod.rs
@@ -12,7 +12,7 @@ use std::{fmt, fs, path::PathBuf};
 
 use crate::{
     common::{get_sql_diagnostics, sql_return_to_string},
-    test_runner::test_generator_util::generate_baseline_test_files,
+    test_runner::test_generator_util::generate_baseline_test_file,
 };
 use thiserror::Error;
 
--- a/integration_test/src/test_runner/mod.rs
+++ b/integration_test/src/test_runner/mod.rs
@@ -127,6 +127,9 @@ pub fn integration_test() -> Result<()> {
     run_integration_tests(false)
 }
 
+/// Run an integration test. The generate argument indicates whether
+/// the test results should written to a file for baseline test file
+/// generation, or be asserted for correctness.
 pub fn run_integration_tests(generate: bool) -> Result<()> {
     let env = create_environment_v3().unwrap();
     let paths = load_file_paths(PathBuf::from(TEST_FILE_DIR)).unwrap();
@@ -157,8 +160,8 @@ pub fn run_integration_tests(generate: bool) -> Result<()> {
     Ok(())
 }
 
-/// load_file_paths reads the given directory and returns a list its file path
-/// names.
+/// load_file_paths reads the given directory and returns a list of its file
+/// path names.
 pub fn load_file_paths(dir: PathBuf) -> Result<Vec<String>> {
     let mut paths: Vec<String> = vec![];
     let entries = fs::read_dir(dir).map_err(|e| Error::InvalidDirectory(format!("{:?}", e)))?;
@@ -222,6 +225,9 @@ fn check_array_length(array: &Vec<Value>, length: usize) -> Result<()> {
     Ok(())
 }
 
+/// Run a query integration test. The generate argument indicates
+/// whether the test results should written to a file for baseline
+/// test file generation, or be asserted for correctness.
 fn run_query_test(
     query: &str,
     entry: &TestEntry,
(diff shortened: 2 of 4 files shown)


integration_test/src/test_runner/mod.rs
```diff
@@ -110,38 +117,14 @@ impl fmt::Display for TestDef {
...
-    let integration_test: IntegrationTest =
-        serde_yaml::from_reader(f).map_err(|e| Error::CannotDeserializeYaml(format!("{:?}", e)))?;
-    Ok(integration_test)
-}
-
 /// integration_test runs the query and function tests contained in the TEST_FILE_DIR directory
 #[test]
 #[ignore]
-pub fn integration_test() -> Result<(), Error> {
+pub fn integration_test() -> Result<()> {
```
mattChiaravalloti, line 123: Since this is arguably the "entry point" of this file, I moved it up to be the first function. That is why you see a few functions on the left-hand side of your screen removed and re-added below.

integration_test/src/test_runner/test_generator_util.rs
```diff
@@ -0,0 +1,168 @@
...
+        expected_precision: Some(expected_precision),
+        expected_scale: Some(expected_scale),
+        expected_nullability: Some(expected_nullability),
+    };
+
+    // 4. Write the TestEntry to a file
+    let now = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap()
+        .as_millis();
```
mattChiaravalloti, line 117: The JDBC driver's test generator uses a nice `MMddHHmmss` date format, but date formatting in rust is a bit of a headache and requires importing external crates (such as `chrono` which has security issues or `time` which is not intuitive) so I decided to just go with milliseconds instead. It is doubtful people will retain generated files over time so the timestamp portion of these file names is not very important anyway, but lmk if you have opinions on changing this.

integration_test/src/test_runner/test_generator_util.rs
```diff
@@ -0,0 +1,168 @@
...
+    serde_yaml::to_writer(writer, &test_entry).map_err(|err| Error::Yaml(err.to_string()))
+}
+
+// Get the expected CDataType for the provided sql_type.
+fn get_expected_data_type(sql_type: &Value) -> CDataType {
+    match sql_type {
+        Value::Number(n) => {
+            let sdt = SqlDataType(n.as_i64().unwrap() as i16);
+            match sdt {
+                SqlDataType::UNKNOWN_TYPE => CDataType::Default,
```
mattChiaravalloti, line 136: We don't support `CDataType::Default`, of course, but I wasn't sure what to put here. `Char`? Something else?

integration_test/src/test_runner/mod.rs
```diff
@@ -8,6 +10,7 @@ use serde_json::value::Value;
 use std::ptr::null_mut;
 use std::{fmt, fs, path::PathBuf};
 
+use crate::test_runner::test_generator_util::generate_baseline_test_file;
 use thiserror::Error;
 
 const TEST_FILE_DIR: &str = "../resources/integration_test/tests";
```
bucaojit, line 16: Perhaps because the test_generator is run as an executable vs test, but the directory that works for the test generator in my attempt to run it is `./resources/integration_test/tests`.  Where resources is in the same directory and not parent dir.

integration_test/src/test_runner/test_generator_util.rs
bucaojit: Are we able to convert all the UNKOWN_TYPEs to char?  This was something I was not sure about when writing the test runner.  `Char` would be fine, similar to how in the JDBC test harness we have the option of using getString().

integration_test/src/test_runner/test_generator_util.rs
```diff
@@ -0,0 +1,168 @@
+use crate::test_runner::{
+    fetch_row, get_column_attribute, get_column_count, get_data, Error, Result, TestEntry,
+};
+use lazy_static::lazy_static;
+use odbc::{safe::AutocommitOn, Allocated, NoResult, Statement};
+use odbc_sys::{CDataType, Desc, SqlDataType, USmallInt};
+use serde_json::{Number, Value};
+use std::string::ToString;
+
+const GENERATED_TEST_DIR: &str = "../resources/generated_test";
```
bucaojit, line 10:
```suggestion
const GENERATED_TEST_DIR: &str = "./resources/generated_test";
```
Same comment here about the `resources` directory location being in same directory as opposed to parent.

integration_test/src/test_runner/test_generator_util.rs
bucaojit: This is good.  
The important part is that it won't overwrite the already generated files and it sorts well to see the latest files which epoch time does.

integration_test/src/test_runner/mod.rs
```diff
@@ -654,7 +679,7 @@ fn get_column_count(stmt: &Statement<Allocated, NoResult, AutocommitOn>) -> Resu
     }
 }
 
-fn fetch_row(stmt: &Statement<Allocated, NoResult, AutocommitOn>) -> Result<bool, Error> {
+fn fetch_row(stmt: &Statement<Allocated, NoResult, AutocommitOn>) -> Result<bool> {
```
bucaojit, line 682: I'm curious, was there a change made that required the `Error` to be added to the result?

bucaojit (changes requested):
Nice work on the test generator! 

I had a few comments that you may have caught already with your testing.  I was able to run it up until it runs SQLGetData() on the results at which point it hits the errors that are fixed in PR 51 and PR 56.

integration_test/src/test_runner/test_generator_util.rs
mattChiaravalloti: I'm not sure, maybe @pmeredit can weigh in on this one

integration_test/src/lib.rs
```diff
@@ -1,5 +1,4 @@
-#[cfg(test)]
-mod test_runner;
+pub mod test_runner;
 
 #[path = "../tests/common/mod.rs"]
 #[cfg(test)]
```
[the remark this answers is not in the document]
mattChiaravalloti: oh, of course 😓 thanks for pointing that out

integration_test/src/test_runner/mod.rs
mattChiaravalloti: I made this refactor to be more consistent with (most of) the rest of this codebase and with the mongosql-rs codebase where we often define in each module an `module::Error` enum and a `module::Result<T>` type defined as `std::Result<T, module::Error>`. Here, I initially made a `test_generator::Error` enum and a `test_generator::Result<T>` but then realized I didn't need that and could just use the `test_runner::Error/Result` types. At that point, I decided to do the refactor!

integration_test/src/test_runner/test_generator_util.rs
pmeredit: I think `Char` is fine.

pmeredit (approved):
LGTM, just a few minor cosmetic changes.

bucaojit (approved):
LGTM!  Looks good with the updates.

mattChiaravalloti:
evergreen merge

mattChiaravalloti:
evergreen merge

PR #57, opened 2022-12-20
SQL-1148: Allow (Mongo) URI to be part of the ODBC URI

pmeredit:
The test I removed was actually an artifact of the old string interface. The test itself really should not have existed.

Note that "DATABASE" is handled differently from the `ClientOptions` so not change with respect to "DATABASE" was needed. It has to be this way because the current database can be changed at any time by the driver, and so must be part of the Dbc Handle.


core/src/odbc_uri.rs
```diff
@@ -148,36 +152,70 @@ impl<'a> ODBCUri<'a> {
...
+                Error::InvalidUriFormat(format!(
                     "One of {:?} is required for a valid Mongo ODBC Uri",
                     names
                 ))
             }
         })
     }
 
-    // remove_to_mongo_uri converts this ODBCUri to a mongo_uri String. It will
+    // try_into_client_options converts this ODBCUri to a mongo_uri String. It will
```
pmeredit, line 168: The previous name I used did not conform to standard rust conventions!

odbc/src/api/functions.rs
```diff
@@ -832,7 +831,7 @@ pub unsafe extern "C" fn SQLDisconnect(connection_handle: HDbc) -> SqlReturn {
 
 fn sql_driver_connect(conn: &Connection, odbc_uri_string: &str) -> Result<MongoConnection> {
     let mut odbc_uri = ODBCUri::new(odbc_uri_string)?;
-    let mongo_uri = odbc_uri.remove_to_mongo_uri()?;
+    let client_options = odbc_uri.try_into_client_options()?;
     let auth_src = odbc_uri.remove_or_else(|| "admin", &["auth_src"]);
```
nbagnard, line 835: To make sure that the KEYWORD list and the options to retrieve are the same, could we have the options/keywords specified as const str and then the regex for parsing the ODBC connection string would use them, and also the odbc_uri remove calls. This was we won't end up with discrepancies between the 2.

nbagnard (changes requested):
I like using ClientOptions much better than using a string for sure 👍 
The only corner case is the one around the auth_source. It is unfortunate that if there is no UID/PWD (or authMech) provided the authSource is discarded from the ClientOptions.
Nathan is also looking at working around this on the JDBC side for the same reason.

core/src/odbc_uri.rs
```diff
@@ -148,36 +152,70 @@ impl<'a> ODBCUri<'a> {
...
-        ))
+        let cred = Credential::builder()
+            .username(user.to_string())
+            .password(pwd.to_string())
+            .build();
+        let mut client_options = ClientOptions::builder()
+            .hosts(vec![ServerAddress::parse(server)?])
+            .credential(cred)
+            .build();
+        if ssl.is_some() && ssl.unwrap() != "0" && ssl.unwrap().to_lowercase() != "false" {
```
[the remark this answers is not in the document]
bucaojit: In the connector we won't be going down this path since it will be setting only the URI, driver, and credentials. 
Since we are allowing users to use the other connection string options if the URI isn't set, it does makes sense to include the ODBC connection settings that we know users will want to set: ssl and connectTimeoutMS.  For users who want more customizing we can say to use the URI option.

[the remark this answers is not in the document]
nbagnard: That works for me! I personally don't know which ones users will want except ssl and connectTimeoutMS so I am fine to start with this.
Thank you @bucaojit for confirming :)

core/src/odbc_uri.rs
```diff
@@ -148,36 +152,70 @@ impl<'a> ODBCUri<'a> {
...
+            if user.is_some() {
+                client_options.credential.as_mut().unwrap().username = user.map(String::from);
+            }
+            let pwd = self.remove(PWD);
+            if pwd.is_some() {
+                client_options.credential.as_mut().unwrap().password = pwd.map(String::from);
+            }
+            return Ok(client_options);
+        }
+        let user = self.remove(USER);
```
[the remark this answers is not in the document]
pmeredit: That's a fair point!

core/src/odbc_uri.rs
```diff
@@ -148,36 +152,70 @@ impl<'a> ODBCUri<'a> {
...
+    pub fn try_into_client_options(&mut self) -> Result<ClientOptions> {
+        let uri = self.remove(URI);
+        if let Some(uri) = uri {
+            return self.handle_uri(uri);
+        }
+        self.handle_no_uri()
+    }
+
+    fn handle_uri(&mut self, uri: &str) -> Result<ClientOptions> {
+        let mut client_options = ClientOptions::parse(uri)?;
```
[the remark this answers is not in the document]
pmeredit: The authSource actually won't be lost in this code. If there is a URI, the only thing it does now is _update_ the `ClientOptions` with credentials. None of the original parsed settings are lost.

I should add a test for that, however!.

core/src/odbc_uri.rs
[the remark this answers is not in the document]
pmeredit: The SSL one is a hold over from the BIC odbc driver. I am fine removing the option if we want :)

Should probably consider supporting all the options we did before?

core/src/odbc_uri.rs
```diff
@@ -438,98 +479,131 @@ mod unit {
...
-                "mongodb://foo:bar@127.0.0.1:27017?ssl=true".to_string(),
+                ClientOptions::parse("mongodb://foo:bar@127.0.0.1:27017/?ssl=true")
+                    .unwrap()
+                    .tls,
                 ODBCUri::new("USER=foo;PWD=bar;SERVER=127.0.0.1:27017;SSL=true")
                     .unwrap()
-                    .remove_to_mongo_uri()
+                    .try_into_client_options()
                     .unwrap()
+                    .tls
```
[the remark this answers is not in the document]
pmeredit: I meant to add more unit tests and totally forgot. Thanks!

odbc/src/api/functions.rs
pmeredit, line 835: I can probably move this auth_src part into the try_into_client_options, also.

core/src/odbc_uri.rs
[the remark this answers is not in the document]
nbagnard: I don't think so. Yes,please add a test, because I don't understand how it can work.
Here is the code from my own side test:
```
fn check_cred(client_options: &ClientOptions) {
    if client_options.credential.is_some() {
        let creds = (client_options.clone()).credential.unwrap();
        if creds.source.is_some() {
            println!("AuthSource = {}", creds.source.unwrap());
        } else {
            println!("No authSource");
        }
    } else {
        println!("No credential")
    }
}
#[test]
fn test_client_options() {
    let uris = vec![
        "mongodb://localhost/?authSource=authDB",
        "mongodb://localhost/",
        "mongodb://uid@pwdlocalhost/?authSource=authDB",
        "mongodb://uid@pwdlocalhost/",
    ];
    println!("---------------------------------");
    for uri in uris {
        println!("-- URI : {} --", &uri);
        let mut client_options = ClientOptions::parse(uri).unwrap();

        print!("Initial parse ==> ");
        check_cred(&client_options);

        if client_options.credential.is_none() {
            print!("Adding UID and PWD retrieved from somewhere else ==> ");
            // Add credential info
            client_options.credential = Some(
                Credential::builder()
                    .username("uid".to_string())
                    .password("pwd".to_string())
                    .build(),
            );

            check_cred(&client_options);
        }
        println!("---------------------------------");
    }
}
```

and here is the output :
```
---------------------------------
-- URI : mongodb://localhost/?authSource=authDB --
Initial parse ==> No credential
Adding UID and PWD retrieved from somewhere else ==> No authSource
---------------------------------
-- URI : mongodb://localhost/ --
Initial parse ==> No credential
Adding UID and PWD retrieved from somewhere else ==> No authSource
---------------------------------
-- URI : mongodb://uid@pwdlocalhost/?authSource=authDB --
Initial parse ==> AuthSource = authDB
---------------------------------
-- URI : mongodb://uid@pwdlocalhost/ --
Initial parse ==> AuthSource = admin
---------------------------------
```

Because the code replaces the credential sub-struct with the one create in the function which only assigns username and password, it looks like there will be no authsource set if the original URI does not contains a UID + PWD.

odbc/src/api/functions.rs
nbagnard: Yes, plus I don't think we'll need it in connect once we move from using a `ping` command to using a `Select 1` query to validate the connection.

core/src/odbc_uri.rs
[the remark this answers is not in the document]
pmeredit: It doesn't replace it, it updates it. The problem is the URI parser doesn't create a `Credential` when there is an `authSrc` but no `UID` or `PWD`

core/src/odbc_uri.rs
```diff
@@ -37,6 +37,7 @@ lazy_static! {
     .case_insensitive(true)
     .build()
     .unwrap();
+    static ref AUTH_SRC_REGEX: Regex = Regex::new(r"authSource=([a-zA-Z_0-9]*)").unwrap();
```
nbagnard, line 40: You want AuthSource to be case insensitive. Maybe use the RegexBuilder because Regex::new seems to create a case-sensitive regex., or create a group for `authSrc` and set the case-sensitivity there.
And for the value, I don't think that you can limit the authSource to alphanumeric characters. One of the reason is that in the description of the authSource option it says `The PLAIN (LDAP), GSSAPI (Kerberos), and MONGODB-AWS (IAM) authentication mechanisms require that [authSource](https://www.mongodb.com/docs/manual/reference/connection-string/#mongodb-urioption-urioption.authSource) be set to $external`.
 I also did not see anywhere that it had to be restricted to alphanumeric characters.  It does appear so from [this page](https://www.mongodb.com/docs/manual/reference/limits/).

[nit] Another thing that I can suggest is to name your capture group. This way, no need to worry about the capture index, does it start at 0 or 1, etc...

In the end, I would use this:
```
lazy_static! {
    static ref AUTH_SRC_REGEX: Regex = RegexBuilder::new(r#"[&?]authSource=(?P<authSrc>[^&]*)"#)
        .case_insensitive(true)
        .build()
        .unwrap();
}
and grab the value with
```
let auth_src_val = AUTH_SRC_REGEX
            .captures(&uri)
            .and_then(|cap| cap.name("authSrc").map(|authSrc| authSrc.as_str()));
```
It support all characters except '&' because this one marks the next option (if the authSource name contains a '&' it will have to be URL encoded). I think that it is OK to let the Rust driver or MDB server handle the validity of the value.
This way, it purposefully allows '$' and will support `$external` (and exotic source names too)
```
My unit tests seem Ok.
Here is the list I tested out:
```
       "mongodb://localhost/?authSource=authDB",
        "mongodb://localhost/",
        "mongodb://localhost/?auTHSource=aut:hD@B",
        "mongodb://localhost/?auTHSource=aut:hD@B&anotherOpt=tgg#fed",
        "mongodb://uid:pwd@localhost/?authSource=$external&anotherOpt=tgg#fed",
        "mongodb://localhost/?something=test&auTHSource=aut:hD@B",
        "mongodb://localhost/?something=test&tlsCertificateKeyFilePassword='myauthSource=aut:hD@B'&authSource=jfhbgvhj"
```

core/src/odbc_uri.rs
```diff
@@ -729,5 +736,26 @@ mod unit {
                 )
             );
         }
+
+        #[test]
+        fn auth_source_maintained_when_user_and_password_not_embedded() {
```
nbagnard, line 741: It's good to have one test, but I think the coverage is really minimal here.
Can you expend this to more combinations? 
The one I can think of are :authSource is the first, the last, the only options, authSource is not present, authSrc= is part of anothter option value (more or less the examples from my test cases described above 😉)

core/src/odbc_uri.rs
nbagnard, line 40: [super nit/question] Why not using lazy_static? If a user never pass the URI we don't need to generate the regex, isn't it?

nbagnard (changes requested):
Yep, I think we can go this route and extract the authSource option.
I spent some time trying to make the regex as bulletproof as I could. The only catch is that `&` **have to** be URL encoded but it is not specified anywhere in the doc. Since the Rust driver seem to also believe it should be, I think we are safe (for now 😉).
I have only a couple of small requests left to make sure we handle it fully, when it is present either in the URI or the ODBC settings, or both.

odbc/src/api/functions.rs
```diff
@@ -850,8 +848,7 @@ fn sql_driver_connect(conn: &Connection, odbc_uri_string: &str) -> Result<MongoC
     // create an impl From Result<T, mongo_odbc_core::Error> to Result<T, ODBCError>
     // hence this bizarre Ok(func?) pattern.
     Ok(mongo_odbc_core::MongoConnection::connect(
-        &mongo_uri,
-        auth_src,
+        client_options,
```
nbagnard, line 851: Can you also remove this line above since app_name is handled in the URI:
`let application_name = odbc_uri.remove(&["app_name", "application_name"]);`

core/src/odbc_uri.rs
pmeredit: If it's not in a `lazy_static` it's actually built at compile time, which is nice if we can do that. I thought this was in a lazy static, but I reverted all the authSource code, so I'm not sure.

odbc/src/api/functions.rs
pmeredit: Do we want to remove the app_name option then?

pmeredit: Ahh never mind.

core/src/odbc_uri.rs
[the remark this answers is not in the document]
pmeredit: Ok, removing these now!

core/src/odbc_uri.rs
pmeredit: Yes, this is correct. & should be percent encoded if used!

pmeredit:
@nbagnard should be good now!

nbagnard (commented):
LGTM!

nbagnard (approved):

bucaojit (approved):
LGTM 💯

pmeredit:
evergreen merge

pmeredit:
evergreen merge

PR #58, opened 2022-12-21
SQL-1147: Support app_name/application_name options in connection string

bucaojit:
Small fix to add app_name/application_name to the list of valid keywords in the URI.

Confirmed that the application_name value set in the PowerBI connector gets set in the ADF logs:
`"commandName": "aggregate", "appName": "MongoDB Atlas SQL - PowerBI", "username": "mhuser" ...`

@pmeredit I can hold off on this change if you are making changes to the KEYWORDS.

### DIFF
--- a/odbc/src/api/odbc_uri.rs
+++ b/odbc/src/api/odbc_uri.rs
@@ -14,6 +14,8 @@ const SSL: &[&str] = &["ssl", "tls"];
 
 lazy_static! {
     static ref KEYWORDS: RegexSet = RegexSetBuilder::new([
+        "^APP_NAME$",
+        "^APPLICATION_NAME$",
         "^AUTH_SRC$",
         "^DATABASE$",
         "^DRIVER$",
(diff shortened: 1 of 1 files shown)


odbc/src/api/odbc_uri.rs
```diff
@@ -14,6 +14,8 @@ const SSL: &[&str] = &["ssl", "tls"];
 
 lazy_static! {
     static ref KEYWORDS: RegexSet = RegexSetBuilder::new([
+        "^APP_NAME$",
```
nbagnard, line 17: Are these 2 different options or just 2 synonyms?
If they are synonyms then which one takes precedence? 
Also, I suppose that this PR is not totally ready/done because then the `appName` needs to be propagated to the MDB client, isn't it?

bucaojit: They are synonyms.  Since it's going through the `remove` function, the first value that is in the connection string is used.
The implementation to add it to the MDB client was already done [here](https://github.com/mongodb/mongo-odbc-driver/blob/be876b0cb7317670b6616eeb750addbc61a9b10b/odbc/src/api/functions.rs#L848), the missing part was adding it to the keyword list.

nbagnard: Perfect!

nbagnard (approved):
LGTM!

pmeredit (approved):
LGTM, don't worry about any merge issues on my end.

bucaojit:
The PowerBI connector has changed to use the URI option instead of the separate ODBC options (APP_NAME, SSL, SERVER, etc).  Since the appName value can be set in the URI ([link](https://www.mongodb.com/docs/manual/reference/connection-string/#mongodb-urioption-urioption.appName)), we will no longer need this change in the ODBC driver to support `app_name/application_name` in the connection string.  
I will go ahead and close this PR without merging, additional changes will be done to remove the existing processing of the option in the connection string.

PR #59, opened 2022-12-21
SQL-1018: Update connection validation method

nbagnard:
Update the connection validation using a `SELECT 1` query instead of a `ping` command to validate the connection.
This allow to validate that the user has indeed access to the database.


nbagnard:
@pmeredit I added you as a reviewer but it is more as an FYI because it is linked to the change you are making for the URI connection parameter

core/src/conn.rs
```diff
@@ -53,16 +52,15 @@ impl MongoConnection {
...
-        Ok(MongoConnection {
+
+        let connection = MongoConnection {
             client,
             current_db: current_db.map(String::from),
             operation_timeout: operation_timeout.map(|to| Duration::new(to as u64, 0)),
-        })
+        };
+
+        MongoQuery::execute(&connection, login_timeout, "select 1")?;
```
bucaojit, line 62: I'm seeing errors in some of our integration tests related to not having the `database` value being set by this point.  We can add a check for the database value prior to here or use the auth_src database again like the ping command.
[Test run ](https://evergreen.mongodb.com/task/mongosql_odbc_driver_ubuntu2004_asan_integration_test_patch_be876b0cb7317670b6616eeb750addbc61a9b10b_63a35ba0e3c3311d0ea695e4_22_12_21_19_17_08)
Error: `[MongoDB][Core] No database provided for query'`

bucaojit (changes requested):
The change to use select 1 look good.  There is more work to do it seems to get the query to work.

core/src/conn.rs
nbagnard: @bucaojit 
I thought that with Patrick changes at this point we always would have a DB. But you are right, since it is an option, it is not true.
I am wondering if there is an issue with our design. If no DB is assigned to the MongoConnection, then the first query will always fail and since we are currently support setting the current catalog via SQLSetConnAttr, we don't have a good way to set one after the connection is created.
So, I see 3 different directions we can go:
- SQLDriverConnect fails because no DB has been provided and it makes no sense to keep going since the connection will not be usable for querying
- As you suggest, I can run the query on `admin` if no DB are provided. The user will not be able to execute any query but will be able to use the connection for catalog information retrieval if he wants to.
- We always set the current DB to admin by default if none is provided. This way, both the connection and queries will work. I am not sure if every user has access to admin, but since this is what we did for ping, it is probably OK. I am not sure I see much value in doing that though, because there is probably not a huge amount of data to be queried on `admin`, but it can let someone discover which DBs are available to him (I would hope they do by the time they connect, but well... you never know).

I'd rather fail at connection time, until we support switching catalog via SQLSetConnAttr because it might otherwise lead to user frustration if they have to redo the whole connection cycle when they missed specifying a DB in the connection setting.

bucaojit: I appreciate the detailed response of the different options.  
Failing at connection time sounds good when the DB is missing.  That would help point users to the problem being connection related especially when the displayed error message isn't clear ie `The driver returned invalid (or failed to return) SQL_DRIVER_ODBC_VER: 03.08`.  And having the connection string work without the DB option for catalog retrieval but not for queries would add to the confusion than help.

nbagnard: I added the ADF_LOCAL_TEST_DB variable so that tests using the function generating the default connection string will always have a DB.
However, something else is up. It seems unrelated but I did not have time to investigate : Tests fail on Windows (SQL_ERROR on connection) but are successful on Linux.
See this [run](https://spruce.mongodb.com/version/63a4d8bf3e8e867f1c698464/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

pmeredit: We know the DM on windows and unixodbc do not behave the same! :(

odbc/src/api/functions.rs
```diff
@@ -833,7 +833,7 @@ pub unsafe extern "C" fn SQLDisconnect(connection_handle: HDbc) -> SqlReturn {
 fn sql_driver_connect(conn: &Connection, odbc_uri_string: &str) -> Result<MongoConnection> {
     let mut odbc_uri = ODBCUri::new(odbc_uri_string)?;
     let mongo_uri = odbc_uri.remove_to_mongo_uri()?;
-    let auth_src = odbc_uri.remove_or_else(|| "admin", &["auth_src"]);
```
pmeredit, line 836: If we are removing the auth_src argument, we should delete this line.

pmeredit (changes requested):

odbc/src/api/functions.rs
nbagnard: That's right

core/src/query.rs
```diff
@@ -51,13 +51,20 @@ impl MongoQuery {
...
-        let cursor = db.aggregate(pipeline, options)?;
+        let cursor: Cursor<Document> = match query_timeout {
+            Some(i) => {
+                if i > 0 {
+                    let opt = AggregateOptions::builder()
+                        .max_time(Duration::from_millis(i as u64))
+                        .build();
+                    db.aggregate(pipeline, opt)?
+                } else {
+                    // If the query timeout is 0, it means "no timeout"
```
pmeredit, line 62: Nice

pmeredit (approved):
LGTM!

bucaojit (approved):
LGTM!

nbagnard:
evergreen merge

PR #60, opened 2022-12-22
SQL-1045: Implement SQLGetTypeInfo

jchemburkar:
Implementation of the SQLGetTypeInfo function. This function generates a result set and stores it in the handle. To do so, [type_info.rs](https://github.com/jchemburkar/mongo-odbc-driver/blob/193b660cde21c46fb91efde684a4fab30d50cd86/core/src/type_info.rs) was created. Similar to [databases.rs](https://github.com/jchemburkar/mongo-odbc-driver/blob/193b660cde21c46fb91efde684a4fab30d50cd86/core/src/databases.rs) and [fields.rs](https://github.com/jchemburkar/mongo-odbc-driver/blob/193b660cde21c46fb91efde684a4fab30d50cd86/core/src/fields.rs), this implements `MongoStatement` in order to create the result set.

I did my best when it came to [mapping metadata](https://github.com/jchemburkar/mongo-odbc-driver/blob/193b660cde21c46fb91efde684a4fab30d50cd86/core/src/type_info.rs#L233), however would call that out as somewhere that could benefit from extra scrutiny. Many of the mappings are not particularly neat between bson data types and fixed tabular column definitions.


odbc/src/api/get_type_info_tests.rs
```diff
@@ -0,0 +1,182 @@
+use crate::{handles::definitions::*, SQLGetTypeInfo};
+use bson::Bson;
+use odbc_sys::{SqlDataType, SqlReturn};
+
+mod unit {
+    use super::*;
+    use mongo_odbc_core::Error;
+
+    fn validate_result_set(data_type: SqlDataType, expectations: Vec<&str>) {
```
mattChiaravalloti, line 9: You don't need to change this but just for your own understanding of our typical approach to unit testing (and possibly a common Rust test paradigm): you _could_ refactor this helper function into a macro that produces all of the test functions for you! _Again, in this case, you don't actually need to do this change since I think ultimately the result would be the same number of lines of code **and** this is a very readable helper function and the call sites are very readable._ The way you _could_ make this change would be having a macro that generates the `#[test] fn test_name() { ... }` code for you, where the `...` is effectively wholesale replaced with the code you have in this helper function's body. The parameters of the macro would be test_name, data_type, and expectations. See [here](https://github.com/mongodb/mongo-odbc-driver/blob/36e6a5ed3ea0a83579c652d8a46afd44712a668c/odbc/src/api/connect_attr_tests.rs#L15) or [here](https://github.com/10gen/mongosql-rs/blob/927ec68cb9b48978e9ed806635c8c851a7dcb0cf/mongosql/src/translator/test.rs#L1) for examples of this in both our rust codebases.

I think I actually prefer this written this way in this case since it is only used a few times and the tests are all already well name and organized 👍 . Just sharing so you have the macro-model of unit testing in your tool belt for the future.

core/src/type_info.rs
```diff
@@ -0,0 +1,304 @@
...
+    &BsonTypeInfo::MAXKEY,              // SqlDataType(0)
+    &BsonTypeInfo::MINKEY,              // SqlDataType(0)
+    &BsonTypeInfo::NULL,                // SqlDataType(0)
+    &BsonTypeInfo::OBJECT,              // SqlDataType(0)
+    &BsonTypeInfo::OBJECTID,            // SqlDataType(0)
+    &BsonTypeInfo::SYMBOL,              //SqlDataType(0)
+    &BsonTypeInfo::TIMESTAMP,           // SqlDataType(0)
+    &BsonTypeInfo::UNDEFINED,           // SqlDataType(0)
+    &BsonTypeInfo::INT,                 // SqlDataType(4)
+    &BsonTypeInfo::BOOL,                // SqlDataType(7)
```
mattChiaravalloti, line 37:
```suggestion
    &BsonTypeInfo::BOOL,                // SqlDataType(-7)
```

core/src/type_info.rs
```diff
@@ -0,0 +1,304 @@
...
+                    3 => match type_info.precision {
+                        Some(precision) => Bson::Int32(precision as i32),
+                        None => Bson::Null,
+                    },
+                    4 | 5 => match **type_info {
+                        BsonTypeInfo::DATE | BsonTypeInfo::STRING => Bson::String("'".to_string()),
+                        _ => Bson::Null,
+                    },
+                    6 => Bson::Null,
+                    7 => Bson::Boolean(**type_info == BsonTypeInfo::OBJECTID),
```
mattChiaravalloti, line 247: Why is this the returned value?

core/src/type_info.rs
```diff
@@ -0,0 +1,304 @@
...
+                    8 => match **type_info {
+                        BsonTypeInfo::STRING => Bson::Int32(1),
+                        _ => Bson::Null,
+                    },
+                    9 => {
+                        match type_info.searchable {
+                            // SQL_SEARCHABLE and SQL_PRED_NONE, respectively
+                            true => Bson::Int32(3),
+                            false => Bson::Int32(0),
```
mattChiaravalloti, line 256:
```suggestion
                        match type_info.searchable {
                            // SQL_SEARCHABLE and SQL_PRED_NONE, respectively
                            true => Bson::Int32(something::SQL_SEARCHABLE),
                            false => Bson::Int32(something::SQL_PRED_NONE),
```
Can we possibly import those values from somewhere so they are clearly named within the code itself? The comment is probably good enough, but named variables would be great.

core/src/type_info.rs
```diff
@@ -0,0 +1,304 @@
...
+        match DATA_TYPES.get((self.current_type_index - 1) as usize) {
+            Some(type_info) => {
+                Ok(Some(match col_index {
+                    1 | 13 => Bson::String(type_info.type_name.to_string()),
+                    2 | 16 => Bson::Int32(type_info.sql_type.0 as i32),
+                    3 => match type_info.precision {
+                        Some(precision) => Bson::Int32(precision as i32),
+                        None => Bson::Null,
+                    },
+                    4 | 5 => match **type_info {
```
mattChiaravalloti, line 242: For all of these that match `**type_info` itself (`4`, `5`, `8`, `10`, `12`, `17`, and `18`), I wonder if we should make these optional fields on the `BsonTypeInfo` struct so that we could follow the same pattern as the others! `match type_info.<relevant_field> { Some(val) => val | None => Null }`. What do you think?

mattChiaravalloti (changes requested):
This is in really good shape 🏋️ I mostly left fyi-style notes, and really only a few suggestions and questions.

odbc/src/api/get_type_info_tests.rs
```diff
@@ -0,0 +1,182 @@
...
+            "symbol",
+            "timestamp",
+            "undefined",
+            "int",
+            "bool",
+            "double",
+            "string",
+            "date",
+        ];
+        validate_result_set(SqlDataType(0), expectations);
```
nbagnard, line 76: You'll need to use Smallint values directly instead after changing the signature for SQLGetTypeInfoW.

odbc/src/api/functions.rs
```diff
@@ -2595,7 +2595,16 @@ pub unsafe extern "C" fn SQLGetStmtAttrW(
 ///
 #[no_mangle]
 pub unsafe extern "C" fn SQLGetTypeInfo(handle: HStmt, _data_type: SqlDataType) -> SqlReturn {
```
nbagnard, line 2597: Can you please change data_type from SqlDataType to Smallint to match the ODBC function signature and avoid problems if someone provide a dataType value which does not match the list of type in SqlDataType. This is related to [SQL-1151](https://jira.mongodb.org/browse/SQL-1151).
The list is not exhaustive and when the initial input does not match, it will end up with EXT_GUID, which does not match any of our types and therefore the type list returned will be empty, but this is just a by-product of luck.
Just a note that this might affect more code than just this because BsonTypeInfo is using a SqlDataType for sql_type and probably will need to be updated to a Smallint.
I think I would add the SQL types in our definition.rs file and make it an enum with ```#[repr(i16)]
#[derive(Debug, PartialEq, Eq, Clone, Copy, FromPrimitive)]``` so that we can apply a similar approach for converting smallint to SqlType (just a suggestion for the name because it has to not clash with SqlDataType) and can still leverage Enum features.

nbagnard, line 2597: You can remove the `_` from _data_type now that it is used ;)

odbc/src/api/get_type_info_tests.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    #[test]
+    fn test_specific_type_one_response() {
+        let expectations = vec!["int"];
+        validate_result_set(SqlDataType(4), expectations);
+    }
+
+    #[test]
+    fn test_invalid_type() {
```
nbagnard, line 86: Are you intending to test an invalid type (aka not a SQL type) or a SQL type which has no match?
I'd like to have both for each. This way no guessing and we know we are covered :)

odbc/src/api/get_type_info_tests.rs
```diff
@@ -0,0 +1,182 @@
...
+                .unwrap()
+                .as_ref()
+                .unwrap()
+                .get_value(1);
+            assert!(value.is_err());
+        }
+    }
+
+    #[test]
+    fn test_get_all_values() {
```
nbagnard, line 116: I think that I would rename it to `test_get_all_type_info_row_values` just to disambiguate what "values" means here.

core/src/type_info.rs
```diff
@@ -0,0 +1,304 @@
...
+                        BsonTypeInfo::DATE | BsonTypeInfo::STRING => Bson::String("'".to_string()),
```
nbagnard, line 243: This looks correct and inline with the Syntax and Semantics doc, but SQLColAttributes reports and empty string.
Can you add literal prefix and suffix to BsoTypeInfo and make the required changes to have the correct info reported throughout?

nbagnard (changes requested):
This looks good 👍 
I have a few request though around refactoring because I noticed that SqlColAttribute is very similar, but things are handled differently and might lead to differences, which is not great.
Also, since you are working on it right now, I'd be happy if you could make the change so that we don't have to when taking care for [SQL-1151](https://jira.mongodb.org/browse/SQL-1151).

odbc/src/api/functions.rs
```diff
@@ -2595,7 +2595,16 @@ pub unsafe extern "C" fn SQLGetStmtAttrW(
...
-    unimpl!(handle);
+    panic_safe_exec!(
+        || {
+            let mongo_handle = MongoHandleRef::from(handle);
+            let stmt = must_be_valid!((*mongo_handle).as_statement());
+            let types_info = MongoTypesInfo::new(_data_type);
```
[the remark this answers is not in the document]
jchemburkar: That makes sense! thanks for the context

core/src/type_info.rs
jchemburkar: good catch, ty!

core/src/type_info.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+}
+
+impl MongoStatement for MongoTypesInfo {
+    // iterate through the list, searching for the next "valid" data type.
+    // a type is valid if its sql type matches the desired sql type, or if we are getting all types.
+    fn next(&mut self, _: Option<&MongoConnection>) -> Result<bool> {
+        loop {
+            self.current_type_index += 1;
+            if self.current_type_index > (DATA_TYPES.len() as isize)
```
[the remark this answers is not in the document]
jchemburkar: Ah this can be removed. I initially had `current_type_index` as an isize, initialized to -1, so that it could be used to directly index into `DATA_TYPES`, however turning it into a usize made the code more readable. This was a relic of that, that I can remove!

core/src/type_info.rs
jchemburkar: I think this is a good idea, can move these to `definitions.rs`

odbc/src/api/get_type_info_tests.rs
jchemburkar: Thanks for laying this all out, this is very helpful. I had used a macro for testing on one of my very first tickets, but was just following an existing pattern. This really helps clarify!

core/src/type_info.rs
```diff
@@ -0,0 +1,304 @@
...
+        // 11 -> FIXED_PREC_SCALE
+        // 12 -> AUTO_UNIQUE_VALUE
+        // 13 -> LOCAL_TYPE_NAME
+        // 14 -> MINIMUM_SCALE
+        // 15 -> MAXIMUM_SCALE
+        // 16 -> SQL_DATA_TYPE
+        // 17 -> SQL_DATETIME_SUB
+        // 18 -> NUM_PREC_RADIX
+        // 19 -> INTERVAL_PRECISION
+        match DATA_TYPES.get((self.current_type_index - 1) as usize) {
```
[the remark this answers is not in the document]
jchemburkar: Ok, I think this makes sense to me too. I will give this a shot

core/src/type_info.rs
jchemburkar: I think this makes a lot of sense, especially given Natacha's feedback as well. Will look into this 😁

core/src/type_info.rs
jchemburkar: Is this re the logic or the value? Logic would be that object id (I think) should be non nullable in all cases. Value I used a boolean but should probably actually make it an integer (because it is possible the value could be in [0, 1, 2]).

odbc/src/api/get_type_info_tests.rs
jchemburkar: That is a good point -- this was meant to be one with no match (95 -> `TIMESTAMP_WITH_TIMEZONE`). I can add one with an int that does not correspond to any Sql type too! I think that is especially helpful given the above change to `Smallint` as well

odbc/src/api/functions.rs
jchemburkar: I can make this change, yes. Unless I'm missing something, I think we can remove this function from that ticket once this is done?

jchemburkar: In implementing, I wonder if a simpler solution could make sense. Specifically, we could compare the `data_type` value to each of our data type's `sql_type.0` (example below). Thus, if we receive a type not in SqlDataType, it will not match with any of our data types, and an empty list of info will be returned. This also limits the scope of changes needed. wdyt?

https://github.com/jchemburkar/mongo-odbc-driver/compare/SQL-1045...jchemburkar:mongo-odbc-driver:SQL-1045-Remove-Enums?expand=1

core/src/type_info.rs
```diff
@@ -0,0 +1,304 @@
...
+                    6 => Bson::Null,
+                    7 => Bson::Boolean(**type_info == BsonTypeInfo::OBJECTID),
+                    8 => match **type_info {
+                        BsonTypeInfo::STRING => Bson::Int32(1),
+                        _ => Bson::Null,
+                    },
+                    9 => {
+                        match type_info.searchable {
+                            // SQL_SEARCHABLE and SQL_PRED_NONE, respectively
+                            true => Bson::Int32(3),
```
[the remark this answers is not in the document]
jchemburkar: thanks for pointing this out -- I just gave these sections a read for familiarity, and agree with what you are saying here

odbc/src/api/functions.rs
jchemburkar: I think this _should_ work if we add the enum and try to match to it in the `SQLGetTypeInfo` function to validate that we received a valid sql type. We can then pass through the smallint to any logic needed by `type_info.rs` (as seen above). My only remaining question would be what SQL types we want in the SqlType enum. Would it just be the 8 or SQL types that correspond to Bson types? that would be my interpretation of the [spec](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/sql-type-identifiers?view=sql-server-ver16) which seems pretty lenient, but wanted to see if you agree.

core/src/col_metadata.rs
```diff
@@ -44,47 +47,59 @@ pub struct MongoColMetadata {
...
-            is_unsigned: false,
+            non_concise_type: bson_type_info.non_concise_type,
+            sql_code: bson_type_info.sql_code,
+            is_unsigned: bson_type_info.is_unsigned.unwrap_or(true),
             is_updatable: false,
         }
     }
+
+    pub fn new(
+        _current_db: &str,
```
mattChiaravalloti, line 88:
```suggestion
        current_db: &str,
```
another place where it looks like we're using an argument so we can remove the `_`

core/src/type_info.rs
```diff
@@ -0,0 +1,306 @@
...
+//     &BsonTypeInfo::OBJECT,              // SqlDataType(0)
+//     &BsonTypeInfo::OBJECTID,            // SqlDataType(0)
+//     &BsonTypeInfo::SYMBOL,              //SqlDataType(0)
+//     &BsonTypeInfo::TIMESTAMP,           // SqlDataType(0)
+//     &BsonTypeInfo::UNDEFINED,           // SqlDataType(0)
+//     &BsonTypeInfo::INT,                 // SqlDataType(4)
+//     &BsonTypeInfo::DOUBLE,              // SqlDataType(8)
+//     &BsonTypeInfo::STRING,              // SqlDataType(12)
+//     &BsonTypeInfo::DATE,                // SqlDataType(93)
+// ];
```
mattChiaravalloti, line 33:
```suggestion
```
old dead code?

mattChiaravalloti (changes requested):
A couple more small changes

core/src/type_info.rs
mattChiaravalloti: This is about the value. I still don't understand why we are specifically singling out ObejctId as "not nullable". In BSON aren't all types except Null/Missing/Undefined "not nullable" since "Null" is actually a type?

core/src/type_info.rs
jchemburkar: yes -- thought I had gotten all of it, and should have double checked before the re-request. apologies for the mess!

core/src/col_metadata.rs
jchemburkar: 👍  can do

nbagnard (changes requested):
The refactoring make me feel better about the consistency of the info we return. Thank you :)
Pretty smart solution for handling INVALID_SQL_TYPE too.
I have a couple more comments and requests. Nothing major though ;)

core/src/bson_type_info.rs
```diff
@@ -1,214 +1,426 @@
...
+        is_auto_unique_value: None,
+        is_unsigned: None,
+        num_prec_radix: None,
     };
     pub const NULL: BsonTypeInfo = BsonTypeInfo {
         type_name: "null",
         sql_type: SqlDataType::UNKNOWN_TYPE,
-        searchable: true,
+        non_concise_type: SqlDataType::UNKNOWN_TYPE,
+        searchable: SQL_PRED_BASIC,
```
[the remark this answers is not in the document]
jchemburkar: this is a good call out, and I believe nulls _cannot_ be compared, if I am understanding [these spec tests](https://github.com/10gen/mongosql-rs/blob/0cc4cc824a1eeab5cf4fc568bb5fa26a075e2dbc/tests/spec_tests/query_tests/operators_comparison.yml#L958) correctly, this may be better as `SQL_PRED_NONE`

core/src/bson_type_info.rs
```diff
@@ -1,214 +1,426 @@
...
+        is_auto_unique_value: None,
+        is_unsigned: None,
+        num_prec_radix: None,
     };
     pub const BSON: BsonTypeInfo = BsonTypeInfo {
         type_name: "bson",
         sql_type: SqlDataType::UNKNOWN_TYPE,
-        searchable: true,
+        non_concise_type: SqlDataType::UNKNOWN_TYPE,
+        searchable: SQL_PRED_BASIC,
```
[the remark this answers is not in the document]
jchemburkar: On closer inspection I am actually a bit unsure _what_ this `bson` type is, unless it is just a default option if no type matches? looking both at the syntax+semantics [types section](https://docs.google.com/document/d/14DnyMSzVNRkoyib8yM62MPJIrkrG1pE0KjRYOWOUuDc/edit#) and the related [bson spec](https://bsonspec.org/spec.html), both line up (in the same order) as the list of datatypes in this file, until we reach `bson` (which does not appear in either specs).

odbc/src/api/get_type_info_tests.rs
jchemburkar: 👍 yes, I think I did so in most places and must have just missed a couple!

core/src/bson_type_info.rs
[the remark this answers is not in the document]
mattChiaravalloti: You are understanding those tests correctly. In general, when either operand of a comparison operator is NULL or MISSING, the result is NULL.

core/src/bson_type_info.rs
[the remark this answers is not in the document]
mattChiaravalloti: Your intuition on this is correct -- the type here that we just call "bson" is not _actually_ a BSON type. We use it instead to represent the "any" type or any mutli-element "anyOf" schema. See [here](https://github.com/mongodb/mongo-odbc-driver/blob/c1b68f647d2364f0a19adf39821ae2bedd283027/core/src/json_schema.rs#L246).

core/src/type_info.rs
mattChiaravalloti: @jchemburkar bumping this one!

core/src/type_info.rs
```diff
@@ -0,0 +1,282 @@
...
+                14 | 15 => match type_info.scale {
+                    Some(scale) => Bson::Int32(scale as i32),
+                    None => Bson::Null,
+                },
+                17 => match type_info.sql_code {
+                    Some(subcode) => Bson::Int32(subcode),
+                    _ => Bson::Null,
+                },
+                18 => match type_info.num_prec_radix {
+                    Some(precision) => Bson::Int32(precision as i32),
```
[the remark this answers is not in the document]
jchemburkar: Thanks for pointing this out -- I believe I added it in `bson_type_info` already and failed to transfer that here. Will update

core/src/type_info.rs
jchemburkar: @mattChiaravalloti thanks for the bump :) Your point makes sense to me, thanks for the clarification. My mental model is stemming from this function being `intended for use in Data Definition Language (DDL) statements`, and relatedly, thinking of these as being the types of a `Column`  rather than of a specific value. Thus, following the logic of [MongoColMetadata nullability](https://github.com/mongodb/mongo-odbc-driver/blob/c1b68f647d2364f0a19adf39821ae2bedd283027/core/src/col_metadata.rs#L220), it would seem any of these types _could_ be nullable if they are not required, other than ObjectId which will populate a value regardless (?). 

This was the logic that led me to this, but definitely happy to change if there are issues or I'm misunderstanding anything!

jchemburkar: Per standup discussion, we will mark all types as nullable

core/src/bson_type_info.rs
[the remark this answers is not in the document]
jchemburkar: ^ thanks for checking my intuition on this, that makes sense to me. So then by same token, I think the [spec tests](https://github.com/10gen/mongosql-rs/blob/0cc4cc824a1eeab5cf4fc568bb5fa26a075e2dbc/tests/spec_tests/query_tests/operators_comparison.yml#L306) imply that in some cases, `anyOf` (and thus, this bson type) could be used in a where clause --> so we can keep `SQL_PRED_NONE`. Does that make sense?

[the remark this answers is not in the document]
mattChiaravalloti: Hmm, I don't see any `WHERE` clauses in the linked spec tests. But in general, it is possible to have `anyOf` types used in `WHERE` clauses because the only constraint on `WHERE` clauses is that the expression there be of type BOOL or NULL (or MISSING). Any number of expressions/expression compositions can result in that type, such as `a = b` where `a` and `b` have `AnyOf` schemas that are comparable.

> anyOf (and thus, this bson type) could be used in a where clause --> so we can keep SQL_PRED_NONE. Does that make sense?

Part of me thinks if it _can_ be used in a where clause, doesn't that mean it should be one of `SQL_PRED_CHAR`, `SQL_PRED_BASIC`, or `SQL_SEARCHABLE`? (Based on [this documentation](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgettypeinfo-function?view=sql-server-ver16) for `SEARCHABLE`). However, I think I agree with you that this should be `SQL_PRED_NONE` because we don't make any guarantees about what this "bson" type _really_ is and there are some types that are truly not comparable in any circumstance, so the safest thing we can say here is `SQL_PRED_NONE`.

tldr: using `SQL_PRED_NONE` for `BsonTypeInfo::BSON`'s `searchable` field sounds good to me 👍

nbagnard (commented):
Let's talk about your latest changes tomorrow. I think I missed something the other day when we discussed about differentiating invalid and not supported types and moving SqlDataType in definition: SqlDataType is not an enum in ODBC sys. It's a struct and it has a bunch of const one for each SQL type. So, the copy in this case does not help or change anything for you, correct? An enum with FromPrimitive would make your life easier

mattChiaravalloti (approved):
lgtm pending the outcome of your discussion with Natacha 👍 no need to re-tag me after you sort out her latest comments!

odbc/src/api/get_type_info_tests.rs
```diff
@@ -0,0 +1,216 @@
...
+                .unwrap()
+                .as_ref()
+                .unwrap()
+                .get_value(1);
+            assert!(value.is_err());
+        }
+    }
+
+    #[test]
+    fn test_get_all_type_info_row_values() {
```
nbagnard, line 150: Before submitting can you move this, test_unsupported_type_empty_list and test_specific_type_one_response into function.yaml, now that we have everything in place to utilize the test harness?
Please don't hate me, this is the last change needed and don't need to retag me before merging.

nbagnard (approved):
LGTM!

Please move the tests which can be from get_type_info_tests.rs to functions.yaml
You'll need to resolve conflict with master also before merging.
No need to retag me though. 

Pretty good work, thanks for taking the time to do the refactoring and making sure the type info are consistent everywhere in the code 👍

jchemburkar:
evergreen merge

PR #62, opened 2023-01-10
SQL-1154: Move dbs and collections and collection filtering on the client side

nbagnard:


core/src/collections.rs
```diff
@@ -85,29 +89,38 @@ impl MongoCollections {
...
-            databases.push(CollectionsForDb {
-                database_name: name,
-                collection_list: collections,
-            });
-        }
+            .list_database_names(None, None)
+            .unwrap()
+            .iter()
+            .filter(|val| {
+                if db_name_filter_regex.is_some() {
```
pmeredit, line 99: [nit] since we have `is_some` with an `else` I think this would look better just as a match (fewer braces).

core/src/collections.rs
```diff
@@ -137,25 +152,41 @@ impl MongoStatement for MongoCollections {
             self.current_database_index = Some(0);
         }
         loop {
-            if self
+            while self
```
pmeredit, line 155: I'm not sure why this is a `while` now instead of an `if`. Can you tell me a case where we need more than one iteration not already covered by the `loop`? Maybe the `loop` can be removed?

core/src/util/mod.rs
```diff
@@ -21,46 +22,40 @@ lazy_static! {
...
-}
-
 // Returns regex for a filter
-pub(crate) fn to_name_regex(filter: &str) -> Regex {
-    let regex_filter = pattern_to_regex(filter);
-    // If this ever fails it reflects a failure in pattern_to_regex, so this unwrap can stay
-    Regex::new(&regex_filter).unwrap()
+pub(crate) fn to_name_regex(filter: &str) -> Option<Regex> {
+    match filter {
+        "%" => None,
```
pmeredit, line 30: Nice

pmeredit (approved):
LGTM! Just a nit and a question .

core/src/collections.rs
pmeredit: Also means removing an `unwrap`!

bucaojit (approved):
Thanks for adding the client side filtering!  Looks good. 

Had a few comments.

core/src/util/mod.rs
```diff
@@ -21,46 +22,40 @@ lazy_static! {
...
+        _ => {
+            let table_type_entries = table_type
+                .split(',')
+                .map(|attr| attr.trim())
+                .collect::<Vec<&str>>();
+            let mut table_type_filters: Vec<CollectionType> = Vec::new();
+            for table_type_entry in &table_type_entries {
+                if TABLE_VALUES.is_match(table_type_entry) {
+                    // Collection and Timeseries types should be mapped to table
+                    // The Rust driver doesn't seem to deserialize timeseries at the moment because
```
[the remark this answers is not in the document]
nbagnard: I just tested and if you have a Timeseries, it fails with ` { kind: BsonDeserialization(DeserializationError { message: "unknown variant 'timeseries' expected 'view' or 'collection'" }), labels: {}, wire_version: None, source: None }`.
I am not sure if it is possible to create a TimeSeries for ADF. Not from the UI at least. But if Timeseries indeed are supported, it will be a problem. I thought that the "non-exhaustive" header just meant that you'd get a default value, but... apparently no.
For the line 224, I'd keep it because when Timeseries are indeed supported it will be fine and it still makes sense now.

core/src/collections.rs
nbagnard: The`loop` is for moving to the next database, the `while` is for moving to the next collection if the current one does not match the filter. We do need both of them because of the client-side filtering. Does that make sense?

core/src/collections.rs
nbagnard: I like that :D

core/src/collections.rs
pmeredit: Yep! thanks!

pmeredit (approved):
LGTM! Thanks!

nbagnard:
evergreen merge

PR #63, opened 2023-01-10
SQL-1159: Use client side filters for SQLColumns

pmeredit:
Also fixed a glaring issue in the test where it would succeed when no rows were returned, haha.

### DIFF
--- a/core/src/fields.rs
+++ b/core/src/fields.rs
@@ -7,9 +7,9 @@ use crate::{
         BsonTypeName,
     },
     stmt::MongoStatement,
-    util::{to_name_regex, to_name_regex_doc},
+    util::to_name_regex,
 };
-use bson::{doc, Bson, Document};
+use bson::{doc, Bson};
 use lazy_static::lazy_static;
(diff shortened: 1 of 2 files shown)


core/src/fields.rs
```diff
@@ -566,9 +568,9 @@ impl MongoStatement for MongoFields {
                 let filter = filter.clone();
                 loop {
                     self.current_field_for_collection += 1;
-                    if !(self.current_field_for_collection as usize)
-                        < self.current_col_metadata.len()
-                        || self.get_next_metadata(mongo_connection.unwrap())
+                    if (self.current_field_for_collection as usize
```
pmeredit, line 571: I think the logic is more obvious this way... also, not wrong (lol)

bucaojit:
@pmeredit I see that the client-side filtering for fields has been done in the [PR 62](https://github.com/mongodb/mongo-odbc-driver/pull/62) along with the query test change.  Is it not fully implemented in that PR?

PR #65, opened 2023-01-12
SQL-1168: Simplfiy Anyof(<one>) and double buffer

terakilobyte:
This PR simplifies schemas that are AnyOf one concrete type that we have the ability to show.

### DIFF
--- a/core/src/json_schema.rs
+++ b/core/src/json_schema.rs
@@ -233,21 +233,17 @@ pub mod simplified {
                     additional_properties: None,
                     items: None,
                     any_of: Some(any_of),
-                } => {
-                    if any_of.is_empty() {
-                        return Err(Error::InvalidResultSetJsonSchema);
-                    }
-                    // if there is an AnyOf that contains one item, we can simplify it
-                    if any_of.len() == 1 {
-                        return Schema::try_from(any_of.into_iter().next().unwrap());
-                    }
-                    Ok(Schema::AnyOf(
+                } => match any_of.len() {
+                    0 => Err(Error::InvalidResultSetJsonSchema),
+                    // AnyOf with a single schema is equivalent to the schema.
+                    1 => Schema::try_from(any_of.into_iter().next().unwrap()),
+                    _ => Ok(Schema::AnyOf(
                         any_of
                             .into_iter()
                             .map(Atomic::try_from)
                             .collect::<Result<BTreeSet<Atomic>>>()?,
-                    ))
-                }
+                    )),
+                },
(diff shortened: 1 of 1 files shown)


integration_test/tests/query_tests.rs
```diff
@@ -1,5 +1,6 @@
 mod common;
 
+#[cfg(target_os = "linux")]
```
terakilobyte, line 3: This test suite routinely segfaults on Windows. Attempting to troubleshoot this with @pmeredit led us to deciding to disable the test for Windows for now.

core/src/json_schema.rs
```diff
@@ -232,12 +233,21 @@ pub mod simplified {
...
+                    // if there is an AnyOf that contains one item, we can simplify it
+                    if any_of.len() == 1 {
+                        return Schema::try_from(any_of.into_iter().next().unwrap());
+                    }
+                    Ok(Schema::AnyOf(
+                        any_of
+                            .into_iter()
+                            .map(Atomic::try_from)
+                            .collect::<Result<BTreeSet<Atomic>>>()?,
+                    ))
```
pmeredit, line 249:
```suggestion
                    match any_of.len() {
                        0 => Err(Error::InvalidResultSetJsonSchema),
                         // if there is an AnyOf that contains one item, we can simplify it
                        1 => Schema::try_from(any_of.into_iter().next().unwrap()),
                        _ => Ok(Schema::AnyOf(
                            any_of
                                .into_iter()
                                .map(Atomic::try_from)
                                .collect::<Result<BTreeSet<Atomic>>>()?,
                        ))
                    }       
```

pmeredit: This cleans up the code a bit, imo.

pmeredit (changes requested):
Also, the double buffer change isn't in here. Perhaps this ticket should just be simplifying AnyOf anyway?

core/src/json_schema.rs
terakilobyte: agreed, thanks for the suggesion!

terakilobyte:
@pmeredit Yeah we (@nbagnard and I) decided to remove that, as looking at it more closely we identified some issues with number of chars vs number of bytes, and decided to split that out into separate work.

nbagnard (approved):
LGTM! 👍

pmeredit (approved):
LGTM! Just make sure to change the commit message

terakilobyte:
evergreen merge

pmeredit:
evergreen merge

PR #66, opened 2023-01-18
SQL-116[23]: Implement SQLPrimaryKeys and SQLForeignKeys

pmeredit:
It made sense to just do them both at once because they both use mostly the same code.


pmeredit:
Note, this should not cause any merge problems with @jchemburkar's PR

nbagnard:
> Note, this should not cause any merge problems with @jchemburkar's PR
That's true but we will have to go back and change things like
```
MongoColMetadata::new(
            "",
            "".to_string(),
            "PKTABLE_CAT".to_string(),
            Schema::Atomic(Atomic::Scalar(BsonTypeName::String)),
            Nullability::NULLABLE
        )
```
to
```
MongoColMetadata::new_metadata_from_bson_type_info(
            "",
            "".to_string(),
            "PKTABLE_CAT".to_string(),
            BsonTypeInfo::STRING,
            Nullability::NULLABLE
        )
```
after this is merged. 
We should keep an eye on who is going to merge first and whoever merges after can do the update.

core/src/stmt.rs
```diff
@@ -24,3 +24,55 @@ pub trait MongoStatement: Debug {
             .map_or(Err(Error::ColIndexOutOfBounds(col_index)), Ok)
     }
 }
+
+#[derive(Debug)]
+pub struct EmptyStatement {
+    pub resultset_metadata: Vec<MongoColMetadata>,
```
nbagnard, line 30:
```suggestion
    pub resultset_metadata: &'static Vec<MongoColMetadata>,
```
If you make this a static ref then there is no need for cloning the metadata and you can do this for example :
```
impl MongoPrimaryKeys {
    pub fn empty() -> EmptyStatement {
        EmptyStatement {
            resultset_metadata: &PK_METADATA,
        }
    }
}
```

odbc/src/api/functions.rs
```diff
@@ -1345,7 +1346,16 @@ pub unsafe extern "C" fn SQLForeignKeys(
     _fk_table_name: *const Char,
     _fk_table_name_length: SmallInt,
 ) -> SqlReturn {
-    unsupported_function(MongoHandleRef::from(statement_handle), "SQLForeignKeys")
+    panic_safe_exec!(
```
nbagnard, line 1349: There is a unit test for Empty statement, but I don't see any test for SQLForeignKeys and SQLPrimaryKeys.
Could you please add them in functions.yaml? I realized that right now only SQLTablesW is tested and I'd rather add them now then later. Plus this increases the odd of finding missing metadata columns or properties reported incorrectly.

core/src/stmt.rs
```diff
@@ -24,3 +24,55 @@ pub trait MongoStatement: Debug {
...
+}
+
+impl MongoStatement for EmptyStatement {
```
nbagnard, line 33: I like this! Can you use it in collections.rs for all_schemas too?
```
    pub fn all_schemas() -> EmptyStatement {
        EmptyStatement {
            resultset_metadata: &COLLECTIONS_METADATA,
        }
    }
```

nbagnard (changes requested):
Pretty straightforward as you said :)
Only a couple of small comments for me and it's good to go

pmeredit:
> > Note, this should not cause any merge problems with @jchemburkar's PR
> > That's true but we will have to go back and change things like
> 
> ```
> MongoColMetadata::new(
>             "",
>             "".to_string(),
>             "PKTABLE_CAT".to_string(),
>             Schema::Atomic(Atomic::Scalar(BsonTypeName::String)),
>             Nullability::NULLABLE
>         )
> ```
> 
> to
> 
> ```
> MongoColMetadata::new_metadata_from_bson_type_info(
>             "",
>             "".to_string(),
>             "PKTABLE_CAT".to_string(),
>             BsonTypeInfo::STRING,
>             Nullability::NULLABLE
>         )
> ```
> 
> after this is merged. We should keep an eye on who is going to merge first and whoever merges after can do the update.

`new` should still work, but it may be cleaner that way!

core/src/stmt.rs
pmeredit: Great idea!

odbc/src/api/functions.rs
bucaojit: At the moment `SQLForeignKeys` can be added to be tested in functions.yaml.  For some reason SQLForeignKeys is implented in `odbc-sys` but not `SQLPrimaryKeys`.  
We did have [SQL-1015](https://jira.mongodb.org/browse/SQL-1015) to investigate how to test these missing functions from the odbc-sys crate.

bucaojit (approved):
Thanks for getting to this work so quickly.  Changes look good, with a few minor comments.  
Nice addition of the `EmptyStatement` struct.

core/src/stmt.rs
pmeredit: Oh, good idea.

odbc/src/api/functions.rs
pmeredit: Yea, I'm not sure about adding a test for PrimaryKeys then, since all it would be doing is ensuring that the metadata matches what I put in.

pmeredit:
I don't know why the resultset-test is failing. It has started failing on master, also.

odbc/src/api/functions.rs
nbagnard: Got it!
It is always good to check that we are returning an empty resultset and the metadata correctly though. It checks that we don't fail on initializing them. 
ah, odbc-sys is not making our life easy :'(

nbagnard (approved):
LGTM! The empty RS is a great addition, thank you for that :)

pmeredit:
evergreen merge

PR #68, opened 2023-01-20
SQL-640: add integration test for power bi data retrieval flow

jchemburkar:


integration_test/tests/power_bi_tests.rs
```diff
@@ -244,4 +245,149 @@ mod integration {
...
+            let column_count_ptr = &mut 0;
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLNumResultCols(stmt as HStmt, column_count_ptr)
+            );
+            let numeric_attribute_ptr = &mut 0;
+            const FIELD_TYPES: [Desc; 4] = [
+                Desc::ConciseType,
+                Desc::Unsigned,
+                // SQL_COLUMN_NAME,
```
jchemburkar, line 329: I want to flag the `SQLColAttribute` calls, specifically the `SQL_COLUMN_*` field types. From the [comments section of the spec] (https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlcolattribute-function?source=recommendations&view=sql-server-ver16), it seems these are ODBC 2.x values that ODBC 3.x is expected to handle/map appropriately. If I am reading / understanding them correctly, I believe they don't exist in the [odbc_sys Desc enum](https://github.com/pacman82/odbc-sys/blob/master/src/desc.rs#L199). I see for some `SQLGetInfo` calls we ignore if an enum value is unavailable (and I do that in a couple places too). I wanted to check if that is appropriate here as well for the column attribute calls, or if we should consider opening a ticket.

bucaojit: Looking through my SQL trace log I see the ODBC 3.x field types being used: SQL_DESC_NAME, SQL_DESC_NULLABLE, SQL_DESC_SCALE, SQL_DESC_LENGTH.  I'm wondering if the flows were captured with an earlier version of our driver before we were reporting and ODBC version.  Are you still seeing these ODBC 2.x values @nbagnard ?  If not, then we should be able to use the SQL_DESC_* values as I see the ones we want here are available in odbc_sys.

nbagnard: The flows were captured with Patrick's POC and I just checked it does not report the SQL_DRIVER_ODBC_VERSION, so it is possible that it defaults to 2.0.
I also see the ODBC 3.x identifiers SQL_DESC_SCALE, SQL_DESC_LENGTH, SQL_DESC_NAME being used now with a recent trace.
I think that you can safely switch to using only SQL_DESC_XXX identifiers here.

integration_test/tests/power_bi_tests.rs
```diff
@@ -244,4 +246,149 @@ mod integration {
...
+            /*
+            SQLGetInfoW(SQL_DRIVER_ODBC_VER)
+            SQLGetInfoW(SQL_DRIVER_NAME)
+            */
+
+            let mut query: Vec<u16> = "select * from example".encode_utf16().collect();
+            query.push(0);
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLExecDirectW(stmt as HStmt, query.as_ptr(), SQL_NTS as i32)
```
nbagnard, line 298: If you don't mind, can you update the asserts for the API calls  slightly so that if the result is not SUCCESS, it prints out the diagnostic? You can look at power_bi_connect and SQLDriverConnectW

integration_test/tests/power_bi_tests.rs
```diff
@@ -244,4 +246,149 @@ mod integration {
...
+            const BUFFER_LENGTH: SmallInt = 300;
+            let output_buffer = &mut [0u16; (BUFFER_LENGTH as usize - 1)] as *mut _;
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLGetInfoW(
+                    conn_handle,
+                    InfoType::GetDataExtensions,
+                    output_buffer as Pointer,
+                    BUFFER_LENGTH,
```
nbagnard, line 316: Can you pass the length in bytes for SQLGetInfo and the other ones below. Patrick's PR is not in yet, but it seems safer to have these news ones right already

integration_test/tests/power_bi_tests.rs
```diff
@@ -244,4 +246,149 @@ mod integration {
...
+                    str_len_ptr
+                )
+            );
+
+            let column_count_ptr = &mut 0;
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLNumResultCols(stmt as HStmt, column_count_ptr)
```
nbagnard, line 324: It would be nice to also assert that the count is the expected one : 2. WDYT?

integration_test/tests/power_bi_tests.rs
```diff
@@ -244,4 +246,149 @@ mod integration {
...
+            );
+            let numeric_attribute_ptr = &mut 0;
+            const FIELD_TYPES: [Desc; 3] = [
```
nbagnard, line 327: Can you rename this FIELD_IDENTIFIER or FIELD_ID? This way there is no confusion with SQL or Mongo types.

integration_test/tests/power_bi_tests.rs
```diff
@@ -244,4 +246,149 @@ mod integration {
...
+                            BUFFER_LENGTH,
+                            str_len_ptr,
+                            numeric_attribute_ptr
+                        )
+                    );
+                });
+            }
+            loop {
+                let result = SQLFetch(stmt as HStmt);
+                assert!(result == SqlReturn::SUCCESS || result == SqlReturn::NO_DATA);
```
nbagnard, line 354: Please add the diagnostic if the assert fails.

integration_test/tests/power_bi_tests.rs
```diff
@@ -244,4 +246,149 @@ mod integration {
...
+                            *field_type,
+                            output_buffer as Pointer,
+                            BUFFER_LENGTH,
+                            str_len_ptr,
+                            numeric_attribute_ptr
+                        )
+                    );
+                });
+            }
+            loop {
```
nbagnard, line 352: Should we add a counter to make sure that fetch is called the expected number of times to exhaust all the rows?

nbagnard (changes requested):
I only have a few minor requests, nothing to do with the flow itself. 
I asked to check that the columns number and number of rows is the expected one. Technically, we should have integration tests which are already checking this out so I am Ok if you think it's not necessary at all. It seems easy enough to do though, so I thought "why not have this extra safety" especially if we are planning on not opening PowerBI and only run these tests

integration_test/tests/power_bi_tests.rs
jchemburkar: I will update these to the 3.0 versions. Thanks for the feedback

integration_test/tests/power_bi_tests.rs
```diff
@@ -88,6 +89,7 @@ mod integration {
 
             // Generate the connection string and add a null terminator because PowerBi uses SQL_NTS for the length
             in_connection_string = generate_default_connection_str();
+            in_connection_string.push_str("DATABASE=integration_test;");
```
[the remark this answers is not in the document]
jchemburkar: That logic in the default connection str function makes sense. The reason I added this was to account for the initial [evergreen failure](https://spruce.mongodb.com/task/mongosql_odbc_driver_ubuntu2004_asan_integration_test_patch_d78bd74fe454f48a55ad32daa2675ac2d097cfd1_63cb1c6fd1fe07422f8ffd83_23_01_20_22_57_51/logs?execution=0) when I opened as a draft, though it is possible I misdiagnosed. Is there somewhere I can check where the `ADF_TEST_LOCAL_DB` variable is set? I don't see it in the `evergreen.yaml`

integration_test/tests/power_bi_tests.rs
jchemburkar: 👍, would have been helpful for ^ also 😄

integration_test/tests/power_bi_tests.rs
jchemburkar: I think that is a good idea. I felt a bit uneasy about blindly looping, so that would make me feel more satisfied as well.

integration_test/tests/power_bi_tests.rs
jchemburkar: that sounds good to me!

integration_test/tests/power_bi_tests.rs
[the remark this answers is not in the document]
nbagnard: ahhhh, of course, because the change is in the draft PR [here](https://github.com/mongodb/mongo-odbc-driver/pull/59/files#diff-18df3e2ba4f534c8a4b8c1d123fb89bdeaa0229641993ac244e63ba3a106fc7c) ....
You need to add this line in "generate expansions" : `export ADF_TEST_LOCAL_DB="${adf_test_local_db}"`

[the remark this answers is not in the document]
jchemburkar: ah that makes sense -- will add this

bucaojit (approved):
LGTM!

integration_test/tests/power_bi_tests.rs
```diff
@@ -368,4 +370,165 @@ mod integration {
...
+            for col_num in 0..*column_count_ptr {
+                FIELD_IDS.iter().for_each(|field_type| {
+                    assert_eq!(
+                        SqlReturn::SUCCESS,
+                        SQLColAttributeW(
+                            stmt as HStmt,
+                            (col_num + 1) as u16,
+                            *field_type,
+                            output_buffer as Pointer,
+                            14,
```
nbagnard, line 468: I couldn't figure out where `14` is form. The actual buffer is length is BUFFER_LENGTH * size_of::<WideChar>, no?

jchemburkar: I was trying to roughly match how we did `get_info`, though it ended up getting a little messy because we are in a loop (so the outputs have varying sizes -- 14 being the size in bytes of the largest). happy to just make this `BUFFER_SIZE * size_of::` however if that is more clear, since I didn't end up checking the `str_len_ptr`.

integration_test/tests/power_bi_tests.rs
```diff
@@ -368,4 +370,165 @@ mod integration {
...
+                    result == SqlReturn::SUCCESS || result == SqlReturn::NO_DATA,
+                    "{}",
+                    get_sql_diagnostics(HandleType::Stmt, stmt as Handle)
+                );
+                match result {
+                    SqlReturn::SUCCESS => {
+                        successful_fetch_count += 1;
+                        assert_eq!(
+                            SqlReturn::SUCCESS,
+                            SQLGetData(
```
nbagnard, line 491: [nit] Technically, you could have a loop calling SQLGetData and using the info returned by the SQLColAttribute calls to pick the target type to ask for. This way you would not be dependent on you knowing that there are exactly 2 columns in the tables and that the first one is a long, the second a string

nbagnard (approved):
LGTM!
Simply change odbc_sys::CDataType::Char to odbc_sys::CDataType::WChar. No need to retag me.

I added a couple of suggestions but these are up to you. The test is still correct as it is.

integration_test/tests/power_bi_tests.rs
jchemburkar: I thought about this too, and wasn't sure whether that would be more or less clear. Thinking about it a bit more, I think it is worth doing so I can make this change!

jchemburkar:
evergreen merge

PR #69, opened 2023-01-21
SQL-1173: Move columns tests to the testrunner

pmeredit:
This moves all the columns tests to the testrunner and removes the old columns (and query) integration test.

Also fixes a bug where we were returning a pointer to memory that could be reclaimed by the system (this probably fixes our intermittent resultset test failure).


integration_test/src/test_runner/mod.rs
```diff
@@ -27,6 +27,8 @@ pub enum Error {
         test: String,
         expected: String,
         actual: String,
+        row: usize,
+        column: usize,
```
pmeredit, line 31: This change here was for my own sanity on fixing the results.

integration_test/src/test_runner/mod.rs
```diff
@@ -302,13 +310,46 @@ fn run_function_test(
...
+                    to_i16(&function[8])?,
+                ))
+            }
+        }
+        "sqlcolumns" => {
+            check_array_length(function, 9)?;
+            unsafe {
+                Ok(odbc_sys::SQLColumns(
+                    statement.handle() as HStmt,
+                    str_or_null(&function[1]),
```
pmeredit, line 329: the `SQLText::ptr` function used by `str_or_null` uses `mem::transmute`... so it does not have the same issue. It may or may not be leaking memory. I have to investigate further to be sure.

odbc/src/api/functions.rs
```diff
@@ -694,7 +694,7 @@ pub unsafe extern "C" fn SQLColumnsW(
             let mongo_handle = MongoHandleRef::from(statement_handle);
             let stmt = must_be_valid!((*mongo_handle).as_statement());
             let catalog_string = input_wtext_to_string(catalog_name, catalog_name_length as usize);
-            let catalog = if catalog_name.is_null() {
+            let catalog = if catalog_name.is_null() || catalog_string.is_empty() {
```
pmeredit, line 697: This was a bug... I think. Either way the empty database is not a valid database name in mongo, so this is safe.

core/src/fields.rs
```diff
@@ -665,7 +665,7 @@ impl MongoStatement for MongoFields {
                     Some(_) => 0i32,
                 }
             }),
-            17 => Bson::Int32(self.current_field_for_collection as i32),
+            17 => Bson::Int32(1 + self.current_field_for_collection as i32),
```
pmeredit, line 668: Ordinal position should actually be 1-indexed!

terakilobyte (approved):
LGTM. Noticed two small things that don't need a re-review.

nbagnard (approved):
LGTM! The only thing I am not sure is if tests need to be split in different files. I would rather have them group in one file and create one for SQLColumns, one for SQLTables, etc...

integration_test/src/test_runner/mod.rs
bucaojit: Nice, that will be helpful.

integration_test/src/test_runner/mod.rs
```diff
@@ -123,14 +125,14 @@ impl fmt::Display for TestDef {
 /// integration_test runs the query and function tests contained in the TEST_FILE_DIR directory
```
bucaojit, line 125: small nit: Let's update the comment to match the function name change

resources/integration_test/tests/columns_no_filter.yml
```diff
@@ -0,0 +1,251 @@
...
+    db: integration_test
+    expected_result:
+      - ["integration_test", null, "example", "_id", 4, "int", 10, 4, 0, 10, 1, "", null, "{\"$numberInt\":\"4\"}" , null , 0, 1, "YES"]
+      - ["integration_test", null, "example", "b", 12, "string", 0, 4294967292, 0, null, 1, "", null, 12, null, 4294967292, 2, "YES"]
+      - ["integration_test", null, "foo", "_id", 4, "int", 10, 4, 0, 10, 1, "", null, "{\"$numberInt\":\"4\"}" , null , 0, 1, "YES"]
+      - ["integration_test", null, "foo", "a", 4294967291, "long", 19, 8, 0, null, 1, "", null, 4294967291, null, 0, 2, "YES"]
+      - ["integration_test_2", null, "example_2", "_id", 8, "double", 15, 8, 15, 10, 1, "", null, 8 , null , 0, 1, "YES"]
+      - ["integration_test_2", null, "example_2", "b", 12, "string", 0, 4294967292, 0, null, 1, "", null, 12, null, 4294967292, 2, "YES"]
+    expected_bson_type: [
+        "string",
```
bucaojit, line 13: Looks like the output here comes from the test generator, do you know if there is a way to format the YAML output to have these lists in one line instead of a newline for each entry?  I checked briefly at `serde_yaml::to_writer()` and couldn't find a way, but it would make the file more readable.  No worries if it's not a quick fix.

bucaojit: One other suggestion is if the metadata for all the SQLColumnsW calls are the same, we can do like the JDBC tests and have only one of the tests check all the metadata.  The remaining `SQLColumnsW` will then only need to check that the results are correct (and skip the metadata checking).  This would be clearer that the metadata is checked by one of the tests if these were in one SQLColumnsW yaml file as Natacha suggests.

bucaojit (approved):
Thanks for fixing the memory leak issue!

LGTM, I had a few suggestions that you can implement if it would be helpful.

resources/integration_test/tests/columns_no_filter.yml
pmeredit: I made these manually. I didn't want them on one line.

pmeredit: Oh, I don't have to check the metadata! Yes!

pmeredit:
evergreen merge

PR #70, opened 2023-01-23
SQL-1150: Clean up missing StatmentAttrs, support SQL_ROWSET_SIZE

pmeredit:
Guess the biggest question is if we want to rename SQL_ATTR_ROWSET_SIZE even though that is no the name in the header (for some reason they did not redefine this one, probably because they use SQL_ATTR_ROW_ARRAY_SIZE to seemingly mean the same thing).


odbc/src/api/definitions.rs
```diff
@@ -153,43 +153,48 @@ pub enum EnvironmentAttribute {
 #[repr(i32)]
 #[derive(Debug, PartialEq, Eq, Clone, Copy, FromPrimitive)]
 pub enum StatementAttribute {
-    SQL_ATTR_APP_ROW_DESC = 10010,
-    SQL_ATTR_APP_PARAM_DESC = 10011,
-    SQL_ATTR_IMP_ROW_DESC = 10012,
-    SQL_ATTR_IMP_PARAM_DESC = 10013,
     SQL_ATTR_CURSOR_SCROLLABLE = -1,
     SQL_ATTR_CURSOR_SENSITIVITY = -2,
```
pmeredit, line 157: I reordered these by number to make it easier to see what we were missing.

odbc/src/api/functions.rs
```diff
@@ -2824,9 +2824,23 @@ unsafe fn sql_get_stmt_attrw_helper(
                 SqlReturn::SUCCESS
             }
             StatementAttribute::SQL_ATTR_METADATA_ID => {
-                todo!();
+                // False means that we treat arguments to catalog functions as case sensitive. This
```
pmeredit, line 2827: This is a driveby fix of a todo that we didn't seem to have a ticket for.

odbc/src/api/definitions.rs
mattChiaravalloti: Good idea. I wasn't sure why it was in this order in odbc_sys in the first place! It was helpful during #64 to keep this order since it's the same order we used in the `functions.rs` `match` statement, but switching it to numeric order is super useful 🧠

odbc/src/api/functions.rs
```diff
@@ -2824,9 +2824,23 @@ unsafe fn sql_get_stmt_attrw_helper(
...
-            _ => {
+            // leave SQL_GET_BOOKMARK as unsupported since it is for ODBC < 3.0 drivers
+            StatementAttribute::SQL_GET_BOOKMARK => {
+                err = Some(ODBCError::UnsupportedStatementAttribute(
+                    statement_attribute_to_string(attribute),
+                ));
+                SqlReturn::ERROR
+            }
+            // Not supported but still relevent to 3.0 drivers
+            StatementAttribute::SQL_ATTR_SAMPLE_SIZE
```
mattChiaravalloti, line 2840:
```suggestion
            // leave SQL_GET_BOOKMARK as unsupported since it is for ODBC < 3.0 drivers
            StatementAttribute::SQL_GET_BOOKMARK 
            // Not supported but still relevent to 3.0 drivers
            | StatementAttribute::SQL_ATTR_SAMPLE_SIZE
```
We could still combine these into one set and have the comment differentiate them. No need to duplicate the `err = ...; SqlReturn::ERROR` code.

odbc/src/api/definitions.rs
```diff
@@ -153,43 +153,48 @@ pub enum EnvironmentAttribute {
...
+    SQL_ATTR_NOSCAN = 2,
+    SQL_ATTR_MAX_LENGTH = 3,
     SQL_ATTR_ASYNC_ENABLE = 4,
-    SQL_ATTR_CONCURRENCY = 7,
+    SQL_ATTR_ROW_BIND_TYPE = 5,
     SQL_ATTR_CURSOR_TYPE = 6,
+    SQL_ATTR_CONCURRENCY = 7,
+    SQL_ATTR_KEYSET_SIZE = 8,
+    // Never renamed to SQL_ATTR_ROWSET_SIZE
+    SQL_ROWSET_SIZE = 9,
```
mattChiaravalloti, line 168: It's unfortunate, but I don't think we should change the name. My preference is to stick with the (weird) conventions from the header files.

odbc/src/api/functions.rs
```diff
@@ -2824,9 +2824,23 @@ unsafe fn sql_get_stmt_attrw_helper(
                 SqlReturn::SUCCESS
             }
             StatementAttribute::SQL_ATTR_METADATA_ID => {
-                todo!();
+                // False means that we treat arguments to catalog functions as case sensitive. This
+                // is a _requirement_ for mongodb where FOO and foo are distinct database names.
+                *(value_ptr as *mut ULen) = SqlBool::False as ULen;
+                SqlReturn::SUCCESS
             }
-            _ => {
```
mattChiaravalloti, line 2829: Should we also do this for `SQLSetStmtAttrW`?

mattChiaravalloti (changes requested):
Looks good but marking as changes requested since I'm curious about `SQLSetStmtAttrW` 🧐

odbc/src/api/definitions.rs
pmeredit: agreed

pmeredit:
evergreen merge

pmeredit:
evergreen merge

mattChiaravalloti (approved):

PR #73, opened 2023-01-25
SQL-1176: corrects odbc version; updates driver version test

terakilobyte:
I extracted the odbc version to a constant located with other constants.

Also, while I was lookihng at the test suite I noticed that we were hard coding a version expectation, even though we derive the driver version from our cargo.toml file. I updated the the associated tests.

### DIFF
--- a/constants/src/lib.rs
+++ b/constants/src/lib.rs
@@ -1,6 +1,7 @@
 pub const VENDOR_IDENTIFIER: &str = "MongoDB";
 pub const DRIVER_NAME: &str = "MongoDB Atlas SQL interface ODBC Driver";
 pub const DBMS_NAME: &str = "MongoDB Atlas";
+pub const ODBC_VERSION: &str = "03.80";
 
 // SQL states
 pub const NOT_IMPLEMENTED: &str = "HYC00";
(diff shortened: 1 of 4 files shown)


jchemburkar (commented):
Changes LGTM! One place I noticed / was curious if it belongs in this pr is the env handle attribute, enum defined [here](https://github.com/mongodb/mongo-odbc-driver/blob/a73cc98a16b89147c129cd8a6b976ae4b2f07839/odbc/src/api/definitions.rs#L14). Do we want this to default to `ODBC_VERSION` as well (formatting aside, since its slightly different)?

nbagnard:
Not sure why I got a notification about this comment specifically, I don't see you tagged me 🤔
Here is an explanation though:
> Changes LGTM! One place I noticed / was curious if it belongs in this pr is the env handle attribute, enum defined [here](https://github.com/mongodb/mongo-odbc-driver/blob/a73cc98a16b89147c129cd8a6b976ae4b2f07839/odbc/src/api/definitions.rs#L14). Do we want this to default to `ODBC_VERSION` as well (formatting aside, since its slightly different)?

These are different info. The client sets which behavior it wants the driver to exhibit with the Environment attribute. We are actually not honoring our part of the contract here and if a client set the odbc version to 2.0 we don't care and still return newer type names and sqlstaes.

bucaojit (approved):
LGTM!

terakilobyte:
evergreen merge

terakilobyte:
evergreen merge

PR #74, opened 2023-01-26
SQL-1185: Fix SQLTablesW not listing table names

bucaojit:
This change fixes an issue seen in Power BI where no tables were being listed with the call:
```
SQLTablesW("%", 1, null, 0, null, 0, "TABLE,VIEW", 10)
```
Added a result set test to verify the sqltables call. 

There are quite a few clippy errors due to a change in Rust allowing variables in the format! string.  Will work on addressing that.

### DIFF
--- a/core/src/col_metadata.rs
+++ b/core/src/col_metadata.rs
@@ -320,8 +320,8 @@ mod unit {
 
             match actual {
                 Err(Error::InvalidResultSetJsonSchema) => (),
-                Err(e) => panic!("unexpected error: {:?}", e),
-                Ok(ok) => panic!("unexpected result: {:?}", ok),
+                Err(e) => panic!("unexpected error: {e:?}"),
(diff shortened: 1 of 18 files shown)


nbagnard:
> There are quite a few clippy errors due to a change in Rust allowing variables in the format! string. Will work on addressing that.

oh yeah, I see what you mean. I saw changes from Nathan this week and I thought he really cared about having the variables in the {} directly, and then yesterday I understood that it was Clippy who forced his hand when I had to go over mongosql-rs code 😅
I feel like the message should not be "variables can be used directly in the `format!` string" but "variables **have to** be used directly in the `format!` string".

nbagnard (approved):
LGTM! 💯

terakilobyte (approved):
🔥 LGTM

bucaojit:
evergreen merge

PR #78, opened 2023-01-31
SQL-1125: Add tdvt data set

pmeredit:
The TPC-H Data sets are quite large and written with sql insert statements, so I decided to just go with tdvt first. If we really want TPC-H also, I can probably import them into mysql and dump csv files for use with my simple python script here.


pmeredit:
```db.runCommand({"aggregate": 1, "pipeline": [{"$sql": {"format": "odbc", "formatVersion": 1, "statement": "SELECT `Tax Rate` FROM staples LIMIT 1"}}], "cursor": {}})
{
  ok: 1,
  cursor: {
    firstBatch: [ { '': { 'Tax Rate': '0.02' } } ],
    id: Long("0"),
    ns: 'cursors.DESKTOP-BQT9LTI-f2ed96e6-7367-4328-a6bd-3507c44704d0'
  }
}```

**Comment by pmeredit** [comment]
I removed the columns test with no filters because it has too many results and will break every time we add more data. I think having one filter turned on each test provides enough coverage.

File: resources/integration_test/tests/columns_column_filter.yml [hunk 1]
```diff
@@ -1,8 +1,11 @@
 tests:
   - description: SQLColumnsW column filter
-    test_definition: [ "sqlcolumnsw", "", 0, "", 0, "", 0, "%i%", 3 ]
+    test_definition: [ "sqlcolumnsw", "", 0, "", 0, "", 0, "%id%", 3 ]
```
pmeredit, line 3: changed this so we don't end up with 37 rows.

nbagnard:
I am surprised that the tests in sqltables.yml are not failing.
This change adds a new DB and new collections. It should changes the catalog results.

I think though that you are right and TPC-H/TPC-DS is not something we should be bothering with at the moment. 

I am wondering also exactly what we we had in mind with using TDVT data. @bucaojit do you remember? Is it about trying out queries targeting the SQL functions and see how we behave? Because if I understand correctly this is not really a concern with the import mode.

It feels like for manual testing, having a dataset which covers all Mongo data types and has a schema close to "real" customer data would be the best? I know our spec said TPC and TDVT dataset, but it now seems a little off.
Sample_supplies might be a good start, especially if Alexi uses it for her demo. 
WDYT?

nbagnard:
> I am surprised that the tests in sqltables.yml are not failing.
> This change adds a new DB and new collections. It should changes the catalog results.

Forget this part. Github was just not showing the files because there were not there when I started looking at it.

pmeredit:
I'm fine with that data set too! I do need to put some decimal128 in here if we decide to go with tdvt data.

nbagnard (approved):
If @bucaojit  thinks that having tdvt as one of the integration tests data set is the way, then I am fine with this.
I was thinking more of dumps to be load data before doing the manual testing, but this works too as long as it is not creating problems for testing the driver itself.
I expect the load time to be quick and not make our tests much slower, right?

bucaojit:
@nbagnard When we asked the Microsoft team about suggestions for verification testing, I don't think we specified that the test would be for import mode only and so they suggested the more thorough TPC type testing.  I did not know what types of queries would be run through import mode at the time.  We do have the option of running native queries and additional datasets would be helpful for that and using Power BI will be more interesting with a larger dataset like this.

I agree that we will also want to add datasets with the different Mongo datatypes that we report as SQL_UNKNOWN_TYPE to test those come across as expected for the user.  We can test the Power BI functions such as Split Columns and Parse to process the complex types.

bucaojit (approved):
LGTM thanks for adding this dataset.

I checked the loading time and took 10 seconds for the larger `staples` dataset which is acceptable compared to the overall 6 minutes test runtime.

pmeredit:
evergreen merge

PR #80, opened 2023-02-01
properly handle nested decimal128; fix bug in cache insertion

terakilobyte:
This PR handles the case where Decimal128 values might be nested inside of arrays or objects.

I also found a bug (by chance, I encountered it because of the changes to the lengths of strings we were expecting) with the way we inserted data into the cache and addressed it in this commit d87678b3b8200e119cf7b92c0e571d5e1a58560c


nbagnard (approved):
LGTM (since we worked on it together yesterday, I already knew the code and this was one of the fastest review I did lately)! 

If you want to change `into_canonical_extjson` to `into_relaxed_extjson` that works for me. No need to retag me ;)

bucaojit (approved):
LGTM!

terakilobyte:
evergreen merge

PR #81, opened 2023-02-02
SQL-1191: Fix types for tdvt data and simplify/fix schema handling of Multiple bson_type

pmeredit:
This adds the proper data types to batters and calcs. Staples with proper data types will be added in a follow on patch under the same ticket because evergreen has patch size limits (it works fine locally).

This also fixes the issue we had with multiple types in schemata. Note that I removed 3 unit tests that required preconditions we now disallow.

Note, you don't need to look at more than the first few lines of each yml file, they were produced with a script! :D


resources/integration_test/testdata/calcs.yml
```diff
@@ -2,458 +2,576 @@ dataset:
...
+    num0: 12.3
+    num1: 8.42
+    num2: 17.86
+    num3: -11.52
+    num4: null
     str0: FURNITURE
     str1: CLAMP ON LAMPS
     str2: one
     str3: e
     time0: '1899-12-30 21:07:32'
```
pmeredit, line 38: I decided not to convert the time types since we don't actually support time and they are mixed dates and datetimes. I think it's fine to leave them as strings.

resources/integration_test/tests/columns_column_filter.yml
```diff
@@ -8,4 +8,3 @@ tests:
       - ["integration_test_2", null, "example_2", "_id", 8, "double", 15, 8, 15, 10, 1, "", null, 8 , null , 0, 1, "YES"]
       - ["tdvt", null, "batters", "_id", 0, "objectId", 24, 24, 0, null, 1, "", null, "{\"$numberInt\":\"0\"}" , null , 0, 22, "YES"]
       - ["tdvt", null, "calcs", "_id", 0, "objectId", 24, 24, 0, null, 1, "", null, "{\"$numberInt\":\"0\"}" , null , 0, 1, "YES"]
-      - ["tdvt", null, "staples", "_id", 0, "objectId", 24, 24, 0, null, 1, "", null, "{\"$numberInt\":\"0\"}" , null , 0, 51, "YES"]
```
pmeredit, line 11: This will be added back in the follow-on patch, want a green evg.

core/src/err.rs
```diff
@@ -15,8 +15,8 @@ pub enum Error {
     ColIndexOutOfBounds(u16),
     #[error("Invalid cursor state: cursor not advanced")]
     InvalidCursorState,
-    #[error("Result set metadata JSON schema must be object with properties")]
-    InvalidResultSetJsonSchema,
```
pmeredit, line 19: Added a reason here since there are multiple causes of this now.

core/src/json_schema.rs
```diff
@@ -112,7 +218,6 @@ pub mod simplified {
     // simplified::Schema to disallow nested AnyOfs.
     #[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone)]
     pub enum Atomic {
-        Any,
```
pmeredit, line 115: There's no reason to treat this separately from just allowing `BsonTypeName::Any`. I initially was converting `Scalar(Any)` to `Any`, and realized there is no advantage to a top level `Any` in this case (still makes sense in mongosql, I think, for brevity).

core/src/json_schema.rs
```diff
@@ -157,54 +266,41 @@ pub mod simplified {
...
+                } => Ok(Atomic::Scalar(BsonTypeName::Any)),
                 json_schema::Schema {
                     bson_type: Some(bson_type),
                     properties,
                     required,
                     additional_properties,
                     items,
                     any_of: None,
-                } => {
-                    let construct_obj = || {
```
pmeredit, line 169: this `construct_obj` closure made sense before I moved the `Multiple` code out of here. Now it isn't necessary :D

core/src/col_metadata.rs
```diff
@@ -274,7 +274,7 @@ impl ObjectSchema {
```
pmeredit, line 274: No real changes in this file, just fallout from the refactor and addition of error messages to the ResultSetSchema error

resources/integration_test/testdata/calcs.yml
bucaojit: I tried this in Power BI and it was able to transform the strings to date, datetime, and time by using `Change Type` on the column.  So it is fine to have here as strings.

bucaojit (approved):
Very nice solution to replacing the `Multiple` with any_of. 

Changes look good and the tdvt test set will be useful.

core/src/json_schema.rs
```diff
@@ -282,6 +363,297 @@ pub mod simplified {
...
+                additional_properties: None,
+                items: None,
+                any_of: Some(vec![
+                    Schema {
+                        bson_type: Some(BsonType::Single(BsonTypeName::Int)),
+                        properties: None,
+                        required: None,
+                        additional_properties: None,
+                        items: None,
+                        any_of: None,
```
mattChiaravalloti, line 399:
```suggestion
                        ..Default::default()
```
[nit] we could make these tests a marginally more readable by shrinking them in this way wherever we just want to set one or two fields and make the rest `None` -- up to you if you think it is worth the change!

core/src/json_schema.rs
```diff
@@ -19,6 +19,110 @@ pub struct Schema {
     pub any_of: Option<Vec<Schema>>,
 }
 
+impl Schema {
+    // Remove multiple recursively removes Multiple Bson Type entries.
+    pub fn remove_multiple(mut self) -> Result<Self, Error> {
+        // it is invalid for both any_of and bson_type to be set because
+        // any_of is a top level constructor.
+        if !(self.bson_type.is_none() || self.any_of.is_none()) {
```
mattChiaravalloti, line 27:
```suggestion
        if self.bson_type.is_some() && self.any_of.is_some() {
```
Personal opinion: positive conditionals are easier to read. "If bson_type and any_of are both Some, then ...", as opposed to "If not (bson_type is None or any_of is None), then ..."

core/src/json_schema.rs
```diff
@@ -217,24 +313,11 @@ pub mod simplified {
...
-                    bson_type: None,
-                    properties: None,
-                    required: None,
-                    additional_properties: None,
-                    items: None,
-                    any_of: Some(any_of),
-                } => match any_of.len() {
-                    0 => Err(Error::InvalidResultSetJsonSchema),
+            let schema = schema.remove_multiple()?;
+            match schema.any_of {
```
mattChiaravalloti, line 317: I love how this cleaned up 💯

core/src/json_schema.rs
```diff
@@ -282,6 +363,297 @@ pub mod simplified {
...
+                }),
+                required: Some(vec!["x".into(), "y".into()]),
+                additional_properties: Some(false),
+                items: None,
+                any_of: None,
+            }
+        );
+
+        remove_multiple_test!(
+            nullable_object,
```
mattChiaravalloti, line 451:
```suggestion
            object_and_other,
```
Arguably, the fact that the other type is "Null" does not matter. This test is just showing that only the "Object" gets the object-related fields, and the other type(s) do not.

core/src/json_schema.rs
```diff
@@ -282,6 +363,297 @@ pub mod simplified {
...
+                }),
+                required: Some(vec!["x".into(), "y".into()]),
+                additional_properties: Some(false),
+                items: None,
+                any_of: None,
+            }
+        );
+
+        remove_multiple_test!(
+            nullable_array,
```
mattChiaravalloti, line 497:
```suggestion
            array_and_other,
```

core/src/json_schema.rs
```diff
@@ -282,6 +363,297 @@ pub mod simplified {
...
+            #[test]
+            fn $func_name() {
+                let res = $input.remove_multiple().unwrap();
+
+                assert_eq!(res, $expected);
+            }
+        };
+    }
+
+    // Testing TryFrom<json_schema::Schema> for json_schema::simplified::Atomic
```
mattChiaravalloti, line 377: This comment should either be removed or moved to be directly above `mod try_from_test`.

core/src/json_schema.rs
```diff
@@ -282,6 +363,297 @@ pub mod simplified {
...
+    mod remove_multiple {
```
mattChiaravalloti, line 378: Other test ideas if you think they are worthwhile:
- More than 2 types in multiple
- Array with Items: Some(Multiple[...]) that turns into a Some(Single(Any))

core/src/json_schema.rs
```diff
@@ -282,6 +363,297 @@ pub mod simplified {
...
+                    additional_properties: Some(false),
+                    items: None,
+                    any_of: None,
+                }))),
+                any_of: None,
+            }
+        );
+    }
+
     macro_rules! try_from_test {
```
mattChiaravalloti, line 657: Did you mention that there was a specific bug you encountered that prompted this change, or is this just a refactor that improves the `try_from` code? If there was a particular bug, should we add a regression unit test for one or both of the `try_from` functions that shows it is now fixed?

mattChiaravalloti (changes requested):
The `json_schema` change is great! I have some thoughts on testing that I think are all minor improvements, but none of them are very consequential. Feel free to reject some or all of them -- just lmk and I can re-review with an approval instead 👍

core/src/json_schema.rs
pmeredit: whoops!

pmeredit, line 377:
```suggestion
```

core/src/json_schema.rs
```diff
@@ -282,6 +363,297 @@ pub mod simplified {
...
+                    required: Some(vec!["x".into(), "y".into()]),
+                    additional_properties: Some(false),
+                    items: None,
+                    any_of: None,
+                }))),
+                any_of: None,
+            }
+        );
+    }
+
```
pmeredit, line 651:
```suggestion
// Testing TryFrom<json_schema::Schema> for json_schema::simplified::Atomic
```

core/src/json_schema.rs
pmeredit: Yes, there was a multiple in a context we didn't expect (I forget exactly what) and I couldn't find a good way to handle it. Initially I thought, "well, I'll just convert to any of" but that then required called back up to the TryFrom for Schema when the context was in Atomic already (since Atomic can't handle AnyOf). At that point I decided the easiest thing to do would be to remove them all a priori, and that that would also simplify the simplification code.

mattChiaravalloti: Worth adding a try_from regression test for that, or is the remove_multiple testing enough?

pmeredit:
evergreen merge

pmeredit:
evergreen merge

PR #79, opened 2023-02-01
SQL-986: add ODBC result set query tests to test harness

jchemburkar:
- Uses [data](https://github.com/mongodb/mongo-jdbc-driver/blob/master/resources/integration_test/testdata/integration.yml) from JDBC test harness, to test similar test cases (found [here](https://github.com/mongodb/mongo-jdbc-driver/blob/master/resources/integration_test/tests/query.yaml)).
- Adds boolean and double types to test runner.

PR #82, opened 2023-02-04
SQL-987: add function tests to test harness

jchemburkar:
Adding tests for functions, notably the various cases of `SQLTables` (using [jdbc tests](https://github.com/mongodb/mongo-jdbc-driver/blob/master/resources/integration_test/tests/dbmd_variable_result_sets_data_methods.yml) and [this match statement](https://github.com/mongodb/mongo-odbc-driver/blob/4450c325ef7acf2aad68c67d687903516341917b/odbc/src/api/functions.rs#L3919) for context).

Noting that there is intentionally no test for `SQLColumns` with no filter, per [this pr](https://github.com/mongodb/mongo-odbc-driver/pull/78#issuecomment-1411262068)

This will need to be updated after the merging of [SQL-986](https://github.com/mongodb/mongo-odbc-driver/pull/79) and [SQL-1191](https://github.com/mongodb/mongo-odbc-driver/pull/81).


nbagnard (approved):
LGTM! 💯 
Be aware that there might changes coming from new dataset(s) and you could have to fix the expected result sets if the changes are in before.

resources/integration_test/tests/functions.yml
```diff
@@ -48,3 +48,28 @@ tests:
     expected_precision: [ 0, 10, 10, 0, 0, 0, 10, 10, 10, 10, 10, 10, 0, 10, 10, 10, 10, 10, 10 ]
     expected_scale: [ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]
     expected_nullability: [ 0, 0, 1, 1, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1, 1, 0, 1, 1, 1 ]
+  - description: SQLGetTypeInfo function getall types
```
bucaojit, line 51: Very cool, you were able to replace the GetTypeInfo unit tests with an entry here.

resources/integration_test/tests/functions.yml
```diff
@@ -48,3 +48,28 @@ tests:
...
+    test_definition: [ "sqlgettypeinfo", 0 ]
+    db: integration_test
+    expected_result:
+    - [ "string", 4294967287, null, "'", "'", null, 1, 1, 3, null, 0, null, "string", null, null, 4294967287, null, null, null ]
```
bucaojit, line 55: That's odd, any idea why these `DATA_TYPE` values are in this format instead of a negative value?

resources/integration_test/tests/functions.yml
bucaojit, line 51:
```suggestion
  - description: SQLGetTypeInfo function get all types
```

resources/integration_test/tests/sqltables.yml
```diff
@@ -1,5 +1,38 @@
...
+  - description: SQLTablesW SQL_ALL_SCHEMA returns empty list
+    test_definition: [ "sqltablesw", "", 0, "%", 1, "", 0, "", 0 ]
+    db: integration_test
+    expected_result: []
+
+  - description: SQLTablesW SQL_ALL_TABLE_TYPES lists valid types
+    test_definition: [ "sqltablesw", "", 0, "", 0, "", 0, "%", 1 ]
+    db: integration_test
+    expected_result:
+      - [ null, null, null, "TABLE", null ]
```
bucaojit, line 32: Looks like we have a small bug here, the 5th value should be an empty string `""`.  
Would you please create a ticket for that and add a skip reason here with the JIRA value?

bucaojit (changes requested):
Nice work with the SqlTables and GetTypeInfo tests.

I had a few small suggestion.  Also, as this ticket is related to the function runner, would you also remove the `sqltableprivileges` function [here](https://github.com/mongodb/mongo-odbc-driver/blob/master/integration_test/src/test_runner/mod.rs#L424).  I was looking at our implemented functions and see that this one will not be supported and does not need to be in the list of functions to eventually test.

resources/integration_test/tests/sqltables.yml
jchemburkar: Is there any preference between adding a ticket and making the chance in this pr? I think it will be a one line difference either way, so happy to do whichever makes more sense.

resources/integration_test/tests/functions.yml
jchemburkar: I _believe_ it is because [here](https://github.com/mongodb/mongo-odbc-driver/blob/7ac6e29497586f4af0e744cb10f06cd6fc45534c/integration_test/src/test_runner/mod.rs#L725) we cast to i64 (as opposed to i32), changing that line seems to yield the negative value. There are some existing tests structured this way too that I had followed, but I can update this + existing tests if you think that cast should be changed.

resources/integration_test/tests/sqltables.yml
jchemburkar: And thanks for catching this discrepancy 😄 Looking at the spec again, I think this test might be correct and it is the `SQL_ALL_CATALOGS` that needs updating: 
```
If TableType is SQL_ALL_TABLE_TYPES and CatalogName, SchemaName, and TableName are empty strings, the result set contains a list of valid table types for the data source. (All columns except the TABLE_TYPE column contain NULLs.)
```
Either way, the above question still stands on if we should update as part of this PR.

resources/integration_test/tests/functions.yml
bucaojit: I see.  Some of the existing tests would then benefit from it being cast to i32 as well to have the expected negative value instead of the two's complement.  Yes, let's update it to casting to i32.

resources/integration_test/tests/sqltables.yml
bucaojit: Good call on double-checking the spec.  
Let's add the change in this PR if it is not too much effort (sound like it shouldn't be).  I ran a quick test of setting remarks to null and table listing worked as expected through Power BI.

jchemburkar: Sounds good, I will add this

bucaojit (approved):
Awesome!  Thanks for making the suggested changes 👍

jchemburkar:
evergreen merge

PR #83, opened 2023-02-04
SQL-642: Add Power BI Flow for Table Listing

bucaojit:


integration_test/tests/power_bi_tests.rs
```diff
@@ -531,4 +531,185 @@ mod integration {
             );
         }
     }
+
+    /// Test PowerBI flow for listing tables
+    ///     - SQLAllocHandle(SQL_HANDLE_STMT)
+    //      - SQLTablesW(null pointer, null pointer, null pointer, "TABLE,VIEW")
+    //      - SQLGetFunctions(SQL_API_SQLFETCHSCROLL)
```
nbagnard, line 538: The flow from SQLGetFunctions to SQLFetch (exclusive) is the same than for test_data_retrieval()
What about we extract the duplicate part in a helper?
It also true for the SQLFetch to SQLMoreResults part. I can see it be bit more cumbersome since it needs to deal with the target C type for each column, but passing in an array of the targeted C types for each column could be enough.

integration_test/tests/power_bi_tests.rs
```diff
@@ -531,4 +531,185 @@ mod integration {
...
+                SQLTablesW(
+                    stmt as HStmt,
+                    null_mut(),
+                    0,
+                    null_mut(),
+                    0,
+                    null_mut(),
+                    0,
+                    table_view.as_ptr(),
+                    10
```
nbagnard, line 586: [nit] Why not using table_view.length - 1 ?

integration_test/tests/power_bi_tests.rs
```diff
@@ -531,4 +531,185 @@ mod integration {
...
+                            *field_type,
+                            output_buffer as Pointer,
+                            BUFFER_LENGTH,
+                            str_len_ptr,
+                            numeric_attribute_ptr,
+                        ),
+                        "{}",
+                        get_sql_diagnostics(HandleType::Stmt, stmt as Handle)
+                    );
+                    assert_eq!(
```
nbagnard, line 640: There is already a good coverage for SQLColAttribute tests via the test harness and tests in queries.yaml. I think that you can skip this validation. 
The PowerBi flows are here more to make sure that we can calls the same functions, with similar input arguments, and this way at least know things are not broken. The other issue I see is that everytime we add a new database or more tables, the `expected_col_attr_string_lengths` array will need to be amended.

integration_test/tests/power_bi_tests.rs
```diff
@@ -531,4 +531,185 @@ mod integration {
...
+                                    col_num + 1,
+                                    odbc_sys::CDataType::WChar,
+                                    output_buffer as Pointer,
+                                    BUFFER_LENGTH as Len,
+                                    str_len_ptr
+                                ),
+                                "{}",
+                                get_sql_diagnostics(HandleType::Stmt, stmt as Handle)
+                            );
+                            assert_eq!(
```
nbagnard, line 688: Same for this, we have tests for SQLTables which are checking the resultset, so I think you can skip this validation.

nbagnard (changes requested):
Great job!
A couple of small comments but it's almost ready to go already :)
If you can extract common part with this test and test_data_retrieval that will be awesome.

integration_test/tests/power_bi_tests.rs
bucaojit: That's a good suggestion, I added helper functions for these two sections.

integration_test/tests/power_bi_tests.rs
bucaojit: Added this 👍

integration_test/tests/power_bi_tests.rs
bucaojit: Sounds good, I added this check based on the last PR comments regarding checking the str_len_ptr length.  You are right in that this is already checked and would require this test to also be updated when the data set changes.  I have removed these validations.

integration_test/tests/power_bi_tests.rs
bucaojit: Sounds good.

integration_test/tests/power_bi_tests.rs
```diff
@@ -531,4 +531,185 @@ mod integration {
...
+                Desc::Scale,
+            ];
+            let expected_col_attr_string_lengths = [
+                &[0, 0, 18, 18, 12, 12, 12],
+                &[12, 12, 22, 22, 12, 12, 12],
+                &[12, 12, 20, 20, 12, 12, 12],
+                &[12, 12, 20, 20, 12, 12, 12],
+                &[12, 12, 14, 14, 12, 12, 12],
+            ];
+            for col_num in 0..*column_count_ptr {
```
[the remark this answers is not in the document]
bucaojit: That's a good idea.  I've added the assert for `*column_count_ptr` in the helper function `get_column_attributes()`.

integration_test/tests/power_bi_tests.rs
```diff
@@ -531,4 +531,185 @@ mod integration {
...
+                let mut field_num: usize = 0;
```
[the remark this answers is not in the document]
bucaojit: Thanks for that, I've removed the string length checking so the `field_num` is no longer needed.

integration_test/tests/power_bi_tests.rs
```diff
@@ -531,4 +531,185 @@ mod integration {
...
+                        }
+                        row_num += 1;
+                    }
+                    // break if SQLFetch returns SQL_NO_DATA
+                    _ => break,
+                }
+            }
+
+            assert_eq!(
+                7, successful_fetch_count,
```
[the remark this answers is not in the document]
bucaojit: Yes, I decided to keep `successful_fetch_count` instead of row_num as it was more descriptive.

integration_test/tests/power_bi_tests.rs
```diff
@@ -199,6 +199,136 @@ mod integration {
...
+                        successful_fetch_count += 1;
+                        for col_num in 0..target_types.len() {
+                            assert_eq!(
+                                expected_sql_returns[col_num],
+                                SQLGetData(
+                                    stmt as HStmt,
+                                    (col_num + 1) as USmallInt,
+                                    target_types[col_num],
+                                    output_buffer as Pointer,
+                                    BUFFER_LENGTH as Len,
```
nbagnard, line 302: [nit] Why not using the real buffer size `BUFFER_LENGTH * size_of::u16`?

nbagnard (approved):
LGTM 💯 
This is great! Thank you for the refactoring 😃

jchemburkar (approved):
LGTM

bucaojit:
evergreen merge

PR #88, opened 2023-02-11
SQL-1064: improve core error messages

jchemburkar:
For the `Mongo` and `BsonDeserialization` errors, I mostly tried to wrap the underlying error with context from where it is being called. I felt a little awkward adding a lot of strings throughout the codebase, if there is a more apparent (or clearer) way to do this, definitely open to it!

For the `ValueAccessError`, it seemed to just be mapped when accessing the broader document using table name failed (and not any specific field), so I just added the table name.


core/src/err.rs
```diff
@@ -9,8 +9,8 @@ pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Error, Debug, Clone)]
 pub enum Error {
-    #[error(transparent)]
-    BsonDeserialization(#[from] bson::de::Error),
+    #[error("Parsing metadata for {0} gave: {1}")]
```
nbagnard, line 12: [nit]
I guess we are only deserializing metadata type information, so it makes sense. But since the error message is user facing, I don't know if they need that level of details.
What about "Retrieving information" instead of "Parsing metadata"?

core/src/err.rs
```diff
@@ -23,22 +23,22 @@ pub enum Error {
...
-    Mongo(#[from] mongodb::error::Error),
+    #[error("Trying to {0} gave: {1}")]
+    Mongo(String, mongodb::error::Error),
     #[error("No database provided for query")]
     NoDatabase,
     #[error("Unknown column '{0}' in result set schema")]
     UnknownColumn(String),
-    #[error(transparent)]
-    ValueAccess(bson::document::ValueAccessError),
+    #[error("Trying to access collection '{0}' gave: {1}")]
```
nbagnard, line 32: Umh, the error happens when getting the value for a specific field.
The collection name might not make any sense to the end user here. It is easily empty, for example with `select 1 from mytable` or if there is an alias, this is the collection name we will see.
What about "Error retrieving data for field {0} : {1}", with 0 being the column index?

nbagnard (changes requested):
This is looking good! I have a couple of small suggestion which could help not confusing users even more, just give enough context.

core/src/conn.rs
```diff
@@ -57,8 +58,11 @@ impl MongoConnection {
     /// Gets the ADF version the client is connected to.
     pub fn get_adf_version(&self) -> Result<String> {
         let db = self.client.database("admin");
-        let cmd_res = db.run_command(doc! {"buildInfo": 1}, None)?;
-        let build_info: BuildInfoResult = bson::from_document(cmd_res)?;
+        let cmd_res = db
+            .run_command(doc! {"buildInfo": 1}, None)
+            .map_err(|e| Error::Mongo("get ADF version".to_string(), e))?;
```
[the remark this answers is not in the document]
jchemburkar: Makes sense, and good to know!

core/src/err.rs
```diff
@@ -9,8 +9,8 @@ pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Error, Debug, Clone)]
 pub enum Error {
-    #[error(transparent)]
-    BsonDeserialization(#[from] bson::de::Error),
+    #[error("Retrieving information for {0} failed with error: {1}")]
+    BsonDeserialization(String, bson::de::Error),
```
mattChiaravalloti, line 13: In general, instead of parametrizing these 3 error types with `String`s to differentiate them, we should create as many error variants as we need. This would avoid the problem you pointed out in your PR summary: including `String`s in all the places these errors are used.

The reason these were grouped together initially is _because_ they were `transparent`. Now that we are not making them `transparent`, we could and should separate them. Let's make an error variant for `DatabaseVersionRetrieval` (or just `DatabaseVersion`) and have the format string directly read: `#[error("Database version could not be retrieved: {0}")]`; we can still parametrize with the underlying error, `mongodb::error::Error`.

mattChiaravalloti (changes requested):
Just one comment, but it is critical to the rest of what I would review. Let's discuss my current comment before I proceed with more feedback!

core/src/err.rs
jchemburkar: This sounds good to me too, I wasn't 100% sure if this was the right path to go down initially, but definitely makes sense. The `DatabaseVersion` error makes sense, and just to double check, is that an example (worth following for all strings throughout the changes I made), or are you specifically just thinking about changing this one error

jchemburkar: ^ I'm definitely ok with doing all, and think it will come out reading cleanly -- my only initial concern was with cluttering up the list of errors in `err.rs`, but I don't think it will end up being that many needed to be added

mattChiaravalloti: Good question -- I meant it as an example worth following for all strings. Basically, every case where we have a distinct error string in this PR, it should be a distinct `Error` enum variant.

The way I'm thinking of it is: when we first created these `Error` variants, we did so liberally. Why should these be the arbitrary cutoff simply because we did not distinguish them in our first version of this code? (My answer to that question is that there _is no reason_ to make that cutoff.) In fact, `MongoParseConnectionString` already demonstrates this idea in action.

jchemburkar: Got it, thank you for the context! Given that, I am happy to go down that path, I think it will end up cleaner, and definitely satisfy my fear as you noted

nbagnard (approved):
LGTM!

mattChiaravalloti (approved):

jchemburkar:
evergreen merge

PR #91, opened 2023-02-16
SQL-1149: provide feedback when SQLColumns schema deserialization fails

jchemburkar:
The largest portion of this PR is updating the result set to return Results as `Result<(bool, Option<Error>)>` for the sake of consuming errors the ODBC functions. I only focussed on one facet, `SQLColumns` deserialization, though other areas of future work are documented in [SQL-1198](https://jira.mongodb.org/browse/SQL-1198), which should be updated when this PR is merged. 

In the `MongoFields` functions I only kept track of the first error that was hit. This is an open question of what ideal behavior is (should we store a vector, eg?). My working assumption was that tools, upon getting `SUCCESS_WITH_INFO`, may just try to get the most recent diagnostic (and thus, keeping track of a list would not be useful). I wanted to flag this in case that is untrue, or we should try to keep track of all errors.

### DIFF
--- a/core/src/collections.rs
+++ b/core/src/collections.rs
@@ -138,10 +138,10 @@ impl MongoStatement for MongoCollections {
     // Move the cursor to the next CollectionSpecification.
     // When cursor is exhausted move to next database in list
     // Return true if moving was successful, false otherwise.
-    fn next(&mut self, _: Option<&MongoConnection>) -> Result<bool> {
+    fn next(&mut self, _: Option<&MongoConnection>) -> Result<(bool, Option<Error>)> {
         if self.current_database_index.is_none() {
             if self.collections_for_db_list.is_empty() {
-                return Ok(false);
+                return Ok((false, None));
             }
             self.current_database_index = Some(0);
         }
@@ -177,13 +177,13 @@ impl MongoStatement for MongoCollections {
                 {
                     // Cursor advance succeeded and the collection matches the filters, update current CollectionSpecification
                     self.current_collection = Some(collection);
-                    return Ok(true);
+                    return Ok((true, None));
                 }
             }
 
             self.current_database_index = Some(self.current_database_index.unwrap() + 1);
             if self.current_database_index.unwrap() >= self.collections_for_db_list.len() {
-                return Ok(false);
+                return Ok((false, None));
             }
         }
     }
--- a/core/src/fields.rs
+++ b/core/src/fields.rs
@@ -569,14 +569,18 @@ impl MongoFields {
 impl MongoStatement for MongoFields {
     // Move the cursor to the next document and update the current row.
     // Return true if moving was successful, false otherwise.
-    fn next(&mut self, mongo_connection: Option<&MongoConnection>) -> Result<bool> {
+    fn next(
+        &mut self,
+        mongo_connection: Option<&MongoConnection>,
+    ) -> Result<(bool, Option<Error>)> {
         match self.field_name_filter.as_ref() {
             None => {
                 self.current_field_for_collection += 1;
-                Ok(
+                Ok((
                     (self.current_field_for_collection as usize) < self.current_col_metadata.len()
                         || self.get_next_metadata(mongo_connection.unwrap()),
-                )
+                    None,
+                ))
             }
             Some(filter) => {
                 let filter = filter.clone();
@@ -586,7 +590,7 @@ impl MongoStatement for MongoFields {
                         >= self.current_col_metadata.len())
                         && !self.get_next_metadata(mongo_connection.unwrap())
                     {
-                        return Ok(false);
+                        return Ok((false, None));
                     }
 
                     if filter.is_match(
@@ -596,7 +600,7 @@ impl MongoStatement for MongoFields {
                             .unwrap()
                             .col_name,
                     ) {
-                        return Ok(true);
+                        return Ok((true, None));
                     }
                 }
             }
--- a/core/src/fields.rs
+++ b/core/src/fields.rs
@@ -532,8 +532,10 @@ impl MongoFields {
                         bson::from_document(db.run_command(get_schema_cmd, None).unwrap())
                             .map_err(Error::BsonDeserialization);
                     if let Err(error) = current_col_metadata_response {
-                        // If there is an Error while deserialization the schema, we don't show the column
-                        e = Some(error);
+                        // If there is an Error while deserializing the schema, we don't show the column
+                        if e.is_none() {
+                            e = Some(error);
+                        }
                         continue;
                     }
                     let current_col_metadata_response = current_col_metadata_response.unwrap();
--- a/core/src/fields.rs
+++ b/core/src/fields.rs
@@ -506,7 +506,11 @@ impl MongoFields {
         }
     }
 
-    fn get_next_metadata(&mut self, mongo_connection: &MongoConnection) -> bool {
+    fn get_next_metadata(
+        &mut self,
+        mongo_connection: &MongoConnection,
+    ) -> Result<(bool, Option<Error>)> {
+        let mut e: Option<Error> = None;
         loop {
             if self.collections_for_db.is_some() {
                 for current_collection in self.collections_for_db.as_mut().unwrap() {
@@ -527,9 +531,9 @@ impl MongoFields {
                     let current_col_metadata_response: Result<SqlGetSchemaResponse> =
                         bson::from_document(db.run_command(get_schema_cmd, None).unwrap())
                             .map_err(Error::BsonDeserialization);
-                    if current_col_metadata_response.is_err() {
+                    if let Err(error) = current_col_metadata_response {
                         // If there is an Error while deserialization the schema, we don't show the column
-                        // TODO : Add a log or warning
+                        e = Some(error);
                         continue;
                     }
                     let current_col_metadata_response = current_col_metadata_response.unwrap();
@@ -541,7 +545,7 @@ impl MongoFields {
                             if !current_col_metadata.is_empty() {
                                 self.current_col_metadata = current_col_metadata;
                                 self.current_field_for_collection = 0;
-                                return true;
+                                return Ok((true, e));
                             }
                         }
                         // If there is an error simplifying the schema (e.g. an AnyOf), skip the collection
@@ -551,7 +555,7 @@ impl MongoFields {
                 }
             }
             if self.dbs.is_empty() {
-                return false;
+                return Ok((false, e));
             }
             let db_name = self.dbs.pop_front().unwrap();
             self.collections_for_db = Some(
@@ -576,21 +580,22 @@ impl MongoStatement for MongoFields {
         match self.field_name_filter.as_ref() {
             None => {
                 self.current_field_for_collection += 1;
-                Ok((
-                    (self.current_field_for_collection as usize) < self.current_col_metadata.len()
-                        || self.get_next_metadata(mongo_connection.unwrap()),
-                    None,
-                ))
+                match (self.current_field_for_collection as usize) < self.current_col_metadata.len()
+                {
+                    true => Ok((true, None)),
+                    false => self.get_next_metadata(mongo_connection.unwrap()),
+                }
             }
             Some(filter) => {
                 let filter = filter.clone();
                 loop {
                     self.current_field_for_collection += 1;
-                    if (self.current_field_for_collection as usize
-                        >= self.current_col_metadata.len())
-                        && !self.get_next_metadata(mongo_connection.unwrap())
+                    if self.current_field_for_collection as usize >= self.current_col_metadata.len()
                     {
-                        return Ok((false, None));
+                        let next = self.get_next_metadata(mongo_connection.unwrap());
+                        if next.is_ok() & !next.unwrap().0 {
+                            return Ok((false, None));
+                        }
                     }
 
                     if filter.is_match(
--- a/core/src/fields.rs
+++ b/core/src/fields.rs
@@ -590,14 +590,20 @@ impl MongoStatement for MongoFields {
                 let filter = filter.clone();
                 loop {
                     self.current_field_for_collection += 1;
+                    let mut e: Option<Error> = None;
                     if self.current_field_for_collection as usize >= self.current_col_metadata.len()
                     {
                         let next = self.get_next_metadata(mongo_connection.unwrap());
-                        if next.is_ok() & !next.unwrap().0 {
-                            return Ok((false, None));
+                        if let Ok((b, error)) = next {
+                            // store the first error we see from get_next_metadata, to return at conclusion of function
+                            if e.is_none() && error.is_some() {
+                                e = error;
+                            }
+                            if !b {
+                                return Ok((false, e));
+                            }
                         }
                     }
-
                     if filter.is_match(
                         &self
                             .current_col_metadata
@@ -605,7 +611,7 @@ impl MongoStatement for MongoFields {
                             .unwrap()
                             .col_name,
                     ) {
-                        return Ok((true, None));
+                        return Ok((true, e));
                     }
                 }
             }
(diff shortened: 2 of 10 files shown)


integration_test/src/test_runner/mod.rs
```diff
@@ -743,7 +743,7 @@ fn get_column_count(stmt: &Statement<Allocated, NoResult, AutocommitOn>) -> Resu
 fn fetch_row(stmt: &Statement<Allocated, NoResult, AutocommitOn>) -> Result<bool> {
     unsafe {
         match odbc_sys::SQLFetch(stmt.handle() as HStmt) {
-            SqlReturn::SUCCESS => Ok(true),
+            SqlReturn::SUCCESS | SqlReturn::SUCCESS_WITH_INFO => Ok(true),
```
jchemburkar, line 746: Note: added this to account for updates to `SQLFetch`. the `baz` view in the `integration_test` database triggers the newly added `SUCCESS_WITH_INFO` where applicable.

core/src/fields.rs
```diff
@@ -569,34 +575,45 @@ impl MongoFields {
...
-                        && !self.get_next_metadata(mongo_connection.unwrap())
+                    let mut e: Option<Error> = None;
+                    if self.current_field_for_collection as usize >= self.current_col_metadata.len()
                     {
-                        return Ok(false);
+                        let next = self.get_next_metadata(mongo_connection.unwrap());
+                        if let Ok((b, error)) = next {
+                            // store the first error we see from get_next_metadata, to return at conclusion of function
+                            if e.is_none() && error.is_some() {
+                                e = error;
```
nbagnard, line 602: If you change get_next_metadata to return a vector, this will have to change to a vector too.
And you will also have to add a vector of errors to the existing one if this is not the first error.
Also, I think there is a missing "continue" because if `get_next_metadata` comes back with an error, then trying to do filter will likely likely fail because `self.current_field_for_collection as usize >= self.current_col_metadata.len()`, no?

integration_test/src/test_runner/mod.rs
nbagnard: umh, why is baz generating a SUCCESS_WITH_INFO right now? Is there a deserialization error or is it something else?

odbc/src/api/functions.rs
```diff
@@ -1365,20 +1365,26 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
                 Err(e) => {
                     stmt.errors.write().unwrap().push(e.into());
-                    return SqlReturn::ERROR;
+                    SqlReturn::ERROR
                 }
-                Ok(b) => {
+                Ok((b, e)) => {
                     let mut stmt_attrs = stmt.attributes.write().unwrap();
+                    if let Some(error) = e.clone() {
+                        stmt.errors.write().unwrap().push(ODBCError::Core(error));
```
nbagnard, line 1373: I don't think that this will work.
This will report a General Error in most cases, whereas we want to return a General Warning.
You need to add the SQL state GENERAL_WARNING (01000) and report this instead.

nbagnard (changes requested):
Good start! The main thing is that it looks like we are only reporting the first skipped error though, we should be able to report all skipped collections/fields.
Also, we should add a General Warning, not a General Error. General Error is associated with SQL_ERROR while the other one is associated with SQL_SUCCESS_WITH_INFO (although we can do whatever we want of course since we just add diagnostic records for it)

core/src/fields.rs
```diff
@@ -527,9 +531,11 @@ impl MongoFields {
                     let current_col_metadata_response: Result<SqlGetSchemaResponse> =
                         bson::from_document(db.run_command(get_schema_cmd, None).unwrap())
                             .map_err(Error::BsonDeserialization);
-                    if current_col_metadata_response.is_err() {
-                        // If there is an Error while deserialization the schema, we don't show the column
-                        // TODO : Add a log or warning
+                    if let Err(error) = current_col_metadata_response {
+                        // If there is an Error while deserializing the schema, we don't show the column
+                        if e.is_none() {
```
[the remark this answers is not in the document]
jchemburkar: 👍 sgtm, I left that as an open question in the PR comment, but wasn't 100% sure. I think turning this into a vector makes sense!

core/src/collections.rs
```diff
@@ -138,10 +138,10 @@ impl MongoStatement for MongoCollections {
...
     // Return true if moving was successful, false otherwise.
-    fn next(&mut self, _: Option<&MongoConnection>) -> Result<bool> {
+    fn next(&mut self, _: Option<&MongoConnection>) -> Result<(bool, Option<Error>)> {
         if self.current_database_index.is_none() {
             if self.collections_for_db_list.is_empty() {
-                return Ok(false);
+                return Ok((false, None));
             }
             self.current_database_index = Some(0);
         }
```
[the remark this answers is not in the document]
jchemburkar: Agreed. I wasn't sure if it was in scope for this ticket (given [sql-1198](https://jira.mongodb.org/browse/SQL-1198)), but can add that to this pr

core/src/fields.rs
```diff
@@ -541,7 +547,7 @@ impl MongoFields {
                             if !current_col_metadata.is_empty() {
                                 self.current_col_metadata = current_col_metadata;
                                 self.current_field_for_collection = 0;
-                                return true;
+                                return Ok((true, e));
```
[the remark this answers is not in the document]
jchemburkar: by resurfacing the error below, do you mean adding the error in the other branch of this match to the (now vector) of errors?

core/src/fields.rs
jchemburkar: Thanks for the heads about changes if errors becomes a vector. I could be mistaken, but I think even if we get errors in `get_next_metadata`, it will still continue until it successfully parses a collection and resets the `current_field_for_collection` (or runs out of available collections, and returns `Ok(false, ...)`, which might eliminate the need for a continue

integration_test/src/test_runner/mod.rs
jchemburkar: I think it is a deserialization error that was just getting skipped over before, but I see ```Err(OdbcFunctionFailed("SQLFetch", "SUCCESS_WITH_INFO", "[MongoDB][Core] missing field `version`"))``` when I run this on my windows host with these changes.

core/src/collections.rs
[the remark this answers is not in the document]
nbagnard: yes, I am not too sure either. I think at least the errors on advance and deserialize_current make sense to skip and report as warning similarly to what is done in fields.rs.
@pmeredit How did you decide which error to skip and which to propagate for fields.rs? For example, why stopping if there is an error on .list_collections(None, None).unwrap()?

core/src/fields.rs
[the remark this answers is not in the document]
nbagnard: oh, sorry, just reading this now :( Yes, a vector probably makes the most sense. Technically a function call can generate multiple warnings. Not sure what tools do with it, I'm sure that we will find out soon enough.

core/src/fields.rs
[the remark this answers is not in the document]
nbagnard: yes, I mean adding it to the vector of errors. I think otherwise right now it is simply dropped, no? Maybe I missed something though and it is handle somewhere and I missed it. If that's the case, ignore this ;)

core/src/fields.rs
nbagnard: ohhh, you are right, for some reason I thought that the filter was inside the loop. Sorry about that.

core/src/collections.rs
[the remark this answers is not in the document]
pmeredit: I don't expect `list_collections` to actually ever fail.

core/src/fields.rs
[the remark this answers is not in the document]
pmeredit: Our handles allow for a vector of errors, so we can just extend that with the result of this.

core/src/collections.rs
```diff
@@ -138,53 +138,74 @@ impl MongoStatement for MongoCollections {
...
-                        .as_ref()
-                        .unwrap()
-                        .contains(&collection.collection_type))
-                    && (self.collection_name_filter.is_none()
-                        || self
-                            .collection_name_filter
-                            .as_ref()
-                            .unwrap()
-                            .is_match(&collection.name))
+                    .advance()
```
nbagnard, line 156: Could you use https://doc.rust-lang.org/std/result/enum.Result.html#method.or_else to add errors to the vector instead having to use match blocks?

core/src/fields.rs
```diff
@@ -570,34 +574,45 @@ impl MongoFields {
...
+                }
             }
             Some(filter) => {
                 let filter = filter.clone();
                 loop {
                     self.current_field_for_collection += 1;
-                    if (self.current_field_for_collection as usize
-                        >= self.current_col_metadata.len())
-                        && !self.get_next_metadata(mongo_connection.unwrap())
+                    let mut e: Vec<Error> = vec![];
```
nbagnard, line 594: It seems odd to create the vector inside the loop. Should it be outside of it and new warning are added when happening on each iteration?

core/src/fields.rs
```diff
@@ -570,34 +574,45 @@ impl MongoFields {
...
+                    if self.current_field_for_collection as usize >= self.current_col_metadata.len()
                     {
-                        return Ok(false);
+                        let next = self.get_next_metadata(mongo_connection.unwrap());
+                        if let Ok((b, errors)) = next {
```
nbagnard, line 598: Can you rename the variables and use something like `Ok((b, warnings))` or maybe even `Ok((has_more_coll, warnings))`?

core/src/fields.rs
```diff
@@ -570,34 +574,45 @@ impl MongoFields {
...
                 loop {
                     self.current_field_for_collection += 1;
-                    if (self.current_field_for_collection as usize
-                        >= self.current_col_metadata.len())
-                        && !self.get_next_metadata(mongo_connection.unwrap())
+                    let mut e: Vec<Error> = vec![];
+                    if self.current_field_for_collection as usize >= self.current_col_metadata.len()
                     {
-                        return Ok(false);
+                        let next = self.get_next_metadata(mongo_connection.unwrap());
```
nbagnard, line 597: [Simple comment] If you did not want to break the condition check, you could use and_then and do something like this
```
if (self.current_field_for_collection as usize
                        >= self.current_col_metadata.len())
                        && !self
                        .get_next_metadata(mongo_connection.unwrap())
                        .and_then(|res| {
                            if let Some(warns) = res.1 {
                                warnings.extend(warns)
                            };
                            Ok(res.0)
                        })
                        .unwrap()
                    {
                        return Ok((false, (!warnings.is_empty()).then_some(warnings)));
                    }
```
I am not sure if it is more readable or less... I find both are very similar. This is maybe just a more "Rust way".

nbagnard (changes requested):
Nice work! It takes some work to drag those warnings around!
I am wondering if it would be possible to create a test checking that we add warnings as we go. I did not see one. It's already good to see that with luck, you have been able to test that warnings are generating SUCCESS_WITH_INFO, but a more complete test case would be awesome.
Another thing I was also wondering is if we could not get rid of the Option<&MongoConnection> in next and only pass in the &Connection, failing in SQLFetch if there is no MongoConnection (that'd be a big problem).

core/src/fields.rs
jchemburkar: yes, good catch, I agree this should be outside the loop

odbc/src/api/functions.rs
```diff
@@ -1370,20 +1370,31 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
             match res {
                 Err(e) => {
                     stmt.errors.write().unwrap().push(e.into());
-                    return SqlReturn::ERROR;
+                    SqlReturn::ERROR
                 }
-                Ok(b) => {
+                Ok((b, e)) => {
                     let mut stmt_attrs = stmt.attributes.write().unwrap();
+                    if let Some(errors) = e.clone() {
```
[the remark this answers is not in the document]
jchemburkar: I was hoping to do this when I initially tried, but calling the macro was complaining about borrowing rules, since the `mongo_handle` has already been used immutably (to get the stmt I believe). I'm not sure if this was intentional, but noticed this in a couple other places in the `SQLFetch` (e.g. [here](https://github.com/mongodb/mongo-odbc-driver/blob/11461a78c549d3afa1e808e0bc11c63b190b1d99/odbc/src/api/functions.rs#L1372)).

One way to work around this would be to make _new_ references to the handle, such as ```add_diag_with_function!(MongoHandleRef::from(statement_handle), warning, "SQLFetch");```, but borrowing the same handle multiple times feels unsafe to me. I'm still getting used to the concept of borrowing however, so definitely possible I'm missing something.

core/src/fields.rs
jchemburkar: 👍 makes sense, and happy to do it the more "rust way"! thanks for the suggestion

jchemburkar:
@nbagnard I made the relevant code changes, and to address the other points of discussion:
- I looked into changing the Option<&MongoConnection> to &MongoConnection. The biggest blocker I see would be the MockQuery, which has a non-existent connection by definition. Maybe if we want to prevent failure in SQLFetch we could just directly check the connection is not none/raise a failure?
- In trying to test this, I found it difficult for reasons we discussed last week. To note on PowerBI, I believe this should not fail/cause an error. I manually tested and cross-referenced the log, and noticed that when SQLColumns is called, it is only for a single table. Thus, even if these errors are generated, there will be no columns returned from SQLColumns, and thus the SQLFetch will yield SQL_NO_DATA. PowerBI then just reports there are no valid columns to be displayed for a given collection. So I think this is at least safe for PowerBI, but the warnings aren't reported/there is nothing prompting a call to SQLGetDiagRec.

pmeredit (approved):
LGTM!

My only possible nit here is we could just return `Vec<Error>` instead of `Option<Vec<Error>>` because the empty vec is isomorphic to `None`, but I'm fine with `Option` if you think that's easier.

odbc/src/api/functions.rs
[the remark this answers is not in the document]
nbagnard: Here is a snippet for SQLFetch which adds the Warnings to the diagnostic records for the handle and log them.
It first populates the diag. records, even though if the result is SQL_NO_DATA I bet now now will call SQLGetDiagRec, but warnings will be logged in every case this way.
If you remove the Option on the Warnings, just check the length of the vector.
```
pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
    panic_safe_exec!(
        || {
            let stmt = must_be_valid!(mongo_handle.as_statement());
            let move_to_next_result = {
                let connection = must_be_valid!((*stmt.connection).as_connection());
                match stmt.mongo_statement.write().unwrap().as_mut() {
                    Some(mongo_stmt) => mongo_stmt
                        .next(connection.mongo_connection.read().unwrap().as_ref())
                        .map_err(|e| e.into()),
                    None => Err(ODBCError::InvalidCursorState),
                }
            };

            if let Ok((has_next, warnings_opt)) = move_to_next_result {
                let mut stmt_attrs = stmt.attributes.write().unwrap();

                // Add the warning to the diagnostic records and log them
                let has_warnings = warnings_opt.is_some();
                if let Some(warnings) = warnings_opt {
                    warnings.iter().for_each(|warning| {
                        add_diag_info!(MongoHandleRef::from(statement_handle), warning.into());
                    });
                }
                if !has_next {
                    stmt_attrs.row_index_is_valid = false;
                    // No more rows
                    return SqlReturn::NO_DATA;
                }
                stmt_attrs.row_index_is_valid = true;

                *stmt.var_data_cache.write().unwrap() = Some(HashMap::new());

                if has_warnings {
                    // No warnings and there is a next row
                    SqlReturn::SUCCESS_WITH_INFO
                } else {
                    SqlReturn::SUCCESS
                }
            } else {
                add_diag_info!(
                    mongo_handle,
                    move_to_next_result.as_ref().unwrap_err().clone()
                );
                // An error happened
                SqlReturn::ERROR
            }
        },
        statement_handle
    );
}

```

nbagnard (approved):
LGTM! 🎉 🎉 🎉 🎉 🎉

jchemburkar:
evergreen merge

PR #96, opened 2023-02-23
SQL-1264: parse appName for telemetry

terakilobyte:
This adds parsing of the appName which will be passed in a [sibling PR](https://github.com/mongodb/mongo-powerbi-connector/pull/16).


core/src/odbc_uri.rs
```diff
@@ -266,6 +273,7 @@ impl<'a> ODBCUri<'a> {
                 ServerAddress::parse(server).map_err(Error::InvalidClientOptions)?
             ])
             .credential(cred)
+            .app_name(self.get_app_name())
             .build())
     }
 }
```
bucaojit, line 279: It would be good to have a unit test below to validate that the `APPNAME` behavior is as expected.

core/src/odbc_uri.rs
```diff
@@ -250,9 +251,15 @@ impl<'a> ODBCUri<'a> {
             );
         }
         Self::set_server_and_source(&mut client_options, server, source)?;
+        client_options.app_name = self.get_app_name();
```
bucaojit, line 254: The app_name can be set in the URI as well ([appName option](https://www.mongodb.com/docs/manual/reference/connection-string/#mongodb-urioption-urioption.appName)).  I think the value in the URI should take precedence if set, since in the PowerBI connector that is the only way the user can modify the appName.  
We can add a check here whether it is already set in client_options, otherwise we set it to `self.get_app_name()`.

bucaojit (changes requested):

core/src/odbc_uri.rs
terakilobyte: Good catch! I had one but for some reason didn't include it in the commit. I've updated it to ensure the mongodb uri overrides the odbc uri, if both are provided.

core/src/odbc_uri.rs
```diff
@@ -266,8 +272,21 @@ impl<'a> ODBCUri<'a> {
...
+    fn get_app_name(&mut self, uri: Option<&str>) -> Option<String> {
+        if let Some(uri) = uri {
+            APP_NAME_REGEX
+                .captures(uri)
+                .and_then(|cap| cap.name("appname").map(|s| s.as_str()))
+                .map_or(self.remove(&[APPNAME]), |s| Some(s))
+        } else {
+            self.remove(&[APPNAME])
+        }
+        .map(String::from)
```
terakilobyte, line 288: Should we default this to "odbc-driver" so that there is some telemetry for the raw odbc driver?

bucaojit: Yeah, it used to be set to `mongo_odbc_driver` using [this variable](https://github.com/mongodb/mongo-odbc-driver/blob/7bcd73ce65d0dde61267d636739d420a96df2db9/core/src/conn.rs#L26).  Maybe we can use that name, or switch the underscores to dashes like with the connectors.

core/src/odbc_uri.rs
```diff
@@ -37,6 +39,10 @@ lazy_static! {
         .case_insensitive(true)
         .build()
         .unwrap();
+    static ref APP_NAME_REGEX: Regex = RegexBuilder::new(r#"[&?]appName=(?P<appname>[^&]*)"#)
```
bucaojit, line 42: Looks like `APP_NAME_REGEX` is no longer used.  Let's remove it.

core/src/odbc_uri.rs
terakilobyte: This seems really out of place here since we have constants centrally located elsewhere. I'm asking Alexi what she wants to name to be and will consolidate that to `constants/src/lib.rs` and delete this variable.

terakilobyte: I got feedback from Alexi on the name and removed this variable since it wasn't being used anywhere.

constants/src/lib.rs
```diff
@@ -2,6 +2,7 @@ pub const VENDOR_IDENTIFIER: &str = "MongoDB";
 pub const DRIVER_NAME: &str = "MongoDB Atlas SQL interface ODBC Driver";
 pub const DBMS_NAME: &str = "MongoDB Atlas";
 pub const ODBC_VERSION: &str = "03.80";
+pub const DEFAULT_APP_NAME: &str = "odbc-driver";
```
nbagnard, line 5: The default app name I see for the JDBC driver is `mongodb-jdbc+{version}`.
What do you think about keeping things similar and use`mongodb-odbc+{version}`?
For example `mongodb-jdbc+2.0.2`.
I think adding the version is worth it for future data collection. It can be useful sometime to know if users are upgrading their driver or stay with older versions and/or see if there are users using older version when a new version with breaking changes is released.

terakilobyte: Good point!

pmeredit (approved):
LGTM!

constants/src/lib.rs
terakilobyte: I've added version information for appName, as long as the user hasn't specified their own appName @nbagnard . Thanks for the suggestion!

constants/src/lib.rs
```diff
@@ -1,7 +1,21 @@
+use lazy_static::lazy_static;
+
 pub const VENDOR_IDENTIFIER: &str = "MongoDB";
 pub const DRIVER_NAME: &str = "MongoDB Atlas SQL interface ODBC Driver";
 pub const DBMS_NAME: &str = "MongoDB Atlas";
 pub const ODBC_VERSION: &str = "03.80";
+pub const DRIVER_SHORT_NAME: &str = "mongodb-odbc";
+
+lazy_static! {
```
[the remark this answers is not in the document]
terakilobyte: I was going to do this initially, but thought they were for two different uses. The driver version I added is for metrics reporting, and the one in `odbc/src/api/utils.rs` is to format the version for return to the client.

Happy to update it so we just have one version format, and can consolidate it to constants if that's what we want to do! LMK 😄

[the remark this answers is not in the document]
terakilobyte: I gave this more thought and I fully understand what you mean. I've made some changes that address both points.

constants/src/lib.rs
```diff
@@ -31,3 +46,136 @@ pub const INVALID_CHARACTER_VALUE: &str = "22018";
...
+        format_cargo_version,
+        expected = "00.01.0000",
+        major = "0",
+        minor = "1",
+        patch = "0"
+    );
+}
+
+#[cfg(test)]
+mod version {
```
terakilobyte, line 125: @nbagnard I thought more about the differing versions potential problem. This test ensures that all versions remain in sync.

nbagnard (approved):
LGTM!

terakilobyte:
evergreen merge

terakilobyte:
evergreen merge

PR #97, opened 2023-02-25
SQL-1292: return binary subtype 4 as uuid

terakilobyte:
This returns binary subtype 4 data in uuid form. In Power Bi, the user will still see it as Binary, but they can then transform it to text, then to JSON. The choice of "extended JSON" here is best intent to make it clear to the user what it is.

### DIFF
--- a/odbc/src/api/data_tests.rs
+++ b/odbc/src/api/data_tests.rs
@@ -67,6 +67,10 @@ const REGEX_STR_VAL: (u16, &str) = (
 );
 const STRING_STR_VAL: (u16, &str) = (STRING_COL, "hello world!");
 const UNIT_STR_STR_VAL: (u16, &str) = (UNIT_STR_COL, "a");
+const GUID_STR_VAL: (u16, &str) = (
+    GUID_COL,
+    "{\"$uuid\":\"d5e3ac31-fd29-49e3-8759-08d116e98b29\"}",
+);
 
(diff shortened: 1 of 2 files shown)


odbc/src/api/data_tests.rs
```diff
@@ -107,7 +111,7 @@ lazy_static! {
                 "unit_str": "a",
                 "guid": Bson::Binary(Binary {
                     subtype: BinarySubtype::Uuid,
-                    bytes: vec![0u8, 1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 9u8, 10u8, 11u8, 12u8, 13u8, 14u8, 15u8],
+                    bytes: vec![0xd5, 0xe3, 0xac, 0x31, 0xfd, 0x29, 0x49, 0xe3, 0x87, 0x59, 0x08, 0xd1, 0x16, 0xe9, 0x8b, 0x29],
```
terakilobyte, line 114: I replaced this with a valid uuid generated from `mongosh`, but was curious if `8u8` was omitted from the original on purpose (which would make it invalid since it needs 16 bytes, and was only 15).

pmeredit (approved):
LGTM assuming we are ok ignoring subtype 3 for now

nbagnard (approved):
LGTM!

terakilobyte:
evergreen merge

PR #100, opened 2023-03-02
SQL-1279: add setup gui for dsn

terakilobyte:
I think this is nearly there. I've a few outstanding questions about what we should do with a few things that I want to wrap up before marking it as ready for merge.


win_setupgui/src/gui/config.rs
```diff
@@ -0,0 +1,269 @@
...
+    #[nwg_layout_item(layout: grid,  row: 5, col: 0, col_span: 2)]
+    database_field: nwg::Label,
+
+    #[nwg_control(flags: "VISIBLE", text: "")]
+    #[nwg_layout_item(layout: grid,  row: 5, col: 2, col_span: 7)]
+    database_input: nwg::TextInput,
+
+    #[nwg_control(flags: "VISIBLE", text: "Log Path")]
+    #[nwg_layout_item(layout: grid,  row: 6, col: 0, col_span: 2)]
+    log_path_field: nwg::Label,
```
terakilobyte, line 68: I still need to make this a file dialogue with directory-picking behavior.

core/src/odbc_uri.rs
```diff
@@ -9,16 +10,16 @@ const EMPTY_URI_ERROR: &str = "URI must not be empty";
...
+pub const DATABASE: &str = "database";
+pub const DRIVER: &str = "driver";
+pub const DSN: &str = "dsn";
+pub const PASSWORD: &str = "password";
+pub const PWD: &str = "pwd";
+pub const SERVER: &str = "server";
+pub const USER: &str = "user";
+pub const UID: &str = "uid";
+pub const URI: &str = "uri";
+pub const APPNAME: &str = "appname";
```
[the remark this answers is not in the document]
pmeredit: Might as well!

core/src/odbc_uri.rs
```diff
@@ -44,7 +44,7 @@ lazy_static! {
 }
 
 #[derive(Debug, PartialEq, Eq)]
-pub struct ODBCUri<'a>(HashMap<String, &'a str>);
+pub struct ODBCUri<'a>(pub HashMap<String, &'a str>);
```
pmeredit, line 47: Let's make this `pub(crate)`. I don't want to leak the abstraction outside of the core crate.

odbc/src/api/mod.rs
```diff
@@ -7,7 +7,8 @@ pub use functions::*;
 mod col_attr_describe_tests;
 #[cfg(test)]
 mod connect_attr_tests;
-mod data;
+pub(crate) mod data;
+pub use data::{input_text_to_string, input_wtext_to_string};
```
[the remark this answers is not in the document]
pmeredit: Yea, we can call it whatever

core/src/odbc_uri.rs
pmeredit: Actually, it looks like all you are using this for is to iterate over the keys. I'd just add an iter() method for ODBCUri that calls 0.iter and not make this `pub(crate)` either.

terakilobyte:
@pmeredit @nbagnard With the proposed changes here, namely the inclusion of the `logpath`, I'd like to create a ticket to use that (if available) by the `file_dbg_macros`. Is it okay to create that ticket greedily while this is in review?

core/src/util/registry.rs
```diff
@@ -0,0 +1,114 @@
+#[cfg(target_os = "windows")]
```
pmeredit, line 1: This cfg here is redundant with the one in mod.rs. I would say just leave the one in mod.rs (and presumably we'll have a module called ini for the other platforms).

core/src/util/registry.rs
```diff
@@ -0,0 +1,114 @@
...
+pub mod windows {
```
pmeredit, line 2: I don't think we need this module here, unless we want to change this to have non windows as a module in this file. In that case we'd want to take the cfg out of the mod file above. I'm fine with going either direction, but let's choose one.

core/src/util/registry.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    pub fn add_to_datasources(dsn: &str, driver: &str) -> io::Result<()> {
+        let hklm = RegKey::predef(HKEY_LOCAL_MACHINE);
+        let (settings, _disp) = hklm.create_subkey(ODBC_DATA_SOURCES)?;
+        match settings.set_value(dsn, &driver) {
+            Ok(_) => Ok(()),
+            Err(e) => {
+                dbg_write!(format!("Error while adding to datasources: {:?}", e));
```
pmeredit, line 73: Since we are actually returning errors, I think it makes more sense to format the errors with this message and only call `dbg_write` at the top where the error last manifests. Reduces redundant code. Do this in any function that returns `Result`. Obviously, the ones that don't return `Result` should call dbg_write :)

cstr/src/lib.rs
```diff
@@ -0,0 +1,137 @@
+use std::ptr::copy_nonoverlapping;
```
pmeredit, line 1: This is a great change.

cstr/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+
+///
+/// input_text_to_string converts an input cstring to a rust String.
+/// It assumes nul termination if the supplied length is negative.
+///
+/// # Safety
+/// This converts raw C-pointers to rust Strings, which requires unsafe operations
+///
+#[allow(clippy::uninit_vec)]
+pub unsafe fn input_text_to_string(text: *const Char, len: usize) -> String {
```
pmeredit, line 50: For consistency, let's called this `input_atext_to_string`

pmeredit: Or we can move `a` and `w` as suffixes. Either way, we should make it clear that this is for ansi text.

win_setupgui/src/api/functions.rs
```diff
@@ -0,0 +1,96 @@
...
+use mongoodbc::{input_text_to_string, input_wtext_to_string};
+use windows::Win32::{
+    Foundation::HWND,
+    System::Search::{ODBC_ADD_DSN, ODBC_CONFIG_DSN, ODBC_REMOVE_DSN},
+};
+
+const ODBC_MAX_DSN_LENGTH: usize = 32;
+const INVALID_DSN_TOKENS: [&str; 14] = [
+    "[", "]", "{", "}", "(", ")", ",", ";", "?", "*", "=", "!", "@", "\\",
+];
```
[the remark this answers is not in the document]
pmeredit: @nbagnard the length is defined as 32 in the header, the tokens themselves are from the spec

pmeredit (changes requested):
Looks great, just a few changes.

core/src/util/dsn.rs
```diff
@@ -0,0 +1,108 @@
...
+            }
+            get_dsn(self)
+        }
+
+        pub fn remove_datasource(&self) -> std::io::Result<()> {
+            remove_from_datasources(&self.dsn)
+        }
+
+        pub fn add_datasource(&mut self) -> std::io::Result<()> {
+            add_to_datasources(&self.dsn, &self.driver_name)
```
nbagnard, line 64: I learned something today and I'll let you decide if this needs to be updated or not.
The value associated to the DSN name under datasource is user-defined so we can do what we want. [The recommendation is to use the name of the associated DBMS](https://learn.microsoft.com/en-us/sql/odbc/reference/install/odbc-data-sources-subkey) (DBMS_NAME from constant in our case would work).

core/src/util/registry.rs
```diff
@@ -0,0 +1,114 @@
+#[cfg(target_os = "windows")]
+pub mod windows {
+    use crate::util::dsn::windows::DSNOpts;
+    use file_dbg_macros::*;
+    use std::io;
+    use winreg::{enums::HKEY_LOCAL_MACHINE, RegKey};
```
nbagnard, line 6: I am now unsure if we can by-pass calling SQLWriteDSNToIni and SQLWritePrivateProfileString because the root key is different depending on the type of DSN created :'(
It is HKEY_LOCAL_MACHINE for System DSN and HKEY_CURRENT_USER for user DSN. 
SQLGetConfigMode could tell us : https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetconfigmode-function
SQLConfigDataSource does not pass the info to ConfigDSN but maps it to User DSN .... https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlconfigdatasource-function?view=sql-server-ver16#comments

core/src/util/registry.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    const ODBC: &str = "SOFTWARE\\ODBC\\ODBC.INI";
+    const ODBC_INST: &str = "SOFTWARE\\ODBC\\ODBCINST.INI";
+    const ODBC_DATA_SOURCES: &str = "SOFTWARE\\ODBC\\ODBC.INI\\ODBC Data Sources";
+
+    pub fn add_dsn(dsn_opts: &DSNOpts) -> io::Result<()> {
```
nbagnard, line 12: I think that add_dsn and add_to_datasource go hand in hand. Same goes for remove_dsn and remove_from_datasource. One should be calling the other or the code should be in only one place, otherwise you might end up with a partial DSN config, no?

integration_test/tests/common/mod.rs
```diff
@@ -14,7 +14,7 @@ pub fn generate_default_connection_str() -> String {
     let db = env::var("ADF_TEST_LOCAL_DB");
     let driver = match env::var("ADF_TEST_LOCAL_DRIVER") {
         Ok(val) => val,
-        Err(_e) => "MongoDB Atlas SQL ODBC Driver".to_string(), //Default driver name
+        Err(_e) => "MongoDB Atlas SQL ODBC DRIVER".to_string(), //Default driver name
```
nbagnard, line 17: Why this change? The name I see in our reg setup script is `MongoDB Atlas SQL ODBC Driver` and Power-BI is using this one as well.
I don't know if driver names are case-sensitive, so maybe it does not matter. Still, I am curious about it ?

nbagnard (changes requested):
Sorry, this is a partial review. I did not look into the UI specific part in details yet, but I think it is important that we figure out the User vs system DSN part so I'd rather give my feedback now instead of waiting another day.

win_setupgui/src/gui/config.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    #[nwg_control(flags: "VISIBLE", text: "DSN")]
+    #[nwg_layout_item(layout: grid, row: 1, col: 0, col_span: 2)]
+    dsn_field: nwg::Label,
+
+    #[nwg_control(flags: "VISIBLE", text: "", focus: true)]
+    #[nwg_layout_item(layout: grid, row: 1, col: 2, col_span: 7)]
+    dsn_input: nwg::TextInput,
+
+    #[nwg_control(flags: "VISIBLE", text: "User")]
```
nbagnard, line 34:
```suggestion
    #[nwg_control(flags: "VISIBLE", text: "User name")]
```

win_setupgui/src/gui/config.rs
nbagnard, line 1: Could we hide the 'Test' button and 'LogPath' file picker until it can be hooked to something?

terakilobyte: Absolutely. It's as easy as commenting them out, mostly.

win_setupgui/src/gui/config.rs
```diff
@@ -0,0 +1,311 @@
...
+                        &format!(
+                            "DSN {dsn} {verbed} successfully",
+                            dsn = dsn_opts.dsn,
+                            verbed = self.gui_opts.verbed
+                        ),
+                    );
+                    self.close();
+                }
+                Err(e) => {
+                    dbg_write!(format!(
```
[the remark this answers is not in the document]
terakilobyte: I did. I thought it would be nice to have this in a log.

integration_test/tests/common/mod.rs
terakilobyte: Oh they absolutely are. This was probably a fat finger vim uppercase. Good catch.

core/src/util/registry.rs
terakilobyte: Ah, I segmented them because, in my mind, the datasource is under `ODBC.INI\ODBC Data Sources`, whereas dsn is under `ODBC.INI`. This is largely gone now though since I've discovered what the installer dll is actually called (odbccp32.dll) and am using those functions.

core/src/util/registry.rs
terakilobyte: SQLGetConfigMode is queryable and tells us the mode. We may end up having to go this route, as using SQLWritePrivateProfileString and SQLGetPrivateProfileString are extremely... flaky. This is something easier to show synchronously, but they have unpredictable behavior it seems.

core/src/odbc_uri.rs
```diff
@@ -9,16 +9,17 @@ const EMPTY_URI_ERROR: &str = "URI must not be empty";
...
+pub const DRIVER: &str = "driver";
+pub const DSN: &str = "dsn";
+pub const PASSWORD: &str = "password";
+pub const PWD: &str = "pwd";
+pub const SERVER: &str = "server";
+pub const USER: &str = "user";
+pub const UID: &str = "uid";
+pub const URI: &str = "uri";
+pub const APPNAME: &str = "appname";
+pub const LOGPATH: &str = "logpath";
```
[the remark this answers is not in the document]
terakilobyte: Absolutely. I can  also disable them show they aren't shown to the user, and skip their logic. If that works for you, I'll ensure to add that ticket number in comments.

win_setupgui/src/gui/config.rs
terakilobyte: I've commented out the logpath stuff and added SQL-1281 as the reason.

For the test button, I thought we wanted that since we won't ship without test functionality.

core/src/util/mod.rs
```diff
@@ -2,6 +2,10 @@ mod decimal128;
 use bson::doc;
 use constants::SQL_ALL_TABLE_TYPES;
 pub use decimal128::Decimal128Plus;
+#[cfg(target_os = "windows")]
+pub mod dsn;
+#[cfg(target_os = "windows")]
+pub(crate) mod odbcinst;
```
terakilobyte, line 8: Marking these two as windows only for now. When we support macos/linux we'll need to figure out which library to link against for the function we need, and dsn can be agnostic.

core/src/util/dsn.rs
```diff
@@ -0,0 +1,308 @@
...
+        // SQL-1281
+        // let mut logpath = String::new();
+        for (key, value) in value.iter() {
+            match key.to_lowercase().as_str() {
+                DATABASE => database = value.to_string(),
+                DSN => dsn = value.to_string(),
+                PASSWORD => password = value.to_string(),
+                PWD => password = value.to_string(),
+                SERVER => server = value.to_string(),
+                URI => server = value.to_string(),
```
nbagnard, line 179: [Discussed offline] Let's do a 1 to 1 mapping between DSNOpt and ODBCuri.

nbagnard (changes requested):
A couple of last small things that I noticed, but overall this is looking very good!!!

core/src/util/dsn.rs
```diff
@@ -0,0 +1,308 @@
...
+        ) {
+            (true, true, true, true, true, true) => Ok(Self {
+                database,
+                dsn,
+                password,
+                server,
+                user,
+                driver_name,
+            }),
+            (_, false, _, _, _, _) => Err(DSNError::DSN(dsn)),
```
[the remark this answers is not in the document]
terakilobyte: Sounds good. I'll add a comment!

core/src/util/dsn.rs
```diff
@@ -0,0 +1,308 @@
...
+                SQLGetPrivateProfileStringW(
+                    to_widechar_ptr(&self.dsn).0,
+                    to_widechar_ptr(key).0,
+                    to_widechar_ptr("").0,
+                    buffer.as_mut_ptr(),
+                    buffer.len() as i32,
+                    to_widechar_ptr(ODBCINI).0,
+                )
+            };
+            let value = unsafe { input_text_to_string_w(buffer.as_mut_ptr(), len as usize) };
```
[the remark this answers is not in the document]
terakilobyte: I will add a hard check. I can't safeguard against it, but it should be impossible based on Microsoft's registry value documentation. They could change it in the future, which could cause this to fail. 🤔 

This may be a good place for a debug write, with something like "If you see this error please report it."

win_setupgui/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition = "2021"
+
+[target.'cfg(windows)'.dependencies]
+windows = { version = "0.44.0", features = [
+    "Win32_Foundation",
+    "Win32_System_Search",
+] }
+native-windows-gui = "1.0.12"
+native-windows-derive = "1.0.3"
+odbc-sys = "0.21.2"
```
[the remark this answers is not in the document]
terakilobyte: native-windows-* is needed for the ui iself. I'll follow up on odbc-sys and remove if it's safe.

win_setupgui/src/config.rs
```diff
@@ -0,0 +1,50 @@
...
+    System::Search::{ODBC_ADD_DSN, ODBC_CONFIG_DSN, ODBC_REMOVE_DSN},
+};
+
+/// ConfigDSN adds, modifies, or deletes data sources from the system information. It may prompt the user for connection information.
+/// It can be in the driver DLL or a separate setup DLL.
+/// # Safety
+/// Because this is a C-interface, this is necessarily unsafe
+///
+#[no_mangle]
+unsafe extern "C" fn ConfigDSNW(
```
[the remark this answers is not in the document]
terakilobyte: I'm not sure if ODBC Admin uses this info, but I'll look into it.

win_setupgui/src/config.rs
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+        ODBC_CONFIG_DSN => config_dsn(dsn_opts.from_private_profile_string(), request),
+        ODBC_REMOVE_DSN => dsn_opts.remove_dsn(),
+        _ => unreachable!(),
+    }
+}
+
+fn parse_attributes(attributes: *mut cstr::WideChar) -> String {
+    unsafe { parse_attribute_string(attributes) }.replace(char::from(0), ";")
```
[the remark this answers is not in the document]
nbagnard: The logic is already geared toward a quirk from DSN options being passed, isn't it?
Maybe just update the function description to detail that it provides a `;` separated string, and we should be all set.

core/src/util/dsn.rs
```diff
@@ -0,0 +1,308 @@
...
+impl DSNOpts {
+    pub fn new(
+        database: String,
+        dsn: String,
+        password: String,
+        server: String,
+        user: String,
+        driver_name: String,
+    ) -> Result<Self, DSNError> {
+        match (
```
pmeredit, line 61: This giant tuple match feels weird. Might as well make it a vec and do
```
if v.iter().all(|x|x) {
    Ok(...)
} else if v[1] == false {
    Err(...)
} else {
    Err(other)
}
```

win_setupgui/src/gui.rs
```diff
@@ -0,0 +1,278 @@
...
+    // log_path_field: nwg::Label,
+
+    // #[nwg_resource(title: "Select Directory", action: nwg::FileDialogAction::OpenDirectory)]
+    // dialog: nwg::FileDialog,
+
+    // #[nwg_control(flags: "VISIBLE")]
+    // #[nwg_layout_item(layout: grid,  row: 6, col: 2, col_span: 1)]
+    // #[nwg_events( OnButtonClick: [ConfigGui::open_dir_picker] )]
+    // directory_button: nwg::Button,
+
```
pmeredit, line 72: Some of these commented out fields appear to be the same. Let's get rid of the duplicates.

pmeredit (changes requested):
Looks pretty good! Just two changes

win_setupgui/src/gui.rs
terakilobyte: I looked again and I didn't see any duplicates of the same name. The button in this line is required to actually pop the dialog open.

pmeredit: Ah, ok

pmeredit (approved):
LGTM

core/src/util/dsn.rs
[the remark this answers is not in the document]
nbagnard: Yes, I think a debug write with the key name could be useful. But unless they have a bunch of giant values, it should easy to figure out which key is problematic

win_setupgui/Cargo.toml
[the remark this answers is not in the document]
nbagnard: ok, so looks like it's needed after all :)

win_setupgui/src/config.rs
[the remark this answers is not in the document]
nbagnard: [Discussed offline] ODBC admin does not seem to read posted errors. So we won't post at least for now.

cstr/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    widestring::decode_utf16_lossy(v.iter().copied()).collect::<String>()
+}
+
+pub fn to_widechar_vec(s: &str) -> Vec<WideChar> {
+    widestring::encode_utf16(s.chars()).collect::<Vec<_>>()
+}
+
+///
+/// input_text_to_string_a converts a u8 cstring to a rust String.
+/// It assumes nul termination if the supplied length is negative.
```
nbagnard, line 20:
```suggestion
/// It assumes null termination if the supplied length is negative.
```

cstr/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    let mut dst = Vec::with_capacity(len);
+    dst.set_len(len);
+    copy_nonoverlapping(text, dst.as_mut_ptr(), len);
+    String::from_utf8_unchecked(dst)
+}
+
+///
+/// input_text_to_string_w converts a u16 cstring to a rust String.
+/// It assumes nul termination if the supplied length is negative.
```
nbagnard, line 47:
```suggestion
/// It assumes null termination if the supplied length is negative.
```

nbagnard (approved):
LGTM! 🎉 💯

terakilobyte:
evergreen merge

PR #103, opened 2023-03-17
SQL-1296: Add the version informations to the Windows DLL

jchemburkar:
Adds product/version info, company name, and file names to mongoodbc.dll. A couple things to note:
1. the ticket mentions not necessarily using the `winres` crate as is. The current implementation does, as the crate is fairly lightweight, and covers what we are looking for out of the box. I also tried implementing this without the cate, but it requires some shared logic and/or hardcoding of paths.
2. relatedly, this does not use a template RC file, as generating one
 with `winres` was only a few lines of code and executes quickly. However, with or without winres we could use a template to seed the values if desired.

Attached are a couple screenshots demonstrating the behavior of building using this branch:

Screenshot 2023-03-17 at 2 28 21 PM
Screenshot 2023-03-17 at 2 28 35 PM


jchemburkar:
evergreen merge

PR #105, opened 2023-03-22
SQL-1280: Add test functionality to Windows SetupGUI

terakilobyte:
I'm violating DRY a bit here, but I think it's okay since it's only twice.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,7 +45,7 @@ checksum = "76464446b8bc32758d7e88ee1a804d9914cd9b1cb264c029899680b0be29826f"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -343,7 +343,7 @@ dependencies = [
 name = "cstr"
 version = "0.0.0"
 dependencies = [
- "odbc-sys 0.21.3",
+ "odbc-sys 0.21.4",
  "widestring 1.0.2",
 ]
 
@@ -377,7 +377,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "scratch",
- "syn",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -394,7 +394,7 @@ checksum = "747b608fecf06b0d72d440f27acc99288207324b793be2c17991839f3d4995ea"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -418,7 +418,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "strsim",
- "syn",
+ "syn 1.0.109",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


win_setupgui/src/gui.rs
```diff
@@ -109,6 +110,57 @@ impl ConfigGui {
         nwg::stop_thread_dispatch();
     }
 
+    fn test_connection(&self) {
+        let dsn_opts = DSNOpts::new(
```
jchemburkar, line 114: non-blocking: maybe this could be created after checking the inputs are valid, since it doesn't appear to be used in the match? not a big difference, but might be easier to read

jchemburkar (approved):
LGTM!

win_setupgui/src/gui.rs
terakilobyte: I decided to do some more cleanup! Thanks for the suggestion.

pmeredit (changes requested):
Let's dynamically load the driver DLL so we don't explode the setup DLL.

pmeredit:
> This currently links in a lot of code from `mongoodbc.dll`, increasing the size from 1.6KB to 10KB. Maybe it'd be better to dynamically link it and expose a function that we can call instead?

Yes :D Otherwise this all looks good to me

core/src/util/dsn.rs
```diff
@@ -79,7 +80,7 @@ impl DSNOpts {
                 server,
                 driver_name,
             })
-        } else if validation[1] == false {
+        } else if !validation[1] {
```
pmeredit, line 83: Good change

Cargo.toml
```diff
@@ -8,6 +8,7 @@ members = [
   "integration_test",
   "file_dbg_macros",
   "win_setupgui",
+  "shared_sql_utils",
```
terakilobyte, line 11: I had to extract some files to a new crate to avoid triangular symbol resolution. Essentially, the setup dll *cannot* include anything from the *odbc crates because the symbols for the new function I introduced to test the connection credentials collide.

terakilobyte:
This PR now includes a re-org of some files/functionality to avoid linker errors, and pre-emptively renames the dlls so the libraries can be found. The need to do this originates from #106

jchemburkar (commented):
This LGTM, and worked like a charm on my host again. I left one question, and will come back for one final pass Monday morning!

Cargo.lock
```diff
@@ -45,7 +45,7 @@ checksum = "76464446b8bc32758d7e88ee1a804d9914cd9b1cb264c029899680b0be29826f"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.109",
```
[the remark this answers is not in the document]
terakilobyte: I didn't update any dependencies, so I assumed this was Cargo doing Cargo things.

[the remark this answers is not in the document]
jchemburkar: ah got it, makes sense

core/src/util/test_connection.rs
```diff
@@ -0,0 +1,157 @@
...
+        assert!(!result);
+        assert!(unsafe {
+            input_text_to_string_w(buffer.as_mut_ptr(), buffer_len as usize)
+                .to_lowercase()
+                .contains("server selection timeout")
+        });
+    }
+
+    // lifted and modified from integration_test\tests\connection_tests.rs
+    // this cannot be included due to dependendy issues
```
pmeredit, line 132: Minor function, fine imo.

terakilobyte:
evergreen merge

PR #108, opened 2023-03-27
SQL-1297: Support DSN connections

terakilobyte:
Adding support for DSN connections. Due to lifetime issues, I refactored ODBCUri to use Strings instead of &strs. Locally I've tested and it finds the dsn properly whether it's user or system. I could add a user dsn to `SetupDSN.reg` to programmatically test this if desired.

### DIFF
--- a/core/src/odbc_uri.rs
+++ b/core/src/odbc_uri.rs
@@ -83,12 +83,10 @@ impl ODBCUri {
     fn process_uri(odbc_uri: String) -> Result<ODBCUri> {
         let mut input = odbc_uri;
         let mut ret = ODBCUri(HashMap::new());
-        while let Some((keyword, value, rest)) = ODBCUri::get_next_attribute(input.into())? {
+        while let Some((keyword, value, rest)) = ODBCUri::get_next_attribute(input)? {
             // if attributes are repeated, the first is the one that is kept.
 
-            ret.0
-                .entry(transform_keyword(&keyword))
-                .or_insert(value.into());
+            ret.0.entry(transform_keyword(&keyword)).or_insert(value);
             if rest.is_none() {
                 break;
             }
@@ -126,7 +124,7 @@ impl ODBCUri {
         } else {
             ODBCUri::handle_unbraced_value(rest)?
         };
-        Ok(Some((keyword.to_lowercase(), value.into(), rest)))
+        Ok(Some((keyword.to_lowercase(), value, rest)))
     }
 
     fn handle_braced_value(input: &str) -> Result<(String, Option<String>)> {
@@ -279,12 +277,8 @@ impl ODBCUri {
             // set as attributes.
             let user = self.remove_mandatory_attribute(USER_KWS)?;
             let pwd = self.remove_mandatory_attribute(PWD_KWS)?;
-            client_options.credential = Some(
-                Credential::builder()
-                    .username(user.to_string())
-                    .password(pwd.to_string())
-                    .build(),
-            );
+            client_options.credential =
+                Some(Credential::builder().username(user).password(pwd).build());
         }
         Self::set_server_and_source(&mut client_options, server, source.map(String::from))?;
         client_options.app_name = client_options.app_name.or(self.handle_app_name());
@@ -295,10 +289,7 @@ impl ODBCUri {
         let user = self.remove_mandatory_attribute(USER_KWS)?;
         let pwd = self.remove_mandatory_attribute(PWD_KWS)?;
         let server = self.remove_mandatory_attribute(SERVER_KWS)?;
-        let cred = Credential::builder()
-            .username(user.to_string())
-            .password(pwd.to_string())
-            .build();
+        let cred = Credential::builder().username(user).password(pwd).build();
         Ok(ClientOptions::builder()
             .hosts(vec![
                 ServerAddress::parse(server).map_err(Error::InvalidClientOptions)?
@@ -314,7 +305,7 @@ impl ODBCUri {
                 if s == POWERBI_CONNECTOR {
                     format!("{}+{}", POWERBI_CONNECTOR, DRIVER_METRICS_VERSION.as_str())
                 } else {
-                    s.to_string()
+                    s
                 }
             })
             .or(Some(DEFAULT_APP_NAME.to_string()))
--- a/core/src/odbc_uri.rs
+++ b/core/src/odbc_uri.rs
@@ -1,8 +1,10 @@
 use crate::err::{Error, Result};
 use constants::{DEFAULT_APP_NAME, DRIVER_METRICS_VERSION};
+use file_dbg_macros::*;
 use lazy_static::lazy_static;
 use mongodb::options::{ClientOptions, Credential, ServerAddress};
 use regex::{Regex, RegexBuilder, RegexSet, RegexSetBuilder};
+use shared_sql_utils::DSNOpts;
 use std::collections::HashMap;
 
 const EMPTY_URI_ERROR: &str = "URI must not be empty";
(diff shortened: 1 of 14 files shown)


integration_test/tests/connection_tests.rs
```diff
@@ -21,4 +21,30 @@ mod integration {
...
         env.connect_with_connection_string(&conn_str).unwrap();
     }
+
+    /**
+     * The following tests require a DSN called "ADF_Test" to be configured on the machine running the tests.
+     */
+
+    #[cfg(target_os = "windows")]
+    mod test_dsn {
+        use odbc::create_environment_v3;
```
bucaojit, line 31: fyi: This odbc-rs crate does not have support for iODBC, so in my PR I added some helper functions to allocate the environment and connect.  Depending on which one is merged first, there will be some merge work with these tests.

core/src/util/test_connection.rs
```diff
@@ -23,12 +23,13 @@ pub unsafe extern "C" fn atlas_sql_test_connection(
...
     let marker = -1i8;
     let conn_str = unsafe { input_text_to_string_w(connection_string, marker as usize) };
-    if let Ok(mut odbc_uri) = ODBCUri::new(&conn_str) {
+    if let Ok(mut odbc_uri) = ODBCUri::new(conn_str) {
         match odbc_uri.try_into_client_options() {
             Ok(client_options) => {
                 match MongoConnection::connect(
                     client_options,
-                    odbc_uri.get("database").copied(),
+                    // odbc_uri.get("database").map(|s| s.as_str()),
```
bucaojit, line 31:
```suggestion
```
Extra commented code here.

core/src/odbc_uri.rs
```diff
@@ -45,35 +46,56 @@ lazy_static! {
...
 }
 
-impl<'a> ODBCUri<'a> {
-    pub fn new(odbc_uri: &'a str) -> Result<ODBCUri<'a>> {
+fn transform_keyword(keyword: &str) -> String {
+    match keyword {
+        UID => USER.to_string(),
+        USER => USER.to_string(),
+        PWD => PASSWORD.to_string(),
+        PASSWORD => PASSWORD.to_string(),
```
bucaojit, line 64: Looks like we don't need `USER` or `PASSWORD` here since the `_` branch will handle these the same way?

core/src/odbc_uri.rs
```diff
@@ -754,11 +791,32 @@ mod unit {
...
+        #[test]
+        fn supplied_args_override_dsn_args() {
+            // we leave dsn out of the uri so that it doesn't try to query the registry in the test
+            use crate::odbc_uri::ODBCUri;
+            let expected_opts = ClientOptions::parse(
+                "mongodb://foo:bar@www.atlas.net:27017/?authSource=authDB&ssl=true",
+            )
+            .unwrap();
+            let expected_cred = expected_opts.credential.unwrap();
+            let opts = ODBCUri::new("UID=foo;PWD=bar;SERVER=www.atlas.net:27017;User=foo2;Password=bar2;uri=mongodb://localhost:29000/?authSource=authDB&ssl=true".to_string())
```
bucaojit, line 812: Would you provide a little more context on this test?  
Is it to confirm that the options specified earlier in the string are not overridden by the ones specified later?

bucaojit (changes requested):
The logic for supporting the DSN entry looks good. 

I have a couple questions and suggestions.

core/src/odbc_uri.rs
terakilobyte: True true, but I wanted the transformation to be explicit. I can remove if you don't think it helps.

core/src/odbc_uri.rs
terakilobyte: Yep, that's correct! If a DSN is encountered in the connection string, it will be fetched and appended to the end of the connection string, and re-processed. However, we want earlier options to be honored first so that users can override existing dsn settings.

shared_sql_utils/src/odbcinst.rs
```diff
@@ -22,3 +22,47 @@ extern "C" {
...
+        default: *const WideChar,
+        buffer: *mut WideChar,
+        buffer_size: i32,
+        filename: *const WideChar,
+    ) -> i32;
+    pub fn SQLGetConfigMode(buffer: *mut u32) -> i32;
+}
+#[cfg(target_os = "macos")]
+#[link(name = "iodbcinst", kind = "dylib")]
+extern "C" {
```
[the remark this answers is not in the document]
terakilobyte: I tried this, and it kept giving me all sorts of issues. I'll look at using an if cfg!

integration_test/tests/connection_tests.rs
terakilobyte: Sounds good to me. Whichever we go with it shouldn't be too hard to refactor these since it's a very basic connection!

shared_sql_utils/src/dsn.rs
```diff
@@ -118,26 +118,42 @@ impl DSNOpts {
...
+                let len = unsafe {
+                    SQLGetPrivateProfileStringW(
+                        to_widechar_ptr(&self.dsn).0,
+                        to_widechar_ptr(key).0,
+                        to_widechar_ptr("").0,
+                        buffer.as_mut_ptr(),
+                        buffer.len() as i32,
+                        to_widechar_ptr(ODBCINI).0,
+                    )
+                };
```
terakilobyte, line 147: I ended up going with scanning the registry for keys users had specified. This way we're not responsible for fetching `PWD`, `pWD`, and any other permutation. Instead, we fetch the keys now and trust `set_field` to handle the returned value correctly, or ignore it.

integration_test/tests/connection_tests.rs
```diff
@@ -21,4 +21,30 @@ mod integration {
...
+        #[test]
+        fn test_valid_dsn_connection() {
+            let env = create_environment_v3().unwrap();
+            let conn_str = "DSN=ADF_Test";
+            env.connect_with_connection_string(conn_str).unwrap();
```
jchemburkar, line 36: just to double check -- is this just confirming this doesn't panic? is there any expected value we should check against (i.e. if we stored the result, like in the following test)?

jchemburkar (approved):
LGTM! I just left a couple questions

core/src/odbc_uri.rs
bucaojit: Thanks for the explanation.  Let's leave it in 👍

core/src/odbc_uri.rs
bucaojit: Ah cool, and the expected_opts is simply a way to get the `user:foo`  `pass:bar` credentials to check against.  Got it thanks!

integration_test/tests/connection_tests.rs
terakilobyte: Yeah, it will fail for numerous reasons (bad credentials or bad uri) if the info is wrong. I can link to a lot of build logs while I was figuring out how to massage the evergreen script to make this pass!

core/src/odbc_uri.rs
```diff
@@ -303,17 +318,19 @@ mod unit {
         fn get_unbraced() {
             use crate::odbc_uri::ODBCUri;
             assert_eq!(
-                ("driver".to_string(), "foo", None),
-                ODBCUri::get_next_attribute("DRIVER=foo").unwrap().unwrap(),
+                ("driver".to_string(), "foo".into(), None),
```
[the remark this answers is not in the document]
terakilobyte: No benefit at all. I'll clean this up to make it more clear what the thing is being converted to.

shared_sql_utils/src/odbcinst.rs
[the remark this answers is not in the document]
bucaojit: Awesome, I was having trouble with these cfg flags as well.  It's good to see you were able to get it to work.

bucaojit (approved):
LGTM Great work! 💯

terakilobyte:
evergreen merge

terakilobyte:
evergreen merge

PR #109, opened 2023-03-28
SQL-1327: Sign msi installer and add to release task

pmeredit:


pmeredit:
the notary service refuses to sign dlls. I'm not actually sure if dll signing is meaningful like dylib signing on macos, anyway.

evergreen.yml
```diff
@@ -154,6 +163,20 @@ functions:
...
           fi
 
+  "sign windows":
+    - command: shell.exec
+      params:
+        working_dir: mongosql-odbc-driver/installer/msi
+        script: |
+          echo "${signing_token_odbc_driver}" > ./signing_auth_token
+          /usr/local/bin/notary-client.py \
+              --key-name "bi-connector" \
```
nbagnard, line 173: Can we change the key name or is there some kind of signing variable tied to key-name, in which case we might want/need another key-name?

evergreen.yml
```diff
@@ -154,6 +163,20 @@ functions:
...
+              --auth-token-file ./signing_auth_token \
+              --comment "Evergreen Automatic Signing (Atlas SQL ODBC) - ${version_id} - ${build_variant}" \
+              --notary-url "${notary_client_url}" \
+              --skip-missing \
```
nbagnard, line 177: It seems odd to skip missing files (from the python script : "Skip missing files, instead of exiting with an error"). If the msi file is not there, it seems like a good idea to fail the signing task, isn't it?

evergreen.yml
```diff
@@ -921,39 +974,53 @@ tasks:
       - func: "install rust toolchain"
       - func: "run windows result set test"
 
+  - name: windows-sign
```
nbagnard, line 977: I think that the release task on Windows should depend on `windows-sign`

nbagnard (changes requested):
I learned something today: Where to find the signing script :D
I don't know if the key name is really important or not. Oliver might have a better idea on this.

nbagnard:
Can you also update the RELEASE.md file with the path to the signed msi?

bucaojit (changes requested):
I had one suggestion regarding the signing task.

evergreen.yml
pmeredit: I have no idea what skip-missing means, I have no suggestion that it means skip missing files.

evergreen.yml
pmeredit: Oh yea, this is a bad oversight. Surprised all my patches worked, definitely just by luck.

bucaojit (approved):
LGTM!

evergreen.yml
bucaojit: Just saw this, yes the build team assigns the key-name and the auth-token associated with it to allow for the signing.  From their [script](https://github.com/10gen/notary-service/blob/316032a525af31a303e70e3fc6831cd7d5f4f332/notary/settings.py#L88) it looks like the same `pfx` file is used for signing of different projects, which if that's the case then getting a new one would be mainly to have a matching `odbc` keyname.  Seems like something we can create a ticket for later if we really wanted to.

nbagnard: Yes, sounds good to me!

evergreen.yml
nbagnard: That's because the release task does not run on PR tests 😄 Only tags will trigger it.

nbagnard (approved):
LGTM!

evergreen.yml
pmeredit: I was running it manually ;)

pmeredit:
evergreen merge

PR #113, opened 2023-03-30
SQL-1297: support dsn connections (multiplatform)

terakilobyte:
I can get this to work on one of my macs, but not the other (I can't get any odbc connections to work on that one).

### DIFF
--- a/cstr/src/lib.rs
+++ b/cstr/src/lib.rs
@@ -1,4 +1,4 @@
-use std::ptr::copy_nonoverlapping;
+use std::{ptr::copy_nonoverlapping, str::from_utf8};
 
 pub type WideChar = u16;
 pub type Char = u8;
@@ -68,6 +68,28 @@ pub unsafe fn input_text_to_string_w(text: *const WideChar, len: usize) -> Strin
     copy_nonoverlapping(text, dst.as_mut_ptr(), len);
     from_widechar_vec_lossy(dst)
 }
+///
+/// parse_attribute_string converts a null-separted u16 doubly-null terminated cstring to a Rust
+/// string separated by `;`.
+///
+/// # Safety
+/// This converts a raw c-pointer to a Rust string, which requires unsafe operations
+///
+#[allow(clippy::uninit_vec)]
+pub unsafe fn parse_attribute_string_a(text: *const Char) -> String {
+    let mut dst = Vec::new();
+    let mut itr = text;
+    {
+        while *itr != 0 || *itr.offset(1) != 0 {
+            dst.push(*itr);
+            itr = itr.offset(1);
+        }
+    }
+    from_utf8(dst.as_slice())
+        .unwrap()
+        .replace(char::from(0), ";")
+    // from_widechar_vec_lossy(dst).replace(char::from(0), ";")
+}
 
 ///
 /// parse_attribute_string converts a null-separted u16 doubly-null terminated cstring to a Rust
@@ -77,7 +99,7 @@ pub unsafe fn input_text_to_string_w(text: *const WideChar, len: usize) -> Strin
 /// This converts a raw c-pointer to a Rust string, which requires unsafe operations
 ///
 #[allow(clippy::uninit_vec)]
-pub unsafe fn parse_attribute_string(text: *const WideChar) -> String {
+pub unsafe fn parse_attribute_string_w(text: *const WideChar) -> String {
     let mut dst = Vec::new();
     let mut itr = text;
     {
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -582,9 +582,16 @@ functions:
           echo "-------------------------"
           cat setup/iodbcinst.ini
           echo "-------------------------"
+          echo "----- DSN Setup -----"
+          sed -i "s,%DRIVER_LIB_PATH%,$DRIVER_LIB_PATH,g" setup/odbc.ini
+          sed -i "s,%ADF_TEST_DB%,${adf_test_local_db},g" setup/odbc.ini
+          sed -i "s,%ADF_TEST_USER%,${adf_test_local_user},g" setup/odbc.ini
+          sed -i "s,%ADF_TEST_PWD%,${adf_test_local_pwd},g" setup/odbc.ini
+          echo "---- END DSN Setup ----"
 
           cat <<EOT >> expansions.yml
             export ODBCINSTINI="$ODBCSYSINI/iodbcinst.ini"
+            export ODBCINI=-"$ODBCSYSINI/odbc.ini"
           EOT
     - command: expansions.update
       params:
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -523,7 +523,7 @@ functions:
       params:
         shell: bash
         working_dir: mongosql-odbc-driver
-        script:  |
+        script: |
           ${prepare_shell}
           sed -i 's@%DRIVER_DLL_PATH%@'"$(echo "$(cygpath -w $(pwd))" | sed s',\\,\\\\\\\\,g')"'@' setup/setupDSN.reg
           sed -i 's@%ADF_TEST_USER%@'"$(echo "${adf_test_local_user}" | sed s',\\,\\\\\\\\,g')"'@' setup/setupDSN.reg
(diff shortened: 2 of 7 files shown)


evergreen.yml
```diff
@@ -575,9 +582,16 @@ functions:
...
+          echo "----- DSN Setup -----"
+          sed -i "s,%DRIVER_LIB_PATH%,$DRIVER_LIB_PATH,g" setup/iodbc.ini
+          sed -i "s,%ADF_TEST_DB%,${adf_test_local_db},g" setup/iodbc.ini
+          sed -i "s,%ADF_TEST_USER%,${adf_test_local_user},g" setup/iodbc.ini
+          sed -i "s,%ADF_TEST_PWD%,${adf_test_local_pwd},g" setup/iodbc.ini
+          echo "---- END DSN Setup ----"
+
           cat <<EOT >> expansions.yml
             export ODBCINSTINI="$ODBCSYSINI/iodbcinst.ini"
+            export ODBCINI=-"$ODBCSYSINI/iodbc.ini"
```
pmeredit, line 594: Is this `-` a typo?

evergreen.yml
```diff
@@ -996,24 +1009,24 @@ functions:
...
-     - command: s3.put
-       params:
-         aws_key: ${aws_key}
-         aws_secret: ${aws_secret}
-         local_file: mongosql-odbc-driver/local_adf/logs/mongodb_test.log
-         remote_file: mongosql-odbc-driver/artifacts/logs/${version_id}/${build_variant}/mongodb_test.log
-         bucket: mciuploads
-         permissions: public-read
-         content_type: application/octet-stream
+    - command: s3.put
```
pmeredit, line 1012: Did your IDE do this? I'm fine with this sort of thing, but I know specific people who don't like to see noise like this in PRs about something else (lol) So just keep in mind when you have reviewers who aren't me that some people _really_ don't like this. I don't think @nbagnard cares either?

win_setupgui/src/config.rs
```diff
@@ -45,10 +44,9 @@ unsafe extern "C" fn ConfigDSNW(
             }
             ODBC_CONFIG_DSN => match dsn_opts.from_private_profile_string() {
                 Ok(dsn) => config_dsn(dsn, request),
-                Err(e) => {
+                Err(_e) => {
                     // we've somehow attempted to read a value from the DSN
                     // that is longer than the registry allows (at the time of writing!)
```
pmeredit, line 49: why was this dbg removed?

shared_sql_utils/src/dsn.rs
```diff
@@ -115,22 +118,49 @@ impl DSNOpts {
...
-                buffer.as_mut_ptr(),
-                buffer.len() as i32,
-                to_widechar_ptr(ODBCINI).0,
-            )
+
+        // SQL-1338 - Support reading utf16 values in dsn keys with unixodbc
+        // SQLGetPrivateProfileStringW is hopelessly broken in unixodbc. As a workaround,
+        // we must use SQLGetPrivateProfileString until if/when unixodbc is fixed or we
+        // implement SQL-1338.
+        // The implication for users is that on linux, they cannot use utf16 unicode values
```
pmeredit, line 128:
```suggestion
        // The implication for users is that on linux, they cannot use unicode
```

pmeredit: [nit]

cstr/src/lib.rs
```diff
@@ -68,16 +68,37 @@ pub unsafe fn input_text_to_string_w(text: *const WideChar, len: usize) -> Strin
...
+/// string separated by `;`.
+///
+/// # Safety
+/// This converts a raw c-pointer to a Rust string, which requires unsafe operations
+///
+#[allow(clippy::uninit_vec)]
+pub unsafe fn parse_attribute_string_a(text: *const Char) -> String {
+    let mut dst = Vec::new();
+    let mut itr = text;
+    {
```
pmeredit, line 82: This has the (slim) potential to access unitialized memory off the end of the cstring pointed to by `text: *const Char`. I think you should pass the length of the string to this function too, and add that `i + 1 < len` to the condition before the `offset` call. It makes the code less clean, so I'm sort of torn on this (since you'll have to keep track of i). I guess alternatively, you could create an end pointer as the `let end = itr.offset(len -1)`

pmeredit (changes requested):

win_setupgui/src/config.rs
terakilobyte: I've removed the debug statements from this pr, knowing that logging statements are being proposed. I wanted to get rid of as many uses of `chrono` here as I could.

cstr/src/lib.rs
terakilobyte: Unfortunately, we won't always know the length 😢 

For example, the list of attributes that are [passed in to ConfigDSN](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/configdsn-function?view=sql-server-ver16) don't include a length.

iODBC's `SQLGetPrivateProfileStringW` returns the keys of an entry, but does not return the length correctly in that it seems to not account for the null separators.

evergreen.yml
terakilobyte: Yeah it was my IDE doing some automatic formatting.

setup/iodbc.ini
```diff
@@ -0,0 +1,10 @@
+[ODBC Data Sources]
+ADF_Test = "MongoDB Atlas SQL ODBC Driver"
+
+[ADF_Test]
+Driver = %DRIVER_LIB_PATH%/libatsql.dylib
+Database = %ADF_TEST_DB%
+User = %ADF_TEST_USER%
+Password = %ADF_TEST_PWD%
+Uri = mongodb://localhost:27017
```
nbagnard, line 9: Should we change it to be able to use an `ADF_TEST_HOST` variable? Just in case we have to use a remote test instance someday? It seems that the Windows setup uses `ADF_TEST_HOST` but in a funky way. It replace it with this URI. So, maybe what we actually want to do is clean-up the Windows setup and set the uri directly in the Windows setup file.

win_setupgui/src/config.rs
nbagnard: If so, please add a `TODO - Log this` to not forget about it ;)

nbagnard (changes requested):
Looking pretty good!
I only have a few small comments.

setup/iodbc.ini
terakilobyte: I thought about this, and wavered back and forth I'll add one.

evergreen.yml
terakilobyte: I checked and I don't seem to have a default formatter set up for yaml files. I'll dig a bit to see what vscode applies by default.

odbc/Cargo.toml
```diff
@@ -16,7 +16,7 @@ lazy_static = "1.4.0"
 num-traits = "0.2.14"
 num-derive = "0.3.3"
 regex = "1.6.0"
-chrono = "0.4.22"
+chrono = "=0.4.22"                                # pinning to =0.4.22 to avoid cargo bumping patch versions and dealing with deprecated methods
```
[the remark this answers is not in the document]
terakilobyte: Yeah. I'll open a ticket to deal with them and add a todo here.

cstr/src/lib.rs
pmeredit: Yea, I'm convinced this is the best we can do.

pmeredit (approved):
LGTM!

setup/odbc.ini
```diff
@@ -0,0 +1,10 @@
+[ODBC Data Sources]
+ADF_Test = "MongoDB Atlas SQL ODBC Driver"
+
+[ADF_Test]
+Driver = %DRIVER_LIB_PATH%/libatsql.so
+Database = %ADF_TEST_DB%
+User = %ADF_TEST_USER%
+Password = %ADF_TEST_PWD%
+Uri = mongodb://localhost:27017
+UnicodeTranslationOption = utf16
```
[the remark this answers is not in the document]
terakilobyte: Removed!

setup/iodbc.ini
terakilobyte: Added

PR #114, opened 2023-03-31
SQL-1277: Add DMG installer for macos x86 build

pmeredit:
Installer works on my Mac.

As with the msi, we'll sign this in the release tasks ticket (I think we should do arch and x86 in the same release ticket).

Enjoy my commit messages :P


evergreen.yml
```diff
@@ -270,7 +270,6 @@ functions:
         working_dir: mongosql-odbc-driver
         script: |
           ${prepare_shell}
-          echo "env $(env)"
```
pmeredit, line 273: This could potentially leak secrets, I never should have had that there permanently, it was for debugging!

installer/dmg/dmg-contents/.resources/bg.png
nbagnard, line 1: Where do you need this? It looks like you are using background.png in the distribution file and I don't see this one mentioned anywhere.
I think you can get rid of the while .resources folder.

installer/dmg/resources/background.png
nbagnard, line 1: ah, almost. This is also not the latest logo.
I'll send you the updated background on Slack.
image

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,166 @@
...
+LOG_FILE="/tmp/postinstall_AtlasSQL_ODBC.log"
+
+# Log an error
+err(){
+    echo "Err: $@" | tee -a $LOG_FILE
+}
+
+# Log an general info
+info(){
+    if [[ ! -e $LOG_FILE ]]; then
```
nbagnard, line 22: [nit] Shouldn't the log file already be setup based on the lines just below?

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,166 @@
...
+    ODBC_PATH="$TARGET_VOLUME/Library/ODBC"
+fi
+INI="$ODBC_PATH/odbcinst.ini"
+info "Drivers configuration will be added to $INI"
+INSTALL_ROOT="/Library/MongoDB/AtlasSQL ODBC/"
+# Get the latest installed version
+LATEST_VERSION=$(ls "$INSTALL_ROOT/"| grep -E "[0-9]\.[0-9]\.[0-9]" | sort | cut -d '/' -f2 | tail -n 1)
+INSTALL_PATH=$INSTALL_ROOT$LATEST_VERSION
+DRIVERS_SECTION="ODBC Drivers"
+MDB="AtlasSQL ODBC $LATEST_VERSION"
```
nbagnard, line 48:
```suggestion
MDB="MongoDB Atlas SQL ODBC $LATEST_VERSION"
```

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,166 @@
...
+    > $LOG_FILE 
+fi
+
+ODBC_PATH="/Library/ODBC"
+if [[ ! -z $TARGET_VOLUME ]]; then
+    ODBC_PATH="$TARGET_VOLUME/Library/ODBC"
+fi
+INI="$ODBC_PATH/odbcinst.ini"
+info "Drivers configuration will be added to $INI"
+INSTALL_ROOT="/Library/MongoDB/AtlasSQL ODBC/"
```
nbagnard, line 43: [nit] I would sleep better without a space in the dir name. This might not be a thing anymore to avoid them in order to save you trouble.
If space are a thing, then maybe :
```suggestion
INSTALL_ROOT="/Library/MongoDB/Atlas SQL ODBC/"
```

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,166 @@
...
+# Get the latest installed version
+LATEST_VERSION=$(ls "$INSTALL_ROOT/"| grep -E "[0-9]\.[0-9]\.[0-9]" | sort | cut -d '/' -f2 | tail -n 1)
```
nbagnard, line 45: umh, so for Windows, we consider on major and minor version and patch updates override the installed driver if any. Both for Mac, we keep all versions.
Shouldn't we have the same approach for Mac, to make things easier?

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,166 @@
...
+            if [[ "$line" =~ ^\[.*\]$ ]]; then
+                PREVIOUS_SECTION=$CURRENT_SECTION
+                # Found a section
+                if grep -iqF "[ODBC Drivers]" <<< "$line"; then
+                    # Found the [ODBC Drivers] section 
+                    # where all drivers are listed
+                    ODBC_DRIVERS_SECTION_EXISTS=1
+                    CURRENT_SECTION="ODBC"
+                elif grep -iqF "[$MDB]" <<< "$line"; then
+                    # Found the [AtlasSQL Unicode ODBC {version}] section
```
nbagnard, line 114: MDB does not have mention of Unicode. I don't think we made the distinction for Windows either.
```suggestion
                    # Found the [Atlas SQL ODBC {version}] section
```

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,166 @@
...
+            else
+                if [ $CURRENT_SECTION == "ODBC" ]  && grep -iqF "$MDB" <<< "$line" ; then
+                    # Found the latest unicode MongoDB driver already configured
+                    # Make sure to mark it "Installed"
+                    echo "$MDB = Installed" >> $INI
+                    MDB_DRIVER_INSTALLED=1
+                elif [ $CURRENT_SECTION == "UNICODE" ] && grep -iqF "Driver" <<< "$line" ; then
+                    # Make sure the driver install path is correct
+                    echo "Driver = $MDB_DRIVERS" >> $INI
+                    MDB_DRIVER_CONFIGURED=1
```
nbagnard, line 136: We also need this key-value pair in the ini for Mac : `DriverUnicodeType = utf16`

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,166 @@
...
+                PREVIOUS_SECTION=$CURRENT_SECTION
+                # Found a section
+                if grep -iqF "[ODBC Drivers]" <<< "$line"; then
+                    # Found the [ODBC Drivers] section 
+                    # where all drivers are listed
+                    ODBC_DRIVERS_SECTION_EXISTS=1
+                    CURRENT_SECTION="ODBC"
+                elif grep -iqF "[$MDB]" <<< "$line"; then
+                    # Found the [AtlasSQL Unicode ODBC {version}] section
+                    CURRENT_SECTION="UNICODE"
```
nbagnard, line 115: [nit] You might want to change the name? Maybe AtlasSQLDriver? It is not incorrect though, just a little confusing I find.

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,166 @@
...
+    echo "" >> $INI
+    echo "[$MDB]" >> $INI
+    echo "Driver = $MDB_DRIVERS" >> $INI
+    echo "" >> $INI
+    $CREATED_NEW_INI_FILE=1
+fi
+
+#If there was an existing odbcinst file, update the odbcinst.ini
+# with the latest driver path
+if [[ $CREATED_NEW_INI_FILE == 0 ]]; then
```
nbagnard, line 89: I'll have to revisit what we do for Windows. If we have the driver name now incorporating the version, we want to flag this for PowerBi users until we support using a DSN for the connector too. They need to make sure that they are using the correct driver.

nbagnard (changes requested):

nbagnard:
@pmeredit Adding Nathan as a reviewer too because he working on code which touch driver keys, so his inputs can help too

terakilobyte:
Ensuring we have consistent names is going to be very important. Can we dynamically calculate the name and version from Cargo and sub that in to text files/vars?

installer/dmg/build-dmg.sh
```diff
@@ -0,0 +1,30 @@
...
+
+VERSION=$1
+ROOT="/Library/MongoDB/AtlasSQL ODBC/$VERSION"
+mkdir -p components/"$ROOT"
+mv ./libatsql.dylib components/"$ROOT"/
+cp ./resources/*.rtf components/"$ROOT"/
+cp ../../README.md components/"$ROOT"/
+
+# build component pkg
+pkgbuild --root=components/ --scripts=scripts/ --identifier='AtlasSQL ODBC' 'mongoodbc-component.pkg'
```
bucaojit, line 14:
```suggestion
pkgbuild --root=components/ --scripts=scripts/ --identifier='MongoDB Atlas SQL ODBC' 'mongoodbc-component.pkg'
```
Another spot to update the name that Natacha is suggesting

installer/dmg/resources/WELCOME.rtf
```diff
@@ -0,0 +1,15 @@
...
+{\*\listoverridetable{\listoverride\listid1\listoverridecount0\ls1}}
+\margl1440\margr1440\vieww22300\viewh20020\viewkind0
+\pard\tx720\tx1440\tx2160\tx2880\tx3600\tx4320\tx5040\tx5760\tx6480\tx7200\tx7920\tx8640\pardirnatural\partightenfactor0
+
+\f0\fs36 \cf0 Welcome to the AtlasSQL ODBC Driver Installation\
+\
+This will install two things:\
+\pard\tx220\tx720\tx1440\tx2160\tx2880\tx3600\tx4320\tx5040\tx5760\tx6480\tx7200\tx7920\tx8640\li720\fi-720\pardirnatural\partightenfactor0
+\ls1\ilvl0\cf0 {\listtext	\uc0\u8226 	}The AtlasSQL ODBC Driver for use with AtlasSQL\
+{\listtext	\uc0\u8226 	}The ODBC Manager for use in configuring ODBC Drivers and DSNs}
```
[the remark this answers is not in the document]
bucaojit: Sort of related is that we've found that the iodbc driver manager installed with the dmg installer has a different behavior than the built code.  Hopefully we can figure out the root cause, but if not maybe the alternative would be to provide it here?

installer/dmg/resources/LICENSE.rtf
```diff
@@ -0,0 +1,207 @@
+{\rtf1\ansi\deff0\nouicompat{\fonttbl{\f0\fnil\fcharset0 Courier New;}}
```
bucaojit, line 1: The formatting of paragraphs are slightly longer than the window that the installer shows so things look misaligned.  It would be nice to have a license that fits better in the window if there is one already formatted to fit.  If not then it's fine to keep as is.

bucaojit (changes requested):

installer/dmg/dmg-contents/.DS_Store
bucaojit, line 1: I don't think this file is needed.

installer/dmg/resources/WELCOME.rtf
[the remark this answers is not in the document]
pmeredit: This is what we did for BIC. It makes sense to keep, I think

installer/dmg/scripts/postinstall
pmeredit: > Ensuring we have consistent names is going to be very important. Can we dynamically calculate the name and version from Cargo and sub that in to text files/vars?

We set the name once. It’s never changing. I think we just need to get it the same.

installer/dmg/dmg-contents/.DS_Store
pmeredit: It’s needed to have the background image in the dmg, though it actually seems to not correctly display the image in the dmg currently. I’ll figure out why.

pmeredit: Ok, I found a script that I think fixes this, it's called create-dmg. Note that it actually does create a .DS_Store, but it does it in a way completely from the command line so we don't need to check in a .DS_Store that I manually created in the gui anymore.... I think. It's working locally, we'll see if it works on Evergreen.

evergreen.yml
```diff
@@ -299,6 +298,23 @@ functions:
...
+  "build dmg":
+    - command: shell.exec
+      params:
+        shell: bash
+        working_dir: mongosql-odbc-driver
+        script: |
+          ${prepare_shell}
+          cp target/release/*.dylib installer/dmg
+          cd installer/dmg
+          if [ "$RELEASE_VERSION" == "snapshot" ]; then
```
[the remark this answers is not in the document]
pmeredit: It's in the compile task. The compile task would fail on every build that wasn't a release if we didn't do this.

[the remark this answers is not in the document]
pmeredit: Well, actually, macos would probably allow a version of "snapshot" unlike windows.

installer/dmg/scripts/postinstall
pmeredit: I'm not sure why that reply to @terakilobyte ended up in this thread. Github is wonky.

pmeredit: @nbagnard we can switch to using major.minor. I cannot recall why we didn't do that in the BIC.

installer/dmg/scripts/postinstall
pmeredit: Would that first one not force all drivers to use utf16 on the system?

installer/dmg/resources/LICENSE.rtf
pmeredit: Screen Shot 2023-04-03 at 12 10 45 PM
It looks normal to me. Does it look different for you?

pmeredit: I found a spot that looked wrong and fixed it. Thanks

installer/dmg/resources/WELCOME.rtf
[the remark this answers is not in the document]
pmeredit: Removed due to ARM

pmeredit:
I believe we are good now. I don't think we want to set the global app setting to utf16?

installer/dmg/scripts/postinstall
bucaojit: Yes it would.  Let's leave the `AppUnicodeType` out then and keep the `DriverUnicodeType = utf16`.

installer/dmg/resources/LICENSE.rtf
bucaojit: Cool thanks :+1:

installer/dmg/dmg-contents/.DS_Store
bucaojit: Okay, sounds good.  
I initially thought it was checked-in on accident, thanks for explaining the significance of the file,

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,169 @@
...
+                    echo "" >> "$INI"
+                fi
+                echo "$line" >> "$INI"
+            else
+                if [ $CURRENT_SECTION == "ODBC" ]  && grep -iqF "$MDB" <<< "$line" ; then
+                    # Found the latest unicode MongoDB driver already configured
+                    # Make sure to mark it "Installed"
+                    echo "$MDB = Installed" >> "$INI"
+                    MDB_DRIVER_INSTALLED=1
+                elif [ $CURRENT_SECTION == "MongoDB Atlas SQL ODBC" ] && grep -iqF "Driver" <<< "$line" ; then
```
bucaojit, line 134:
```suggestion
                elif [[ $CURRENT_SECTION == "MongoDB Atlas SQL ODBC" ]] && grep -iqF "Driver" <<< "$line" ; then
```
This should have double brackets since the value has spaces otherwise it gets an `bash: [: too many arguments` error for the line.

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,169 @@
...
+    info ""$INI" does not exists, creating"
+    touch "$INI"
+    echo "[ODBC Drivers]" >> "$INI"
+    echo "$MDB = Installed" >> "$INI"
+    echo "" >> "$INI"
+    echo "[$MDB]" >> "$INI"
+    echo "Driver = $MDB_DRIVERS" >> "$INI"
+    echo "DriverUnicodeType = utf16" >> "$INI"
+    echo "" >> "$INI"
+    $CREATED_NEW_INI_FILE=1
```
bucaojit, line 85:
```suggestion
    CREATED_NEW_INI_FILE=1
```
No need for the `$` since we're setting the value.

installer/dmg/scripts/postinstall
bucaojit, line 134: There still seems to be one more small bug here, where if I run the installer it keeps appending the Driver and DriverUnicodeType.  I think it may be because the script finds `Driver` and `DriverUnicodeType` in the `[MongoDB Atlas SQL ODBC 0.1]` section and enters this.

bucaojit (changes requested):
Great work on the updates.  
I was trying out the installer and was seeing a few small issues with it updating the `odbcinst.ini` did a bit of troubleshooting and found these small fixes.

installer/dmg/dmg-contents/.DS_Store
pmeredit: Yep, the DS_Store is a database that controls all the Finder settings for a window from size, to opening position to background image.

installer/dmg/scripts/postinstall
pmeredit: That's... weird. I didn't write this line

installer/dmg/scripts/postinstall
pmeredit: I noticed that too, and didn't think it was an issue. Now I wonder if it is. The original postinstall script I wrote was python which was MUCH easier (an actually INI parser!), but sadly Apple removed Python from the automatically installed programs :(

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,169 @@
...
+
+                # Found a new section right after [ODBC Drivers]
+                # Need to add the drivers lines in [ODBC Drivers] section
+                if [ $PREVIOUS_SECTION == "ODBC" ]; then
+                    flagDriversAsInstalled
+                    echo "" >> "$INI"
+                fi
+                echo "$line" >> "$INI"
+            else
+                if [ $CURRENT_SECTION == "ODBC" ]  && grep -iqF "$MDB" <<< "$line" ; then
```
pmeredit, line 129:
```suggestion
                if [[ $CURRENT_SECTION == "ODBC" ]]  && grep -iqF "$MDB" <<< "$line" ; then
```

installer/dmg/scripts/postinstall
pmeredit:
```suggestion
                elif [[ $CURRENT_SECTION == "MongoDB Atlas SQL ODBC" ]] && grep -iqF "Driver" <<< "$line" ; then
```

installer/dmg/scripts/postinstall
```diff
@@ -0,0 +1,169 @@
...
+                elif grep -iqF "[$MDB]" <<< "$line"; then
+                    # Found the [MongoDB Atlas SQL ODBC {version}] section
+                    CURRENT_SECTION="MongoDB Atlas SQL ODBC"
+                else
+                    CURRENT_SECTION="OTHER"
+                fi
+
+                # Found a new section right after [ODBC Drivers]
+                # Need to add the drivers lines in [ODBC Drivers] section
+                if [ $PREVIOUS_SECTION == "ODBC" ]; then
```
pmeredit, line 123:
```suggestion
                if [[ $PREVIOUS_SECTION == "ODBC" ]]; then
```

installer/dmg/scripts/postinstall
pmeredit:
```suggestion
                if [[ $CURRENT_SECTION == "ODBC" ]]  && grep -iqF "$MDB" <<< "$line" ; then
```

installer/dmg/scripts/postinstall
pmeredit:
```suggestion
                elif [[ $CURRENT_SECTION == "MongoDB Atlas SQL ODBC" ]] && grep -iqF "Driver" <<< "$line" ; then
```

pmeredit:
Ok, rust based postinstall script seems to work great.

macos_postinstall/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+fn write_odbc_file(path: &str, ini: Ini) {
+    ini.write_to_file(path).unwrap()
+}
+
+fn main() {
+    let args = env::args().collect::<Vec<_>>();
+    let target_volume = args[3].clone();
```
bucaojit, line 90: nit: I think we could add a check to make sure we have enough arguments passed in so we don't get an index out of range error her.

bucaojit (approved):
LGTM. 

The post install program is much more readable in Rust :rocket: 
I had one small nit, and looks like the `WELCOME.rtf` has some cleanup comments to implement.

evergreen.yml
```diff
@@ -347,11 +375,22 @@ functions:
...
       params:
         build_variants:
-          - windows-64
+          - macos
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: mongosql-odbc-driver/target/release/atsql.pdb
-        remote_file: mongosql-odbc-driver/artifacts/${version_id}/${build_variant}/release/atsql.pdb
+        local_file: mongosql-odbc-driver/target/release/libatsql.dylib
+        remote_file: mongosql-odbc-driver/artifacts/${version_id}/${build_variant}/release/libatsql.dylib
```
nbagnard, line 382: Do we have pdb files to upload as well?

macos_postinstall/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+use std::{
+    env,
+    fs::{self, File},
+    io::Write,
+    sync::Mutex,
+};
+
+const LOG_FILE: &str = "/tmp/postinstall_MongoDB_Atlas_SQL_ODBC.log";
+const ODBC_PATH: &str = "/Library/ODBC";
+const INSTALL_ROOT: &str = "/Library/MongoDB/MongoDB Atlas SQL ODBC";
```
nbagnard, line 13: The name of the driver key is actually "MongoDB Atlas SQL ODBC Driver" I checked the PowerBI and ODBC code.
But please triple check.

installer/dmg/build-dmg.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+
+rm ./*.dmg || true
+rm dmg-contents/*.pkg || true
+
+VERSION=$1
+ROOT="/Library/MongoDB/MongoDB Atlas SQL ODBC/$VERSION"
```
nbagnard, line 7:
```suggestion
ROOT="/Library/MongoDB/MongoDB Atlas SQL ODBC Driver /$VERSION"
```

installer/dmg/distribution.xml
```diff
@@ -0,0 +1,16 @@
...
+    <title>MongoDB Atlas SQL ODBC Driver __VERSION__</title>
+    <welcome file="WELCOME.rtf" mim-type="application/rtf" />
+    <license file="LICENSE.rtf" mim-type="application/rtf" />
+    <background file="background.png" mime-type="image/png" scaling="tofit" />
+    <pkg-ref id="mongodb_odbc"/>
+    <options customize="allow" require-scripts="false"/>
+    <choices-outline>
+        <line choice="mongodb_odbc"/>
+    </choices-outline>
+    <choice id="mongodb_odbc" title="MongoDB Atlas SQL ODBC" enabled="false" visible="true">
```
nbagnard, line 12: [nit] Maybe good to add `Driver` everywhere it's missing for consistency.
```suggestion
    <choice id="mongodb_odbc" title="MongoDB Atlas SQL ODBC Driver" enabled="false" visible="true">
```

installer/dmg/resources/WELCOME.rtf
```diff
@@ -0,0 +1,15 @@
+{\rtf1\ansi\ansicpg1252\cocoartf1561\cocoasubrtf400
+{\fonttbl\f0\fswiss\fcharset0 Helvetica;}
+{\colortbl;\red255\green255\blue255;}
+{\*\expandedcolortbl;;}
+{\*\listtable{\list\listtemplateid1\listhybrid{\listlevel\levelnfc23\levelnfcn23\leveljc0\leveljcn0\levelfollow0\levelstartat1\levelspace360\levelindent0{\*\levelmarker \{disc\}}{\leveltext\leveltemplateid1\'01\uc0\u8226 ;}{\levelnumbers;}\fi-360\li720\lin720 }{\listname ;}\listid1}}
+{\*\listoverridetable{\listoverride\listid1\listoverridecount0\ls1}}
+\margl1440\margr1440\vieww22300\viewh20020\viewkind0
+\pard\tx720\tx1440\tx2160\tx2880\tx3600\tx4320\tx5040\tx5760\tx6480\tx7200\tx7920\tx8640\pardirnatural\partightenfactor0
+
+\f0\fs36 \cf0 Welcome to the AtlasSQL ODBC Driver Installation\
```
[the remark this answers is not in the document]
nbagnard: I think you missed this one ;)

installer/dmg/resources/WELCOME.rtf
```diff
@@ -0,0 +1,15 @@
...
+\
+This will install two things:\
+\pard\tx220\tx720\tx1440\tx2160\tx2880\tx3600\tx4320\tx5040\tx5760\tx6480\tx7200\tx7920\tx8640\li720\fi-720\pardirnatural\partightenfactor0
+\ls1\ilvl0\cf0 {\listtext	\uc0\u8226 	}The AtlasSQL ODBC Driver for use with AtlasSQL\
```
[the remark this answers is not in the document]
nbagnard: And that one. Unless Git is playing tricks on me. But reloading still shows the same.

installer/dmg/resources/WELCOME.rtf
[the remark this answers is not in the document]
nbagnard: You want to remove this line or change it to tell them that they can use the ODBC manager to view their config/setup DSN(s)

macos_postinstall/src/main.rs
nbagnard: That looks like a very good idea from @bucaojit.
I couldn't figure out where are the args passed. I bet this is DMG magic. Better make sure we just don't crash

macos_postinstall/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+use lazy_static::lazy_static;
+use std::{
+    env,
+    fs::{self, File},
+    io::Write,
+    sync::Mutex,
+};
+
+const LOG_FILE: &str = "/tmp/postinstall_MongoDB_Atlas_SQL_ODBC.log";
+const ODBC_PATH: &str = "/Library/ODBC";
```
nbagnard, line 12: We only install the driver in the system path.Same as BIC, right?

macos_postinstall/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+use itertools::sorted;
+use lazy_static::lazy_static;
+use std::{
+    env,
+    fs::{self, File},
+    io::Write,
+    sync::Mutex,
+};
+
+const LOG_FILE: &str = "/tmp/postinstall_MongoDB_Atlas_SQL_ODBC.log";
```
nbagnard, line 11: I found that my tmp directory on mac isn't just /tmp/ but something more cumbersome (kind of like Windows).
@terakilobyte Do you know if it matters? Because if we can write directly into /tmp/ it seems just as good.
Otherwise, maybe we want to use what we have in the ODBC logger
`env::temp_dir();`

macos_postinstall/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+        target_volume + "/" + ODBC_PATH
+    } else {
+        ODBC_PATH.to_string()
+    };
+    let ini_file = odbc_path.clone() + "/odbcinst.ini";
+    info(&format!(
+        "Drivers configuration will be added to {ini_file}"
+    ));
+    let latest = get_latest_version();
+    let mdb_driver_key = format!("MongoDB Atlas SQL ODBC {latest}");
```
nbagnard, line 101: I think we said we were to get rid of the version in this PR and did not need to create a follow up JIRA.
The driver key is :
```suggestion
    let mdb_driver_key = format!("MongoDB Atlas SQL ODBC Driver");
```

macos_postinstall/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+    let res = fs::create_dir_all(&odbc_path);
+    if res.is_err() {
+        err(&format!("Failed to create ODBC_PATH: {res:?}"));
+        panic!();
+    }
+
+    let mut ini = parse_odbc_file(&ini_file);
+    let drivers_section = ini.section_mut(Some(DRIVERS_SECTION));
+    if let Some(drivers_section) = drivers_section {
+        drivers_section.insert(mdb_driver_key.clone(), "Installed");
```
nbagnard, line 116: Will this duplicate the key when doing an upgrade?

macos_postinstall/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+    } else {
+        ini.with_section(Some(DRIVERS_SECTION))
+            .set(mdb_driver_key.clone(), "Installed");
+    }
+
+    ini.with_section(Some(mdb_driver_key))
+        .set("Description", "MongoDB Atlas SQL ODBC Driver")
```
nbagnard, line 123: If you want to add the version mention somewhere, here is the spot

macos_postinstall/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+    let mut ini = parse_odbc_file(&ini_file);
+    let drivers_section = ini.section_mut(Some(DRIVERS_SECTION));
+    if let Some(drivers_section) = drivers_section {
+        drivers_section.insert(mdb_driver_key.clone(), "Installed");
+    } else {
+        ini.with_section(Some(DRIVERS_SECTION))
+            .set(mdb_driver_key.clone(), "Installed");
+    }
+
+    ini.with_section(Some(mdb_driver_key))
```
nbagnard, line 122: How will this work when upgrading?

macos_postinstall/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+    let odbc_path = if target_volume != "/" {
+        target_volume + "/" + ODBC_PATH
+    } else {
+        ODBC_PATH.to_string()
+    };
+    let ini_file = odbc_path.clone() + "/odbcinst.ini";
+    info(&format!(
+        "Drivers configuration will be added to {ini_file}"
+    ));
+    let latest = get_latest_version();
```
nbagnard, line 100: Can we stop the process in case the user is downgrading?

nbagnard (changes requested):

evergreen.yml
pmeredit: pdb files are windows

macos_postinstall/src/main.rs
pmeredit: nope, I checked that. It's treated as a map so duplicate keys are not allowed.

macos_postinstall/src/main.rs
pmeredit: It replaces the old one.

macos_postinstall/src/main.rs
pmeredit: Tripled checked correct

macos_postinstall/src/main.rs
pmeredit: Yes, I didn't update this since yesterday ;)

macos_postinstall/src/main.rs
pmeredit: Agreed!

macos_postinstall/src/main.rs
pmeredit: As mentioned, this works fine with downgrading.

evergreen.yml
nbagnard: Cool! 
Who doesn't like elf and dwarf, but elves and dwarfs themselves lol

macos_postinstall/src/main.rs
nbagnard: Perfect!

macos_postinstall/src/main.rs
nbagnard: All good! Thanks for testing this out :)

pmeredit:
@nbagnard I've verified that this last built driver is doing what we want on macos

installer/dmg/resources/WELCOME.rtf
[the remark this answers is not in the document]
pmeredit: I added a link to the ODBC Manager page

installer/dmg/resources/WELCOME.rtf
```diff
@@ -0,0 +1,9 @@
+{\rtf1\ansi\ansicpg1252\deff0\nouicompat{\fonttbl{\f0\fswiss\fcharset0 Helvetica;}}
+{\*\generator Riched20 10.0.22621}\viewkind4\uc1 
+\pard\tx720\tx1440\tx2160\tx2880\tx3600\tx4320\tx5040\tx5760\tx6480\tx7200\tx7920\tx8640\f0\fs36\lang9 Welcome to the MongoDB Atlas SQL ODBC Driver Installation\par
+\par
+This will install the MongoDB AtlasSQL ODBC Driver for use with Atlas SQL.\par
+We also recommend installing {\field{\*\fldinst HYPERLINK "http://www.odbcmanager.net/"}{\fldrslt ODBC Manager}} for creating DSNs\par
+\par
+}
+
```
nbagnard, line 9: Github is showing that the file is missing a new line at the end.

nbagnard (approved):
LGTM

pmeredit:
evergreen merge

PR #115, opened 2023-03-31
SQL-1337: Fix for Power BI error 'table has no visible columns'

bucaojit:
The generated schema for a row with an empty array has no bsonType specified, it also has ` "additionalProperties": false`.  
This change takes into account the possible additionalProperties: false to avoid returning an error and instead return a `BsonTypeName::Any`.

```
{ "_id" : 0, "entry" : [ ] }
{ "_id" : 1, "entry" : [ ] }
```
Schema:
```
{  "bsonType": [ "object" ],
  "properties": {
    "_id": { "bsonType": [ "double"],  "additionalProperties": false},
    "entry": {
      "bsonType": [ "array"],
      "items": {
        "additionalProperties": false
      },
      "additionalProperties": false
    }
  },
  "additionalProperties": false}
```
We currently don't log an error when this happens, there is an [existing PR](https://github.com/mongodb/mongo-odbc-driver/pull/92) to add the error message so I did not include one here.

### DIFF
--- a/core/src/json_schema.rs
+++ b/core/src/json_schema.rs
@@ -263,7 +263,7 @@ pub mod simplified {
                     bson_type: None,
                     properties: None,
                     required: None,
-                    additional_properties: None,
+                    additional_properties: None | Some(false),
                     items: None,
                     any_of: None,
                 } => Ok(Atomic::Scalar(BsonTypeName::Any)),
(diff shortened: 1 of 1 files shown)


terakilobyte (approved):
Nice fix! LGTM

jchemburkar (approved):
LGTM!

pmeredit:
I'm not convinced we even need an error because I think this is a valid schema. The empty schema should also be treated as Any.

core/src/json_schema.rs
```diff
@@ -263,7 +263,7 @@ pub mod simplified {
                     bson_type: None,
                     properties: None,
                     required: None,
-                    additional_properties: None,
+                    additional_properties: None | Some(false),
```
pmeredit, line 266: Wait... when did they add non-top level |?!!! This is awesome and would simplify so much of my code. This definitely wasn't in rust 1.0.

pmeredit (approved):
LGTM!

core/src/json_schema.rs
bucaojit: Cool!  This looks like the version that added the feature: 
https://blog.rust-lang.org/2021/06/17/Rust-1.53.0.html#or-patterns

bucaojit:
evergreen merge

PR #116, opened 2023-03-31
SQL-1281: Logging

terakilobyte:
This PR touches a lot of files, mainly a lot of small changes. It introduces the `log4rs` logging framework, and usage of the `log::{error, info, debug}` macros.

I tried to be as safe as possible in the logger initialization, but wrapped it in a panic catch so that failure to instantiate the logger would not impact the operation of the driver.

I thought about using trace rather than debug, but turns out the runtime puts some info in the trace as well, so opted to not do that. If we need to log things out for debugging purposes, we can always use `log::warn` and easily grep the logs.

I chose a file appender strategy with a SWAG size for file rotation and up to 10 logs.

Users will be able to override the log level at connection time by supplying the `loglevel` connection attribute.


integration_test/tests/connection_tests.rs
```diff
@@ -20,6 +20,7 @@ mod integration {
     fn test_default_connection() {
         let env_handle = allocate_env();
         let conn_str = crate::common::generate_default_connection_str();
+        // let conn_str = format!("{conn_str};LOGLEVEL=debug");
```
terakilobyte, line 23: I left this here on purpose to demonstrate dynamic log level configuration. I still need to figure out how to test this.

logger/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+            unsafe {
+                SQLGetPrivateProfileString(
+                    to_char_ptr("MongoDB Atlas SQL ODBC Driver").0,
+                    to_char_ptr("Driver").0,
+                    to_char_ptr("").0,
+                    buffer.as_mut_ptr(),
+                    buffer.len() as i32,
+                    to_char_ptr("odbcinst.ini").0,
+                )
+            };
```
terakilobyte, line 41: Why SQLGetPrivateProfileString instead of SQLGetPrivateProfileStringW? Just a choice, but easily converted if we want.

odbc/src/api/functions.rs
```diff
@@ -1053,6 +1062,18 @@ fn sql_driver_connect(conn: &Connection, odbc_uri_string: &str) -> Result<MongoC
...
+    if let Some(log_level) = odbc_uri.remove(&["loglevel"]) {
+        let env = unsafe { conn.env.as_ref() };
+        if let Some(env) = env {
+            if let Some(env) = env.as_env() {
+                if let Some(logger) = env.logger.as_ref() {
+                    logger.set_log_level(log_level);
+                }
+            }
+        }
+    }
```
terakilobyte, line 1075: This feels tedious, but I think it's better than match for this purpose.

cstr/src/lib.rs
```diff
@@ -89,6 +89,26 @@ pub unsafe fn parse_attribute_string(text: *const WideChar) -> String {
     from_widechar_vec_lossy(dst).replace(char::from(0), ";")
 }
 
+///
+/// parse_attribute_string converts a null-separted u16 doubly-null terminated cstring to a Rust
```
mattChiaravalloti, line 93:
```suggestion
/// parse_attribute_string_a converts a null-separted u16 doubly-null terminated cstring to a Rust
```
Should we include the suffix in this function comment, and the `_w` for the previous function? And should we distinguish their behaviors in the comments as well? Right now, the comments for `parse_attribute_string_w` and `parse_attribute_string_a` are identical, which feels incorrect.

evergreen.yml
```diff
@@ -538,10 +538,11 @@ functions:
         working_dir: mongosql-odbc-driver
         script:  |
           ${prepare_shell}
+          echo "Minor version: $MINOR_VERSION"
```
mattChiaravalloti, line 541: This may be debugging code but tbh I find this kind of stuff useful in evergreen tasks. Was this intentionally left in the PR? Are there other things you had in here one point but removed (maybe `echo Major version: "$MAJOR_VERSION"`?).

logger/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+use log4rs::Handle;
+use log4rs::{
+    append::rolling_file::{
+        policy::compound::{
+            roll::fixed_window::FixedWindowRoller, trigger::size::SizeTrigger, CompoundPolicy,
+        },
+        RollingFileAppender,
+    },
+    config::{Appender, Config, Root},
+};
```
mattChiaravalloti, line 14:
```suggestion
use log4rs::{
    append::rolling_file::{
        policy::compound::{
            roll::fixed_window::FixedWindowRoller, trigger::size::SizeTrigger, CompoundPolicy,
        },
        RollingFileAppender,
    },
    config::{Appender, Config, Root},
    encode::pattern::PatternEncoder,
    Handle,
};
```
[nit] grouping `use` statements

odbc/src/api/mod.rs
```diff
@@ -55,15 +55,34 @@ macro_rules! set {
...
+    ($level:ident, $mongo_handle:expr, $info:expr, $fct_name:expr) => {
+        let handle_info = $mongo_handle.get_handle_info();
+        let message = format!("{handle_info} {}:: {}", $fct_name, $info);
+        $level!("{}", message);
+    };
+}
+
+///
+/// Adds a line in the trace formatted like this [{handle info}]{function name} - "{message to log}"]
+/// The handle info provide the address of the current handle and it's parent handle.
```
mattChiaravalloti, line 72:
```suggestion
/// The handle info provides the address of the current handle and it's parent handle.
```

shared_sql_utils/src/dsn.rs
```diff
@@ -38,12 +39,12 @@ pub struct DSNOpts {
...
     pub user: String,
     pub server: String,
-    // SQL-1281
-    // pub logpath: String,
     pub driver_name: String,
+    pub log_level: String,
 }
 
 impl DSNOpts {
+    #[allow(clippy::too_many_arguments)]
```
mattChiaravalloti, line 47: We could avoid this with a specialized `struct` for the args, that way when we invoke this `new()` constructor, the arguments will all be clearly named/ordered.

logger/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+impl Logger {
+    pub fn new() -> Option<Self> {
+        // Due to numerous reasons why the logger could fail to initialize, we wrap it in a catch_unwind
+        // so that logger failure does not cause our dll to crash.
+        match std::panic::catch_unwind(|| {
+            let mut buffer = [0u16; 1024];
+            unsafe {
+                SQLGetPrivateProfileStringW(
```
mattChiaravalloti, line 33: I wonder about our level of abstraction here. Instead of calling odbc functions here in the `logger` crate, `Logger::new()` can take `driver_path` as an argument, and we can have a util function in the odbc crate that does this work and calls `Logger::new()` with the appropriate `driver_path` value.

mattChiaravalloti (changes requested):
Impressive work 💯 For the most part, this was easy for me to follow. Only a handful of meaningful comments and questions.

logger/src/lib.rs
terakilobyte: I thought about this. My worry was this would add weight to the initial `SQLAllocHandle` function, and my design hope was to initialize the logger such that it was completely segmented from the main driver and avoid interfacing with the SQL Installer functions as much as possible (supporting DSN connections being an area where it absolutely needs to).

That said, I was looking for a reason to change the entry to accept a path to make testing it easier, and this is reason enough!

evergreen.yml
terakilobyte: I had it for debugging purposes when I thought we were going to have to support side-by-side. This is the only thing I added because provided the major and minor version, or said "snapshot".

cstr/src/lib.rs
terakilobyte: Yes, absolutely. I actually changed these in #113 , see [here](https://github.com/mongodb/mongo-odbc-driver/pull/113/files#diff-2f5e9a143c4e8edf9d00c1f6589095b53467dbf69f8e1438957e8c3ed7671f6eR71-R79) and [here](https://github.com/mongodb/mongo-odbc-driver/pull/113/files#diff-2f5e9a143c4e8edf9d00c1f6589095b53467dbf69f8e1438957e8c3ed7671f6eR95-R103).

I can also change it here and deal with the small merge conflict.

mattChiaravalloti: Oh thanks for the links. Looks good 👍

logger/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+#[cfg(test)]
+mod driver {
+
+    use std::{fs, path::Path};
+
+    use super::*;
+    use log::{debug, error, info};
+
+    #[test]
+    fn logger() {
```
mattChiaravalloti, line 113: Why do the 3 log levels we use (`info`, `debug`, `error`) result in the same lines being logged? Aren't the filters supposed to filter out certain log messages? Like `debug` includes everything, `info` omits `debug!` logs, and `error` omits `debug!` and `info!` logs? I may be misunderstanding this.

mattChiaravalloti (commented):
Looks good but I'll hold off approving since I have one last question that may or may not result in changes.

logger/src/lib.rs
terakilobyte: Logs are hierarchical. error level is the highest we use, which filters out info and debug. Likewise, info will filter out debug, but not error. Debug includes everything above it. This way we still see error messages when the loglevel is at info, for example.

mattChiaravalloti: Oh my goodness I entirely missed the `!`s in the `assert!`s. That's why I was so confused. I read this as all three log levels including all 3 log lines every time.

mattChiaravalloti (approved):

logger/src/lib.rs
terakilobyte: and I misread your comment!

core/src/fields.rs
```diff
@@ -550,8 +550,10 @@ impl MongoFields {
                             }
                         }
                         // If there is an error simplifying the schema (e.g. an AnyOf), skip the collection
-                        // TODO: SQL-1198: Add a log or warning
```
pmeredit, line 553: We should update SQL-1198, perhaps even closing it?

core/src/odbc_uri.rs
```diff
@@ -73,7 +73,7 @@ impl ODBCUri {
         }
         let mut ret = ODBCUri::process_uri(odbc_uri.clone())?;
         if ret.get(DSN).is_some() {
-            let mut dsn_opts = DSNOpts::from_attribute_string(&odbc_uri);
+            let mut dsn_opts = DSN::from_attribute_string(&odbc_uri);
```
pmeredit, line 76: hmm, interesting that rust can actually distinguish the DSN str from the DSN struct. The rust style guide would prefer Dsn for the struct (opposite of the go style guide): https://rust-lang.github.io/api-guidelines/naming.html

So lets go ahead and rename the struct DSN to Dsn.

cstr/src/lib.rs
```diff
@@ -89,6 +91,28 @@ pub unsafe fn parse_attribute_string(text: *const WideChar) -> String {
...
 
+///
+/// parse_attribute_string_a converts a null-separated doubly null terminated *Char string to a Rust
+/// string separated by `;`.
+///
+/// # Safety
+/// This converts a raw c-pointer to a Rust string, which requires unsafe operations. Additionally, it
+/// has the small possibility of reading into unallocated memory should the input string not be doubly
+/// null terminated. Only use this method if you are certain the input string is doubly null terminated.
+///
```
pmeredit, line 102: Good comment

integration_test/tests/connection_tests.rs
```diff
@@ -22,6 +26,49 @@ mod integration {
...
+        };
+
+        let driver_path = unsafe { cstr::parse_attribute_string_w(buffer.as_mut_ptr()) };
+
+        let log_file_path = Path::new(&driver_path)
+            .parent()
+            .unwrap()
+            .parent()
+            .unwrap()
+            .join("logs/mongo_odbc.log")
```
pmeredit, line 62: The advantage of join is it uses the correct file separator, so I think we should join `logs` and `mongo_odbc.log` even though Windows pretty much doesn't care \ vs / anymore.

I know I'm using `/` explicitly in the postinstall script for macos, but it's specifically not to be used on windows (lol)

logger/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+    }
+
+    fn file_appender(log_dir: &str) -> Result<RollingFileAppender, std::io::Error> {
+        let file_path = format!("{log_dir}/mongo_odbc.log");
+        let roller_pattern = format!("{log_dir}/mongo_odbc.log.{{}}");
+
+        let roller = FixedWindowRoller::builder()
+            .build(&roller_pattern, 10)
```
pmeredit, line 82: Just going to assume the logging library correctly handles windows vs nix

pmeredit (changes requested):
Looks great! Just a few minor changes

pmeredit:
I'm also unsure why windows integration test is failing when the result-set tests are passing. I tried to rerun it and it's still failing :/

integration_test/tests/connection_tests.rs
terakilobyte: In testing it was able to handle the forward slash, but I've now changed it everywhere to use `Path` to make doubly sure.

nbagnard (changes requested):
Just a few comments! Looking good so far!

evergreen.yml
```diff
@@ -538,10 +538,11 @@ functions:
...
           sed -i 's@%DRIVER_DLL_PATH%@'"$(echo "$(cygpath -w $(pwd))" | sed s',\\,\\\\\\\\,g')"'@' setup/setupDSN.reg
           sed -i 's@%ADF_TEST_USER%@'"$(echo "${adf_test_local_user}" | sed s',\\,\\\\\\\\,g')"'@' setup/setupDSN.reg
           sed -i 's@%ADF_TEST_PWD%@'"$(echo "${adf_test_local_pwd}" | sed s',\\,\\\\\\\\,g')"'@' setup/setupDSN.reg
-          sed -i 's@%ADF_TEST_HOST%@'"$(echo "mongodb://localhost:27017" | sed s',\\,\\\\\\\\,g')"'@' setup/setupDSN.reg
+          sed -i 's@%ADF_TEST_HOST%@'"$(echo "${adf_test_host}" | sed s',\\,\\\\\\\\,g')"'@' setup/setupDSN.reg
```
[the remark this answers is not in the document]
terakilobyte: I've also renamed the placeholder to `ADF_TEST_URI`

logger/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+        let driver_path = driver_path.into();
+        // Due to numerous reasons why the logger could fail to initialize, we wrap it in a catch_unwind
+        // so that logger failure does not cause our dll to crash.
+        match std::panic::catch_unwind(|| {
+            let log_dir = if driver_path.is_empty() {
+                std::env::temp_dir()
+            } else {
+                let path = Path::new(&driver_path);
+                path.parent()
+                    .map(|p| p.parent().map(|p| p.join("logs")).unwrap())
```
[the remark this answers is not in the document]
terakilobyte: I've added a full doc comment to the `new` function!

terakilobyte:
The integration test that inspects the logs has been extremely flaky on evergreen. I may comment the entire thing out if it doesn't show stability.

logger/src/lib.rs
```diff
@@ -21,6 +21,16 @@ pub struct Logger {
...
 impl Logger {
+    /// Initializes the logger with the given path as its root. The logger will create a logs folder
+    /// two levels up from the given path and write its logs there. For example, if the given path is
+    /// `C:\\Program Files\\mongo_odbc\\1.1\\bin\\atsql.dll`, the logger will write its logs to
+    /// `C:\\Program Files\\mongo_odbc\\1.1\\logs\\mongo_odbc.log`. `std::path::Path` is used to assemble
+    /// the paths so this is OS safe.
+    ///
+    /// If the given path is empty, or there is an error accessing the path, the logger will write its logs to the temp directory.
+    ///
+    /// The logger is wrapped in a catch_unwind so that logger failure does the driver to crash. In this case
```
nbagnard, line 32:
```suggestion
    /// The logger is wrapped in a catch_unwind so that logger failure does not cause the driver to crash. In this case
```

nbagnard (approved):
LGTM!

pmeredit (approved):
LGTM!

PR #119, opened 2023-04-14
add signing tasks as part of snapshot

terakilobyte:
Short one, just updating the snapshot task so we don't get a "system failed" when manually running it.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -1472,6 +1472,10 @@ tasks:
         variant: "*"
       - name: asan-integration-test
         variant: "*"
+      - name: windows-sign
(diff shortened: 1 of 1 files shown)


nbagnard (approved):
Sounds good to me, although I am not sure we care to sign the snapshot. The other option if signing becomes a bottleneck will be to extract copying the signed artifact in the S3 bucket for snapshot and only do it for release.

terakilobyte:
I don't think we care, but it's done on the periodic build so may as well do it here.

terakilobyte:
evergreen merge

PR #122, opened 2023-04-20
SQL-1350: Update WiX Installer to enable more than one version installed

terakilobyte:
This PR introduces the ability for users to install multiple versions of the driver "side-by-side" with the installer updating the Driver and Setup values under `ODBCINST/MongoDB Atlas SQL ODBC Driver` properly. We can't use the `ODBCDriver` element because if it encounters the `ODBCINST/MongoDB Atlas SQL ODBC Driver` entry it will override the BIN directory the installer specifies to the BIN directory of the original location, and write the files there.

Additionally, it updates `logger` to write to the user's documents folder. I introduced the Directories crate to abstract this and ensure cross-platform compatibility. This change needed to be introduced because Program Files is/can be locked down to restricted access, resulting in no log files being generated.

I'm still investigating proper repair functionality. When a user uninstalls, it will delete all files and `RegistryValue`s the installer created. In order to restore the the values, the user can use a previous installation to repair. However, this seems to require the original MSI associated with that installation is still present and available. It looks like a custom script can be run, allowing us to copy the MSI to the installation directory. I also looked at network repair, i.e. giving the url to the original MSI to try to redownload the MSI and run the repair, but that looks less likely.


installer/msi/BinaryFragment.wxs
```diff
@@ -1,29 +1,24 @@
...
+    <Fragment>
+        <DirectoryRef Id="BIN">
+            <?if $(var.Platform)=x64?>
+            <?define win64Flag = "yes"?>
+            <?else ?>
+            <?define win64Flag = "no"?>
+            <?endif ?>
+            <Component Id="atsql.files" Guid="*" Win64="$(var.win64Flag)">
+                <File Id="atsql.dll" Name="atsql.dll" Source="atsql.dll" />
+                <File Id="atsqls.dll" Name="atsqls.dll" Source="atsqls.dll" />
```
terakilobyte, line 12: I changed the names here after looking at other installers and considered it a simplification.

installer/msi/BinaryFragment.wxs
```diff
@@ -1,29 +1,24 @@
...
+                <RegistryKey ForceCreateOnInstall="yes" ForceDeleteOnUninstall="yes" Id="MongoDBKey" Key="SOFTWARE\MongoDB\[ProductName]" Root="HKLM" />
+                <RegistryValue Id="Version" Key="SOFTWARE\MongoDB\[ProductName]" Name="Version" Root="HKLM" Type="string" Value="[ProductVersion]" />
+                <RegistryValue Id="mongodb.reg.1" Root="HKLM" Key="SOFTWARE\ODBC\ODBCINST.INI\ODBC Drivers" Name="MongoDB Atlas SQL ODBC Driver" Type="string" Value="Installed" />
+                <RegistryValue Id="mongodb.reg.2" Root="HKLM" Key="SOFTWARE\ODBC\ODBCINST.INI\MongoDB Atlas SQL ODBC Driver" Name="Driver" Type="string" Value="[#atsql.dll]" />
+                <RegistryValue Id="mongodb.reg.3" Root="HKLM" Key="SOFTWARE\ODBC\ODBCINST.INI\MongoDB Atlas SQL ODBC Driver" Name="Setup" Type="string" Value="[#atsqls.dll]" />
```
terakilobyte, line 17: Using the `ODBCDriver` element isn't possible with our use-case. When it encounters and existing key, it changes the bin directory to the first installation and writes the new files there. So, we have to manually write the entries.

installer/msi/build-msi.ps1
```diff
@@ -6,40 +6,113 @@
...
 #>
 Param(
-  [string]$Arch,
-  [string]$Version,
-  [string]$VersionLabel,
-  [string]$UpgradeCode
+    [string]$Arch,
+    [string]$Version,
+    [string]$VersionLabel
+    # [string]$UpgradeCode
```
terakilobyte, line 18: There's a lot of formatting changes in this file (xml-tools). This is open to debate on whether we want to pass in an `$UpgradeCode`.

installer/msi/build-msi.ps1
```diff
@@ -6,40 +6,113 @@
...
 
 $ErrorActionPreference = 'Stop'
 
 $ProjectName = "Atlas SQL ODBC"
-$sourceDir = pwd
-$resourceDir = pwd
-$binDir = pwd
+$sourceDir = Get-Location
+$resourceDir = Get-Location
+$binDir = Get-Location
```
terakilobyte, line 26: On windows, `pwd` is an alias for `Get-Location`. Since this is a powershell script, it's better to use `Get-Location`

installer/msi/build-msi.ps1
```diff
@@ -6,40 +6,113 @@
...
+# $WixPath = "C:\Program Files (x86)\WiX Toolset v3.11\bin"
 $wixUiExt = "$WixPath\WixUIExtension.dll"
 
 # we currently only support x64, but we'll leave the 32 bit support here
 # in case we eventually decide to provide a 32-bit driver
+# upgradeCodes and productCodes are pre-generated for the next several releases 
 if ($Arch -eq "x64") {
-    $upgradeCode = "a4303fe6-c8ca-11ed-afa1-0242ac120002"
-} else {
+    switch ($VersionLabel) {
```
terakilobyte, line 37: and... I took this idea from the postgres installer. Not sure if we want to keep it. It definitely made testing a lot easier.

The pros of this are that we can see the guids easily and they are in source control, and we don't have to look back through logs to see what the codes were. We also aren't relying on the guid utility on the evergreen host to always produce the same guid with the same input (it should, but who knows).

The cons are eventually we'll need to come back and add more guids if this list isn't exhaustive enough.

As an aside, we aren't being fully accurate with this process. 1.0 -> 1.1 is a minor upgrade, but to support the use-case we want, we change the product code to make it a major upgrade.

logger/src/lib.rs
```diff
@@ -40,9 +41,30 @@ impl Logger {
...
+                        .document_dir()
+                        .map(|p| {
+                            p.join("MongoDB")
+                                .join("Atlas SQL ODBC")
+                                .join(version)
+                                .join("logs")
+                        })
+                        .unwrap_or_else(|| std::env::temp_dir());
+                    if !log_dir.exists() {
+                        std::fs::create_dir_all(&log_dir).unwrap();
```
terakilobyte, line 62: logging to the user's Documents location instead of where the driver is installed so we don't have to worry about updating our docs to show the user how to update the permissions on the directory, if they even can.

terakilobyte:
Hmm... the evergreen failure is due to `thread 'test_runner::resultset_tests' panicked at 'called `Result::unwrap()` on an `Err` value: DriverConnect("ERROR", "[MongoDB][API] The feature SQLSetConnectAttr is not implemented")'`. This is popping up elsewhere as well.

nbagnard (commented):
This is looking good!
I'd like to hear from @pmeredit about the product and upgrade code stored directly in the PS script since I find it hard to convince myself that I see the benefits except for testing manually.

installer/msi/BinaryFragment.wxs
```diff
@@ -1,29 +1,24 @@
...
+            </Component>
+        </DirectoryRef>
+        <ComponentGroup Id="base">
+            <ComponentRef Id="atsql.files" />
```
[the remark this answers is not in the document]
terakilobyte: Can do!

installer/msi/FeatureFragment.wxs
```diff
@@ -3,13 +3,9 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
   <Fragment>
     <FeatureGroup Id="fg_All">
-      <Feature Id="ProductAll"
-               Title="!(wix.ProductName)"
-               Description="!(wix.ProductName)"
-               Level="1"
-               ConfigurableDirectory="INSTALLLOCATION">
+      <Feature Id="ProductAll" Title="!(wix.ProductName)" Description="!(wix.ProductName)" Level="1" ConfigurableDirectory="INSTALLLOCATION">
```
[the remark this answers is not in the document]
terakilobyte: I'll use a different xml formatter that does this. There are quite a few to choose from!

logger/src/lib.rs
terakilobyte: It will write to the currently logged in user's documents directory, so not a shared directory among all users.

installer/msi/build-msi.ps1
pmeredit: Still disagree. `Get-Location` is super long and less clear what it is. Everyone knows `pwd`.

installer/msi/build-msi.ps1
pmeredit: It only matters for snapshot testing

logger/src/lib.rs
pmeredit: This is correct. It's based on the current user not the user that installed it.

installer/msi/BinaryFragment.wxs
pmeredit: In the future, let's reformat entire files in separate tickets. It's virtually impossible to tell what actually changed here.

installer/msi/BinaryFragment.wxs
```diff
@@ -1,29 +1,24 @@
...
+            <?define win64Flag = "yes"?>
+            <?else ?>
+            <?define win64Flag = "no"?>
+            <?endif ?>
+            <Component Id="atsql.files" Guid="*" Win64="$(var.win64Flag)">
+                <File Id="atsql.dll" Name="atsql.dll" Source="atsql.dll" />
+                <File Id="atsqls.dll" Name="atsqls.dll" Source="atsqls.dll" />
+                <RegistryKey ForceCreateOnInstall="yes" ForceDeleteOnUninstall="yes" Id="MongoDBKey" Key="SOFTWARE\MongoDB\[ProductName]" Root="HKLM" />
+                <RegistryValue Id="Version" Key="SOFTWARE\MongoDB\[ProductName]" Name="Version" Root="HKLM" Type="string" Value="[ProductVersion]" />
+                <RegistryValue Id="mongodb.reg.1" Root="HKLM" Key="SOFTWARE\ODBC\ODBCINST.INI\ODBC Drivers" Name="MongoDB Atlas SQL ODBC Driver" Type="string" Value="Installed" />
```
pmeredit, line 15:
```suggestion
                <RegistryValue Id="mongodb.reg.installed" Root="HKLM" Key="SOFTWARE\ODBC\ODBCINST.INI\ODBC Drivers" Name="MongoDB Atlas SQL ODBC Driver" Type="string" Value="Installed" />
```

installer/msi/BinaryFragment.wxs
```diff
@@ -1,29 +1,24 @@
...
+                <RegistryValue Id="mongodb.reg.2" Root="HKLM" Key="SOFTWARE\ODBC\ODBCINST.INI\MongoDB Atlas SQL ODBC Driver" Name="Driver" Type="string" Value="[#atsql.dll]" />
```
pmeredit, line 16:
```suggestion
                <RegistryValue Id="mongodb.reg.driver" Root="HKLM" Key="SOFTWARE\ODBC\ODBCINST.INI\MongoDB Atlas SQL ODBC Driver" Name="Driver" Type="string" Value="[#atsql.dll]" />
```

installer/msi/BinaryFragment.wxs
pmeredit, line 17:
```suggestion
                <RegistryValue Id="mongodb.reg.setup" Root="HKLM" Key="SOFTWARE\ODBC\ODBCINST.INI\MongoDB Atlas SQL ODBC Driver" Name="Setup" Type="string" Value="[#atsqls.dll]" />
```

pmeredit: Hah, now we don't have to use 8 characters for the name. Oh well.

installer/msi/build-msi.ps1
pmeredit: I'd prefer just keeping them in uuidgen. This file will grow unbounded like this. It doesn't matter too much either way, though.

Regenerating them is easy because we can just run the uuidgen command from the evergreen file.

pmeredit (approved):
LGTM modulo some minor nits. Don't worry about the formatting now, just in the future don't reformat an entire file that you are also modifying parts of ;)

installer/msi/BinaryFragment.wxs
pmeredit: These are minor suggestions that don't really matter. I just prefer names that have meaning versus 1,2,3

evergreen.yml
```diff
@@ -337,11 +337,7 @@ functions:
               MINOR_VERSION=$(echo "$RELEASE_VERSION" | sed 's|\([0-9]\+[.][0-9]\+\)[.][0-9]\+|\1|')
               VERSION_LABEL="$RELEASE_VERSION"
           fi
-          # This UUID is used for the upgrade code and should be consistent across minor version_id
-          # bug patches, that is why it is seeded with the minor version. Being consistent across
-          # bug patches but different across minor versions allows for users to have multiple
-          # minor versions of the driver installed side by side on Windows.
-          UUID=$(uuidgen -m -N "mongoodbc_no_conflicts_$MINOR_VERSION" -n @oid)
+
```
terakilobyte, line 340: This is all removed now since we aren't providing side-by-side installations.

installer/msi/FeatureFragment.wxs
```diff
@@ -1,16 +1,16 @@
...
-    </FeatureGroup>
-  </Fragment>
-</Wix>
+    <Fragment>
+        <FeatureGroup Id="fg_All">
+            <Feature Id="ProductAll"
+                     Title="!(wix.ProductName)"
+                     Description="!(wix.ProductName)"
+                     Level="1"
+                     ConfigurableDirectory="PROJECT">
```
terakilobyte, line 10: Since we aren't support side-by-side, I removed the INSTALLLOCATION id

logger/src/lib.rs
```diff
@@ -22,10 +24,7 @@ pub struct Logger {
 
 impl Logger {
     /// Initializes the logger with the given path as its root. The logger will create a logs folder
-    /// two levels up from the given path and write its logs there. For example, if the given path is
-    /// `C:\\Program Files\\mongo_odbc\\1.1\\bin\\atsql.dll`, the logger will write its logs to
-    /// `C:\\Program Files\\mongo_odbc\\1.1\\logs\\mongo_odbc.log`. `std::path::Path` is used to assemble
-    /// the paths so this is OS safe.
+    /// in the user's documents directory, such as User/Documents/MongoDB/Atlas ODBC Driver/1.1/logs
```
nbagnard, line 27:
```suggestion
    /// in the user's documents directory, such as User/Documents/MongoDB/Atlas SQL ODBC Driver/1.1/logs
```

nbagnard (approved):
LGTM!

terakilobyte:
evergreen merge

PR #127, opened 2023-04-28
SQL-1413: Copy signed msi to local folder instead of unsigned

nbagnard:
I also changed hard-coded variant names to use ${build_variant} for copying to/from the mciuploads bucket because we had a bit of both and this could potentially become a problem if/when we change variant.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -513,7 +513,7 @@ functions:
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        remote_file: mongosql-odbc-driver/artifacts/${version_id}/${build_variant}/release/mongoodbc-signed.msi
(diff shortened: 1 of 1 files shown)


nbagnard:
> I also changed hard-coded variant names to use ${build_variant} for copying to/from the mciuploads bucket because we had a bit of both and this could potentially become a problem if/when we change variant.

Actually, this won't work because the task are done via the "release" variant, so we need to hard code the directory :(

nbagnard:
And yes, it looks like signing renames the file
See here : https://spruce.mongodb.com/task/mongosql_odbc_driver_windows_64_windows_sign_644a94f90ae606971d42f874_23_04_27_15_30_01/files?execution=0

jchemburkar (approved):
LGTM!

nbagnard:
evergreen merge

PR #128, opened 2023-05-01
SQL-1369: add basic developer info in DEVELOPER.md

jchemburkar:


DEVELOPER.md
```diff
@@ -0,0 +1,43 @@
...
+
+### To run unit tests
+
+Similar to building, standard cargo commands can be used here:
+
+```
+cargo test unit
+```
+
+### Other types of tests
```
jchemburkar, line 34: note: I considered transcribing steps from evergreen.yml here, but given that would require updating this doc anytime they update (which seems not totally uncommon), I left it like this for now, but would reconsider if it may be worthwhile

DEVELOPER.md
```diff
@@ -0,0 +1,43 @@
...
+The resulting build files will be output to the `target` directory in the `debug` or `release` folders respectively.
+
+## Running Tests
+
+### To run unit tests
+
+Similar to building, standard cargo commands can be used here:
+
+```
+cargo test unit
```
terakilobyte, line 31: The tests should also be run with `--features` needed

DEVELOPER.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+### Other types of tests
+The other tests that are run are integration and result set tests. These involve more setup, and that setup is largely operating system specific. For more information on how to run these, consider checking [evergreen.yml](evergreen.yml), and following the steps taken for the given operating system-task (i.e. `macos-integration-test` or `windows-result-set-test`)
```
terakilobyte, line 35: One potential issue here is that `go` is set up differently on a local machine, and on windows especially it's possible to run the integration tests without cygwin by using wsl to run adf, so you only have to worry about the integration tests.

terakilobyte (changes requested):
A few comments/clarifications I think we should make.

DEVELOPER.md
jchemburkar: 👍 just to double check, does this 1. only apply to mac? and 2. does it apply to unit tests or just to the build (which I should update), just looking at evergreen file I see the features in `compile macos <debug/release>` but not `run macos unit tests`

DEVELOPER.md
jchemburkar: That makes sense, and thanks for calling out windows (I have only had the chance to run on windows hosts, which === evergreen). I can just hedge the language here so it is just nudging in the right direction, or add more specific instructions for windows/mac (I would probably need a little help with local windows in that case). What do you think is the best way to proceed?

DEVELOPER.md
terakilobyte: I believe it only applies to mac. As a general rule, it should be done for all tests. Ones that may be impacted that come to mind are the data tests which may very much depend on the length of the buffer returned.

DEVELOPER.md
terakilobyte: I can contribute what I do on Windows if you want. It's actually pretty easy. I run WSL and run adf there (so linux setup), then on windows it "just connects" when I run the tests. The only thing is to make sure environment variables are properly set.

DEVELOPER.md
```diff
@@ -0,0 +1,114 @@
...
+```
+cargo build --features odbc-sys/iodbc,cstr/utf32 --profile=release-with-debug
+```
+And for an optimized build with debugging info (windows, linux)
+```
+cargo build --profile=release-with-debug
+```
+The resulting build files will be output to the `target` directory in the `debug` or `release` folders respectively.
+
+## Setting up the driver manager on MacOS
```
jchemburkar, line 24: This mirrors the windows version in the main `README.md` pretty closely, but given the relative state of both I thought this made sense in the developer readme for now. Happy to move it though if you think it belongs there instead!

DEVELOPER.md
```diff
@@ -0,0 +1,114 @@
...
+cd "$iODBC_dir"
+./configure --prefix="$INSTALLED_ODBC_PATH"
+make 
+make install
+```
+Next, using the steps above, compile the release version with debug info. Additionally, set the library paths:
+```
+export LIBRARY_PATH="$INSTALLED_ODBC_PATH/lib"
+export LD_LIBRARY_PATH="$INSTALLED_ODBC_PATH/lib"
+export DYLD_LIBRARY_PATH="$INSTALLED_ODBC_PATH/lib"
```
jchemburkar, line 50: also wanted to note that our conclusion from earlier was correct, and adding this one line to my personal setup (after following the rest of these steps) fixed the problem.

terakilobyte (approved):
LGTM. I made on comment... we can totally remove the custom profile from here. I'll make a ticket to remove the profile from evergreen and cargo.

jchemburkar:
evergreen merge

PR #129, opened 2023-05-04
SQL-1285: Remove ANSI endpoint

nbagnard:
Where to find the list of W functions => https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/unicode-function-arguments

### DIFF
--- a/odbc/src/api/diag.rs
+++ b/odbc/src/api/diag.rs
@@ -7,23 +7,8 @@ use crate::{
 };
 use cstr::WideChar;
 use odbc_sys::Pointer;
-use odbc_sys::{Char, Integer, SmallInt, SqlReturn};
+use odbc_sys::{Integer, SmallInt, SqlReturn};
 use std::ptr::copy_nonoverlapping;
-
-///
-/// set_sql_state writes the given sql state to the [`output_ptr`].
-///
-/// # Safety
-/// This writes to a raw C-pointer
-///
-pub unsafe fn set_sql_state(sql_state: &str, output_ptr: *mut Char) {
-    if output_ptr.is_null() {
-        return;
-    }
-    let sql_state = &format!("{sql_state}\0");
-    let state_u8 = sql_state.bytes().collect::<Vec<u8>>();
-    copy_nonoverlapping(state_u8.as_ptr(), output_ptr, 6);
-}
 
 ///
 /// set_sql_statew writes the given sql state to the [`output_ptr`].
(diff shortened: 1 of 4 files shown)


terakilobyte (approved):
Changes LGTM. Looks like clippy and rustfmt are being tyrants.

odbc/src/api/diag.rs
```diff
@@ -40,34 +25,6 @@ pub unsafe fn set_sql_statew(sql_state: &str, output_ptr: *mut WideChar) {
...
 }
 
-///
-/// get_diag_rec copies the given ODBC error's diagnostic information
-/// into the provided pointers.
-///
-/// # Safety
-/// This writes to multiple raw C-pointers
-///
-pub unsafe fn get_diag_rec(
```
jchemburkar, line 50: This looks good! I think we can simplify the `get_diag_field` function now as well to remove the `is_wstring` flag

odbc/src/api/functions.rs
```diff
@@ -663,77 +602,6 @@ pub unsafe extern "C" fn SQLColumnPrivilegesW(
...
 
-///
-/// [`SQLColumns`]: https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/SQLColumns-function
-///
-/// # Safety
-/// Because this is a C-interface, this is necessarily unsafe
-///
-#[named]
-#[no_mangle]
-pub unsafe extern "C" fn SQLColumns(
```
jchemburkar, line 674: For these, and the other functions that _were_ implemented and now removed, should we add them to the list of `Unsupported Operators` in the README?

jchemburkar (approved):
LGTM! Just had a couple small thoughts

odbc/src/api/functions.rs
nbagnard: Good call!
I looked at the README and I don't really see a need to go into this level of details. I'll add a line stating that the driver does not support ANSI endpoints, that should be enough, no need to list all the ANSI functions.
But, I still need to also clean-up the list of unsupported functions.

nbagnard:
@jchemburkar, @terakilobyte : I updated the README, do you mind giving it one last look? I re-requested your review to make sure that it makes sense to you. I think it's simpler to state that we don't support ANSI functions and then list the ones that we don't support at all. But let me know if that is not working for you. I also cleaned the README and removed the 32-bit steps because we don't support 32-bit.
I left the registry update scripts because if people are coming to this page, they won't have access to the installer.

odbc/src/api/diag.rs
```diff
@@ -150,41 +106,25 @@ pub unsafe fn get_diag_field(
...
-                            buffer_length as usize,
-                            string_length_ptr,
-                        ),
-                    },
+                    DiagType::SQL_DIAG_SQLSTATE => i16_len::set_output_wstring_as_bytes(
+                        error.get_sql_state(),
+                        diag_info_ptr,
+                        buffer_length as usize,
+                        string_length_ptr,
+                    ),
```
terakilobyte, line 114: Not related to this PR at all, but I'd really like refactoring all the string writing to cstr. I could create a ticket. WDYT?

terakilobyte (approved):
LGTM. Left one small "correction", no need to re-tag if you accept it!

odbc/src/api/diag.rs
nbagnard: > Not related to this PR at all, but I'd really like refactoring all the string writing to cstr. I could create a ticket. WDYT?

@terakilobyte Please do :)

jchemburkar (approved):
The updates to the README look good to me!

nbagnard:
evergreen merge

PR #132, opened 2023-05-15
SQL-1285: SQLGetEnvAttr and SQLSetEnvAttr don't have Ansi/Wide flavors

nbagnard:
Working patch : https://spruce.mongodb.com/version/646241a33e8e8628c178bacb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/odbc/src/api/env_attr_tests.rs
+++ b/odbc/src/api/env_attr_tests.rs
@@ -1,9 +1,9 @@
-use crate::{map, SQLSetEnvAttr};
 use crate::{
     api::definitions::*,
     handles::definitions::{Env, EnvState, MongoHandle},
     SQLGetDiagRecW, SQLGetEnvAttr,
 };
+use crate::{map, SQLSetEnvAttr};
 use odbc_sys::{HEnv, HandleType, Integer, Pointer, SqlReturn};
 use std::{collections::BTreeMap, ffi::c_void, mem::size_of};
 
--- a/odbc/src/api/env_attr_tests.rs
+++ b/odbc/src/api/env_attr_tests.rs
@@ -1,8 +1,8 @@
-use crate::map;
+use crate::{map, SQLSetEnvAttr};
 use crate::{
     api::definitions::*,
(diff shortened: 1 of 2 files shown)


pmeredit (commented):
The docs: https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/SQLGetEnvAttr-function?view=sql-server-ver16

claim that there is a SQLGetEnvAttrW... I'm confused now. In our case we never return string values for anything we support, so it shouldn't matter, and the DM should automatically call the ANSI version when the wide version doesn't exist.

terakilobyte (approved):
This LGTM. To Patrick's comment and for my own edification, I thought only functions that accept or return a string should have the `W` decorator. Since `SQLGetEnvAttr` doesn't return a string but well defined attributes, it shouldn't have a `W` decorator? Should we implement [`SQLConnectW`](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/function-mapping-in-the-driver-manager?view=sql-server-ver16#:~:text=The%20Driver%20Manager%20recognizes%20a%20driver%20as%20a%20Unicode%20driver%20if%20SQLConnectW%20is%20supported%20by%20the%20driver.)

nbagnard:
@pmeredit @terakilobyte I agree that it would make sense to have a W flavor, however, if you look at [this page](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/unicode-function-arguments?view=sql-server-ver16), SQLGetEnvAttr is not listed. And, if you check the headers for [Windows](https://github.com/microsoft/ODBC-Specification/blob/master/Windows/inc/sqlucode.h) and Unix, the function SQLGetEnvAttrW also does not exist.

pmeredit (approved):
LGTM

nbagnard:
> the docs contradict themselves, but I'm fine with this.

Yes, I know.

nbagnard:
evergreen merge

PR #133, opened 2023-05-18
SQL-1434: Support setting SQL_ATTR_CURRENT_CATALOG

terakilobyte:
Turns out we supported getting this attribute in SQLGetConnectAttrW but never implemented setting it via SQLSetConnectAttrW.

### DIFF
--- a/core/src/conn.rs
+++ b/core/src/conn.rs
@@ -6,24 +6,27 @@ use mongodb::sync::Client;
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 
+#[derive(Debug, Default, Clone)]
+pub struct ConnectionAttributes {
+    // SQL_ATTR_CURRENT_CATALOG: the current catalog/database
+    // for this Connection.
+    pub current_catalog: Option<String>,
+    // SQL_ATTR_LOGIN_TIMEOUT: SQLUINTEGER, timeout in seconds
+    // to wait for a login request to complete.
+    pub login_timeout: Option<u32>,
+    // SQL_ATTR_CONNECTION_TIMEOUT: SQLUINTER, timeout in seconds
+    // to wait for any operation on a connection to timeout (other than
+    // initial login).
+    pub connection_timeout: Option<u32>,
+}
 #[derive(Debug)]
 pub struct MongoConnection {
     /// The mongo DB client
     pub client: Client,
-    /// The current database set for this client.
-    /// All new queries will be done on this DB.
-    /// We stick with mongo terminology here and ODBC
-    /// terminology in the odbc wrappers, hence
-    /// current_db here and current_catalog in the
-    /// odbc/handles code.
-    pub current_db: Option<String>,
-    /// Number of seconds to wait for any request on the connection to complete before returning to
-    /// the application.
-    /// Comes from SQL_ATTR_CONNECTION_TIMEOUT if set. Used any time there is a time out in a
-    /// situation not associated with query execution or login.
-    pub operation_timeout: Option<Duration>,
     /// The UuidRepresentation to use for this connection.
     pub uuid_repr: Option<UuidRepresentation>,
+    // all the possible Connection settings
+    pub attributes: ConnectionAttributes,
 }
 
 impl MongoConnection {
@@ -39,20 +42,18 @@ impl MongoConnection {
     /// setting specified in the uri if any.
     pub fn connect(
         mut user_options: UserOptions,
-        current_db: Option<String>,
-        operation_timeout: Option<u32>,
-        login_timeout: Option<u32>,
+        connection_attributes: ConnectionAttributes,
     ) -> Result<Self> {
-        user_options.client_options.connect_timeout =
-            login_timeout.map(|to| Duration::new(to as u64, 0));
+        user_options.client_options.connect_timeout = connection_attributes
+            .login_timeout
+            .map(|to| Duration::new(to as u64, 0));
         let client = Client::with_options(user_options.client_options)
             .map_err(Error::InvalidClientOptions)?;
         let uuid_repr = user_options.uuid_representation;
         let connection = MongoConnection {
             client,
-            current_db,
-            operation_timeout: operation_timeout.map(|to| Duration::new(to as u64, 0)),
             uuid_repr,
+            attributes: connection_attributes,
         };
         // Verify that the connection is working and the user has access to the default DB
         MongoQuery::execute(&connection, None, "select 1")?;
--- a/core/src/conn.rs
+++ b/core/src/conn.rs
@@ -26,7 +26,8 @@ pub struct MongoConnection {
     /// The UuidRepresentation to use for this connection.
     pub uuid_repr: Option<UuidRepresentation>,
     // all the possible Connection settings
-    pub attributes: ConnectionAttributes,
+    // this is an option for ergonomics only. In reality, this will always be Some
+    pub attributes: Option<ConnectionAttributes>,
 }
 
 impl MongoConnection {
@@ -42,9 +43,11 @@ impl MongoConnection {
     /// setting specified in the uri if any.
     pub fn connect(
         mut user_options: UserOptions,
-        connection_attributes: ConnectionAttributes,
+        connection_attributes: Option<ConnectionAttributes>,
     ) -> Result<Self> {
         user_options.client_options.connect_timeout = connection_attributes
+            .as_ref()
+            .unwrap()
             .login_timeout
             .map(|to| Duration::new(to as u64, 0));
         let client = Client::with_options(user_options.client_options)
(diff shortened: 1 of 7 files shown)


bucaojit (approved):
LGTM!  Thanks for getting to this right away 🥇

pmeredit (approved):
Great! Can't believe we missed this before :P

pmeredit:
Actually hold on a second.

pmeredit (changes requested):
You're going to need to set the current_catalog in the `MongoConnection` struct for this to work right, which exists inside of the `Connection` handle.

In fact, it'd make more sense to remove the current_catalog from the attributes entirely and only report the value from MongoConnection.

HOWEVER, it might be allowable to set the current_catalog before CONNECTING the Connection. In which case we do need to preserve the attribute (and set it in both places if both places exist when Set is called, and also populate the field in the MongoConnection from the attributes when connect is called)

Sorry for this, this seems to be a bug in what we were already doing.

terakilobyte:
@pmeredit This is ready to look at again. The current tests cover the surface area of the changes pretty well (I had to go through with a debugger to remove a threadlock issue), but I feel like maybe I need to add more.

core/src/conn.rs
```diff
@@ -6,24 +6,27 @@ use mongodb::sync::Client;
...
+    // for this Connection.
+    pub current_catalog: Option<String>,
+    // SQL_ATTR_LOGIN_TIMEOUT: SQLUINTEGER, timeout in seconds
+    // to wait for a login request to complete.
+    pub login_timeout: Option<u32>,
+    // SQL_ATTR_CONNECTION_TIMEOUT: SQLUINTER, timeout in seconds
+    // to wait for any operation on a connection to timeout (other than
+    // initial login).
+    pub connection_timeout: Option<u32>,
+}
```
terakilobyte, line 21: This definition was relocated here and is shared between core and odbc

core/src/conn.rs
```diff
@@ -6,24 +6,27 @@ use mongodb::sync::Client;
...
-    /// We stick with mongo terminology here and ODBC
-    /// terminology in the odbc wrappers, hence
-    /// current_db here and current_catalog in the
-    /// odbc/handles code.
-    pub current_db: Option<String>,
-    /// Number of seconds to wait for any request on the connection to complete before returning to
-    /// the application.
-    /// Comes from SQL_ATTR_CONNECTION_TIMEOUT if set. Used any time there is a time out in a
-    /// situation not associated with query execution or login.
-    pub operation_timeout: Option<Duration>,
```
terakilobyte, line 24: All removed in favor of attributes

odbc/src/api/functions.rs
```diff
@@ -868,22 +868,33 @@ fn sql_driver_connect(conn: &Connection, odbc_uri_string: &str) -> Result<MongoC
...
+            .as_deref()
+            .map(|s| s.to_string());
+        database = if catalog_database.is_some() {
+            catalog_database
+        } else {
+            odbc_uri.remove(&["database"])
+        };
+    }
+    let mut attrs = conn.attributes.write().unwrap().take().unwrap();
+    attrs.current_catalog = database;
```
terakilobyte, line 891: This feels a little verbose, but makes the Rust compiler happy. This thankfully is also the only place that I take a read lock immediately followed by a write lock on the same RwLock, so the only place this extra scope is introduced. I'm thinking about taking one write lock here for efficiency. Thoughts?

odbc/src/api/functions.rs
```diff
@@ -1321,25 +1351,35 @@ unsafe fn sql_get_connect_attrw_helper(
...
+
+        let has_connection = conn.has_mongo_connection();
+        // unfortunate clones here, but the types in a ConnectionAttributes struct are small
+        let attributes = if has_connection {
+            mongo_connection.as_ref().unwrap().attributes.clone()
+        } else {
+            connection_attributes.as_ref().unwrap().clone()
+        };
 
+        // Only attributes that are Either or After are synced to the attributes in the mongo_connection
```
terakilobyte, line 1365: oops, dead comment and will be removed next commit!

odbc/src/api/functions.rs
```diff
@@ -2761,9 +2804,49 @@ unsafe fn set_connect_attrw_helper(
     let sql_return = {
         let conn = must_be_valid!((*conn_handle).as_connection());
 
+        let has_connection = conn.has_mongo_connection();
+
+        // Attributes could be set before or after the mongo_connection is established, so we
+        // have to check and write to the proper place. Keep in mind the conn.attributes Option<RwLock<ConnectionAttributes>> is TAKEN
+        // when a mongo_connection is established, so writing to both isn't safe
```
terakilobyte, line 2811: Dead comments that will be removed. We discussed offline that we won't bother validating if the dm is passing attributes to us when it should or shouldn't.

odbc/src/api/functions.rs
```diff
@@ -2761,9 +2804,49 @@ unsafe fn set_connect_attrw_helper(
...
+                        .unwrap()
+                        .attributes
+                        .current_catalog = Some(db);
+                } else {
+                    conn.attributes
+                        .write()
+                        .unwrap()
+                        .as_mut()
+                        .unwrap()
+                        .current_catalog = Some(db);
```
terakilobyte, line 2848: Thinking about ripping this out into a function.

terakilobyte: Actually, this would be a bit harder without a wrapper to generalize it.

pmeredit: It's fine.

odbc/src/api/functions.rs
```diff
@@ -1211,6 +1213,20 @@ fn sql_free_handle(handle_type: HandleType, handle: *mut MongoHandle) -> Result<
                     .as_connection()
                     .ok_or(ODBCError::InvalidHandleType(HANDLE_MUST_BE_CONN_ERROR))?
             };
+            // We must take the connection attributes in the mongo_connection and put them back
+            // in the connection attributes for the conn
+            if conn.has_mongo_connection() {
```
pmeredit, line 1218: You shouldn't need to copy these just to drop them.

pmeredit (changes requested):

terakilobyte:
@pmeredit I hit this with a big test. The good news is, it covers all of the movement back and forth between potential locations the attributes could be in depending on the connection state.

odbc/src/api/functions.rs
```diff
@@ -836,6 +836,19 @@ pub unsafe extern "C" fn SQLDisconnect(connection_handle: HDbc) -> SqlReturn {
         || {
             let conn_handle = MongoHandleRef::from(connection_handle);
             let conn = must_be_valid!((*conn_handle).as_connection());
+
+            // prior to dropping the mongo_connection, we need to move it's ConnectionAttributes
+            // to the conn's ConnectionAttributes - the inverse of what we do when a mongo_connection
```
pmeredit, line 841:
```suggestion
            // to the conn_handle's ConnectionAttributes - the inverse of what we do when a mongo_connection
```

odbc/src/api/functions.rs
```diff
@@ -1211,6 +1226,20 @@ fn sql_free_handle(handle_type: HandleType, handle: *mut MongoHandle) -> Result<
...
+                let mongo_connection_attributes = conn
+                    .mongo_connection
+                    .read()
+                    .unwrap()
+                    .as_ref()
+                    .unwrap()
+                    .attributes
+                    .clone();
+                let mut connection_attributes = conn.attributes.write().unwrap();
+                *connection_attributes = Some(mongo_connection_attributes);
```
pmeredit, line 1241:
```suggestion
                let mongo_connection_attributes = conn
                    .mongo_connection
                    .read()
                    .unwrap()
                    .as_ref()
                    .unwrap()
                    .attributes
                    .clone();
                let mut connection_attributes = conn.attributes.write().unwrap();
                *connection_attributes = Some(mongo_connection_attributes);
```

This clone isn't necessary.

odbc/src/api/functions.rs
```diff
@@ -1211,6 +1226,20 @@ fn sql_free_handle(handle_type: HandleType, handle: *mut MongoHandle) -> Result<
                     .as_connection()
                     .ok_or(ODBCError::InvalidHandleType(HANDLE_MUST_BE_CONN_ERROR))?
             };
+            // We must take the connection attributes in the mongo_connection and put them back
+            // in the connection attributes for the conn
+            if conn.has_mongo_connection() {
```
pmeredit, line 1231:
```suggestion
```

pmeredit: Sorry this is two diffs because github doesn't seem to let you edit around your own comment lol

pmeredit (changes requested):
Feel free to merge after making these changes without retagging me

odbc/src/api/functions.rs
terakilobyte: this is dead code I thought I had removed. It's gone now!

terakilobyte:
evergreen merge

PR #135, opened 2023-05-24
SQL-1439: clear diagnostics between odbc function calls

terakilobyte:
The only functions that we implement that should not clear diagnostics are SQLGetDiagRec and SQLGetField (we don't implement SQLError).

I also set SQLAllocHandle and SQLFreeHandle to keep diagnostics to avoid any potential for calling unallocated memory, and freeing the handle is probably the ultimate way to clear diagnostics anyway.

Most of the changes are pretty mundane as I renamed the `panic_safe_exec` and that propagated everywhere.

### DIFF
--- a/odbc/src/api/data_tests.rs
+++ b/odbc/src/api/data_tests.rs
@@ -655,7 +655,7 @@ mod unit {
                             .unwrap()
                             .errors
                             .read()
-                            .unwrap()[1]
+                            .unwrap()[0]
                     ),
                 );
                 guid_val_test(
--- a/odbc/src/api/functions.rs
+++ b/odbc/src/api/functions.rs
@@ -111,7 +111,8 @@ macro_rules! odbc_unwrap {
     }};
 }
 
-// panic_safe_exec executes `function` such that any panics do not crash the runtime.
+// panic_safe_exec_clear_diagnostics executes `function` such that any panics do not crash the runtime,
+// while clearing any diagnostics in the $handle's error vec.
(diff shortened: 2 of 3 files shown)


odbc/src/api/data_tests.rs
```diff
@@ -655,7 +655,7 @@ mod unit {
                             .unwrap()
                             .errors
                             .read()
-                            .unwrap()[1]
+                            .unwrap()[0]
```
terakilobyte, line 658: After clearing diagnostics, this needed to be changed!

odbc/src/api/functions.rs
```diff
@@ -171,19 +172,80 @@ macro_rules! panic_safe_exec {
         };
     }};
 }
-pub(crate) use panic_safe_exec;
+pub(crate) use panic_safe_exec_clear_diagnostics;
+
+// panic_safe_exec_keep_diagnostics executes `function` such that any panics do not crash the runtime.
+// If a panic occurs during execution, the panic is caught and turned into a String.
+// The panic message is added to the diagnostics of `handle` and SqlReturn::ERROR returned.
+macro_rules! panic_safe_exec_keep_diagnostics {
```
terakilobyte, line 180: I duplicated the macro and just renamed it to make it clear what it was doing rather than adding an argument.

pmeredit (approved):
LGTM!

bucaojit (approved):
LGTM!  Looks good with the macro update 👍

terakilobyte:
evergreen merge

PR #136, opened 2023-05-25
SQL-1440: filter out admin and empty strings

terakilobyte:
I've opened a ticket with ADF, and commented the ticket number above the relevant lines I added.

### DIFF
--- a/core/src/collections.rs
+++ b/core/src/collections.rs
@@ -114,6 +114,8 @@ impl MongoCollections {
             )
             .unwrap()
             .iter()
+            // MHOUSE-7119 - admin database and empty strings are showing in list_database_names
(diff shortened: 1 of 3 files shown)


bucaojit (changes requested):
There's another list_database_names in [fields.rs](https://github.com/mongodb/mongo-odbc-driver/blob/d711fc0b75e81d3e28c573adc6b8be462a44e1eb/core/src/fields.rs#L482) that this filter should be added to.

bucaojit (approved):
LGTM!

As this will possibly be fixed by ADF soon, we will not worry about adding this as server-side filtering.

EthanHardyMongo (approved):
LGTM!

terakilobyte:
evergreen merge

PR #138, opened 2023-06-05
SQL-1334: Clean up ODBC evergreen file

EthanHardyMongo:
Basically, I took all the same tasks that were made for each variant and merged them into a more general task (e.g., `windows_unit_test`, `ubuntu_unit_test`, etc. merged into `unit_test`).

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -1362,44 +1362,40 @@ tasks:
   - name: compile
     depends_on:
     commands:
+      - func: "install iODBC"
+        variants: [macos, macos-arm]
       - func: "install rust toolchain"
+        variants: [ubuntu2204, windows-64]
       - func: "install unix odbc"
         variants: [ubuntu2204]
       - func: "set packages version"
       - func: "check packages version"
       - func: "compile release"
+        variants: [ubuntu2204, windows-64]
+      - func: "compile macos release"
+        variants: [macos, macos-arm]
       - func: "build msi"
         variants: [windows-64]
       - func: "tar linux artifacts"
         variants: [ubuntu2204]
-      - func: "mciuploads release artifacts"
-
-  - name: compile-macos
-    depends_on:
-    commands:
-      - func: "install iODBC"
-      - func: "set packages version"
-      - func: "check packages version"
-      - func: "compile macos release"
       - func: "build dmg"
+        variants: [macos, macos-arm]
       - func: "mciuploads release artifacts"
 
   - name: compile-debug
     commands:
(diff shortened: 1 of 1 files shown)


evergreen.yml
```diff
@@ -1510,26 +1496,14 @@ tasks:
...
-      - name: ubuntu-result-set-test
-        variant: "ubuntu2204"
-      - name: asan-unit-test
+      - name: unit-test
+        variant: "*"
+      - name: integration-test
         variant: "*"
-      - name: asan-integration-test
+      - name: result-set-test
         variant: "*"
```
EthanHardyMongo, line 1504: One thing I thought I should mention is that now the `snapshot` task depends on all of the variants passing the `unit_test` and `integration_test` since the `asan` tests need to pass for all variants and I can't break the tests apart if they are in a more general test.

evergreen.yml
```diff
@@ -1510,26 +1496,14 @@ tasks:
...
-        variant: "ubuntu2204"
-      - name: ubuntu-result-set-test
-        variant: "ubuntu2204"
-      - name: asan-unit-test
+      - name: unit-test
         variant: "*"
-      - name: asan-integration-test
+      - name: integration-test
         variant: "*"
-      - name: windows-sign
```
EthanHardyMongo, line 1529: Side note: Is there a reason that the `asan-unit-test` and `asan-integration-test` needed to use all variants?

bucaojit: That may have been a small oversight since there's only one variant that runs the asan tests.

evergreen.yml
```diff
@@ -1362,143 +1362,129 @@ tasks:
...
     commands:
+      - func: "install iODBC"
+        variants: [macos, macos-arm]
       - func: "install rust toolchain"
+        variants: [ubuntu2204, windows-64]
       - func: "install unix odbc"
         variants: [ubuntu2204]
       - func: "set packages version"
       - func: "check packages version"
       - func: "compile debug"
```
bucaojit, line 1395:
```suggestion
      - func: "compile ubuntu and win debug"
```
Same as above, if you think it would help with clarifying the task

bucaojit (changes requested):
Nice work, this makes the evergreen file much cleaner. 

I had a few suggestions.

pmeredit (approved):
LGTM!

bucaojit (approved):
Nice work.  LGTM!

EthanHardyMongo:
evergreen merge

PR #140, opened 2023-06-16
update to use tags

terakilobyte:
dependencies correctly show up under snapshot, so should also show up for sign

https://spruce.mongodb.com/task/mongosql_odbc_driver_release_snapshot_patch_28437756461ac8bb86261d730efaf90ccfe4565d_648cba31c9ec444d557450c9_23_06_16_19_39_33/logs?execution=0

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -1485,13 +1485,13 @@ tasks:
       - name: rustfmt
         variant: "*"
       - name: unit-test
-        variants: [windows-64, ubuntu2204]
+        variant: ".release-variant"
       - name: integration-test
-        variants: [windows-64, ubuntu2204]
+        variant: ".release-variant"
       - name: result-set-test
-        variants: [windows-64, ubuntu2204]
+        variant: ".release-variant"
       - name: sign
-        variants: [windows-64, ubuntu2204]
+        variant: ".release-variant"
     commands:
       - func: "upload release"
       - func: "upload debug"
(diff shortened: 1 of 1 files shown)


nbagnard (changes requested):
What about the remaining `variants: [xx,xx]`?
Does it run everything under a task (or tries to) when it can't understand the prop?
I think it'd be safer to clean up these as well.

EthanHardyMongo:
> What about the remaining `variants: [xx,xx]`?
> Does it run everything under a task (or tries to) when it can't understand the prop?
> I think it'd be safer to clean up these as well.

we found out that `variants: [xx,xx]` only causes issues when it is under the `depends_on:`. It works when under `commands:`

nbagnard:
> > What about the remaining `variants: [xx,xx]`?
> > Does it run everything under a task (or tries to) when it can't understand the prop?
> > I think it'd be safer to clean up these as well.
> 
> we found out that `variants: [xx,xx]` only causes issues when it is under the `depends_on:`. It works when under `commands:`

Ohhh , what a mess. I'm still not sure we want a mix of both and open the door to potential mistakes again tough.

terakilobyte:
@nbagnard Yeah it's tough. Also, there ended up being a bug filed because something they advertised as works, does in fact not work.

EthanHardyMongo (approved):
LGTM! Regarding the remaining `variants: [x,y]`, maybe we can just wait until they fix the space separated list feature (i.e., `variant: "A B C"`) that currently doesn't work. Then we can quickly change the rest of the remaining `variants: [x,y]`. Because If we try to change it now, I think we'll have to make tags which could end up making the file less readable.

terakilobyte:
merging pre-lgtm (lgtm received in slack)

terakilobyte:
evergreen merge

nbagnard (approved):

PR #141, opened 2023-06-21
SQL-1448: Update signing tasks to use the new Garasign platform - ODBC driver

EthanHardyMongo:
This ticket covers replacing uses of Notary Service with Garasign. In this ticket, I created a few new evergreen variables: `sql_engines_artifactory_username`, `sql_engines_artifactory_auth_token`, `release_tools_container_registry`, `odbc_garasign_username`, `odbc_garasign_password`, `garasign_jsign_image`, `garasign_gpg_image`. 

Two things I'm wondering about:
1. Despite the sign task for Windows running successfully, I see this error in the logs: `ERROR StatusLogger Log4j2 could not find a logging implementation. Please add log4j-core to the classpath. Using SimpleLogger to log to the console...`. I'm not sure if this matters.
2. Should I create a signing verification step like in the tableau driver? There's nothing like this already in the `evergreen.yml`, so I'm not sure how necessary it is to do.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -201,12 +201,12 @@ functions:
         working_dir: mongosql-odbc-driver/installer/msi
         script: |
           podman run \
-          -e GRS_CONFIG_USER1_USERNAME=${garasign_username} \
-          -e GRS_CONFIG_USER1_PASSWORD=${garasign_password} \
-          --rm \
-          -v $(pwd):$(pwd) -w $(pwd) \
-          ${garasign_jsign_image} \
-          /bin/bash -c "jsign -a mongo-authenticode-2021 --replace --tsaurl http://timestamp.digicert.com -d SHA-256 mongoodbc.msi"
+            -e GRS_CONFIG_USER1_USERNAME=${garasign_username} \
+            -e GRS_CONFIG_USER1_PASSWORD=${garasign_password} \
+            --rm \
+            -v $(pwd):$(pwd) -w $(pwd) \
+            ${garasign_jsign_image} \
+            /bin/bash -c "jsign -a mongo-authenticode-2021 --replace --tsaurl http://timestamp.digicert.com -d SHA-256 mongoodbc.msi"
 
   "sign ubuntu":
     - command: shell.exec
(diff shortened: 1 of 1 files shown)


evergreen.yml
```diff
@@ -1574,7 +1574,7 @@ buildvariants:
       - name: windows-test-integration-group
       - name: windows-test-result-set-group
       - name: sign
-        run_on: linux-64-amzn-build
+        run_on: ubuntu2204-large
```
EthanHardyMongo, line 1577: I added this because `windows-64-vs2019-large` doesn't have docker or Podman; however, I know that `ubuntu2204-large` has docker.

nbagnard: Sounds good to me. We absolutely don't have to use a Windows host to sign on Windows, you are 100% correct 👍

nbagnard (approved):
LGTM!

evergreen.yml
```diff
@@ -506,7 +506,7 @@ functions:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
         remote_file: mongosql-odbc-driver/artifacts/${version_id}/windows-64/release/mongoodbc-signed.msi
-        local_file: mongosql-odbc-driver/installer/msi/mongoodbc-signed.msi
+        local_file: mongosql-odbc-driver/installer/msi/mongoodbc.msi
```
bucaojit, line 509: nit: add the `display_name:` here to set it to `mongoodbc-signed.msi`

bucaojit (commented):
Regarding your questions:
1. This seems okay, it's reverting to SimpleLogger so that will log the issue should there be one.  Looks like something the build team hasn't added to the container. 
2. I'm not sure how to do this or if the tools are installed on the host, send a quick message to the build team to see if they think there is a way to do this.

EthanHardyMongo:
> 2. I'm not sure how to do this or if the tools are installed on the host, send a quick message to the build team to see if they think there is a way to do this.

I asked them and they told me that I can use `gpg --verify` for `gpg`, but for `jsign` there is no verification method. They mentioned using `Signtool`, but we would have to install this on the host I think. For now, I created verification for `ubuntu` since it was a trivial task.

bucaojit (approved):
LGTM!

PR #142, opened 2023-06-28
SQL-1444: Add missing get_info values found commonly in BI tools

pmeredit:


odbc/src/api/functions.rs
```diff
@@ -2238,6 +2344,70 @@ unsafe fn sql_get_infow_helper(
...
+///
+/// [`SQLGetInfo`]: https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/SQLGetInfo-function
+///
+///
+/// # Safety
+/// Because this is a C-interface, this is necessarily unsafe
+///
+#[named]
+#[no_mangle]
+pub unsafe extern "C" fn SQLGetInfo(
```
pmeredit, line 2391: PowerBI calls SQLGetInfo instead of SQLGetInfoW for some reason

odbc/src/api/functions.rs
```diff
@@ -2211,13 +2205,125 @@ unsafe fn sql_get_infow_helper(
...
+                InfoType::SQL_LOCK_TYPES => {
+                    i16_len::set_output_fixed_data(
+                        &SQL_LCK_NO_CHANGE,
+                        info_value_ptr,
+                        string_length_ptr,
+                    )
+                }
+                InfoType::SQL_POS_OPERATIONS => {
+                    i16_len::set_output_fixed_data(
+                        &0,
```
jchemburkar, line 2307: non-blocking: we group together things that share the same constant value, wondering for all of the `&0`'s, if they are all bitmasks for "unsupported" maybe we can group them together with a comment?

odbc/src/api/functions.rs
```diff
@@ -1776,58 +1782,41 @@ unsafe fn sql_get_env_attrw_helper(
...
-        connection_handle
-    )
-}
+macro_rules! sql_get_info_helper {
+($connection_handle:ident,
+ $info_type:ident,
+ $info_value_ptr:ident,
+ $buffer_length:ident,
+ $string_length_ptr:ident,
+ $string_func:path,
```
jchemburkar, line 1791: apologies if I'm missing it, but curious why this became a param, it seems like both calls to this macro use the same value for this

odbc/src/api/get_info_tests.rs
```diff
@@ -281,102 +296,221 @@ mod unit {
...
+        info_type = InfoType::SQL_TXN_ISOLATION_OPTION as u16,
+        expected_sql_return = SqlReturn::SUCCESS,
+        expected_length = std::mem::size_of::<u32>() as i16,
+        expected_value = SQL_TXN_SERIALIZABLE,
+        actual_value_modifier = modify_u32_value,
+    );
+
+    test_get_info!(
+        sql_database_name,
+        info_type = InfoType::SQL_DATABASE_NAME as u16,
```
jchemburkar, line 410: It seems difficult to test the "current catalog is some" case for this info type via these tests, curious if it came up at all in your POC work. Could be useful to add (possibly in an integration test) sometime in the future!

jchemburkar (approved):
LGTM! Left a few questions

odbc/src/api/functions.rs
pmeredit: I was considering that too. My main reason against that is that they may change independently in the future, but I'm very ambivalent.

odbc/src/api/functions.rs
pmeredit: One call passes "SQLInfo" the other passes "SQLInfoW", it's only used in tracing :) Though thanks for bringing this up, because it should also be used in the error message.

pmeredit: Oh you mean the string function. This was meant to be an ansi version of the string function, but this is actually working for power bi. That's troubling.

odbc/src/api/get_info_tests.rs
pmeredit: We do have a test for the ConnectAttr version, but you're correct we don't have it for this SQLGetInfo version. I can probably just add an integration test now.

pmeredit: Also, now I see that we are returning NO_DATA, not test, for the ConnectAttr version, and we should probably be consistent

pmeredit: Ok, added a test where we set the catalog first.

odbc/src/api/definitions.rs
```diff
@@ -555,12 +556,25 @@ pub const SQL_ROW_NUMBER_UNKNOWN: isize = -2;
...
 pub const SQL_SC_SQL92_ENTRY: u32 = 0x00000001;
 pub const SQL_INFO_Y: &str = "Y";
 pub const SQL_GB_GROUP_BY_CONTAINS_SELECT: u16 = 0x0002;
 pub const SQL_CB_PRESERVE: u16 = 2;
+pub const SQL_CA1_NEXT: u32 = 0x00000001;
+pub const SQL_CA2_READ_ONLY_CONCURRENCY: u32 = 0x00000001;
+pub const SQL_CA2_MAX_ROWS_SELECT: u32 = 0x00000080;
+pub const SQL_CA2_CRC_EXACT: u32 = 0x00001000;
+pub const MONGO_CA2_SUPPORT: u32 =
+    SQL_CA2_CRC_EXACT | SQL_CA2_MAX_ROWS_SELECT | SQL_CA2_READ_ONLY_CONCURRENCY;
```
nbagnard, line 571: Do we actually implement the logic for honouring the max rows setting?

nbagnard (changes requested):

odbc/src/api/functions.rs
```diff
@@ -1024,6 +1024,12 @@ pub unsafe extern "C" fn SQLExecDirectW(
         || {
             let query = input_text_to_string_w(statement_text, text_length as usize);
             let mongo_handle = MongoHandleRef::from(statement_handle);
```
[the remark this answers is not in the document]
pmeredit: Probably safe on debug.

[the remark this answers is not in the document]
pmeredit: I moved several of the tracing that should really be debug to debug (for example, there is no reason to log handle freeing on info). We should include in the documentation that debug level will log queries which can potentially leak sensitive column and table names. When we include the parser in the driver later on we could do a ticket to obscure the query identifiers.

odbc/src/api/get_info_tests.rs
pmeredit: Why would it matter? I never see any issue with supporting extra behavior.

odbc/src/api/functions.rs
```diff
@@ -2211,12 +2205,91 @@ unsafe fn sql_get_infow_helper(
...
+                    let attributes = conn.attributes.read().unwrap();
+                    if attributes.current_catalog.is_some() {
+                        $string_func(
+                            attributes.current_catalog.as_ref().unwrap().as_str(),
+                            info_value_ptr,
+                            buffer_length as usize,
+                            string_length_ptr,
+                        )
+                    } else {
+                        SqlReturn::NO_DATA
```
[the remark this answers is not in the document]
pmeredit: I'm fine with connection not open, actually. Let's do that.

odbc/src/api/functions.rs
```diff
@@ -1776,58 +1782,41 @@ unsafe fn sql_get_env_attrw_helper(
...
-    info_value_ptr: Pointer,
-    buffer_length: SmallInt,
-    string_length_ptr: *mut SmallInt,
-) -> SqlReturn {
     let conn_handle = MongoHandleRef::from(connection_handle);
     let mut err = None;
     let sql_return = match FromPrimitive::from_u16(info_type) {
         Some(some_info_type) => {
+            trace_odbc!(
+                info,
```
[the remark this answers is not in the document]
pmeredit: Absolutely.

odbc/src/api/get_info_tests.rs
```diff
@@ -63,6 +63,39 @@ macro_rules! test_get_info {
     }
 }
 
+macro_rules! test_get_info_no_data {
```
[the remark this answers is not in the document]
pmeredit: This is going to change anyway!

odbc/src/api/get_info_tests.rs
```diff
@@ -281,102 +329,250 @@ mod unit {
...
+        actual_value_modifier = modify_u32_value,
+    );
+
+    test_get_info_no_data!(
+        sql_database_name_default,
+        info_type = InfoType::SQL_DATABASE_NAME as u16,
+    );
+
+    #[test]
+    fn sql_database_name_set() {
```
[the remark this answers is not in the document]
pmeredit: It has never been completely clear that we need to maintain these "connection state" values since there is no way for an end user to even query them

resources/run_adf.sh
```diff
@@ -336,8 +336,11 @@ if [[ $? -ne 0 ]]; then
     fi
     rm -f $MONGOHOUSE_MQLRUN
     $GO run cmd/buildscript/build.go tools:download:mqlrun
-    rm -f $MONGOSQL_LIB
-    $GO run cmd/buildscript/build.go tools:download:mongosql
+
+    if [[ $HAVE_LOCAL_MONGOHOUSE -ne 1 || ! -f "$MONGOSQL_LIB" ]]; then
```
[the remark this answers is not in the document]
pmeredit: I don't actually even want to check this in. That was an oversight

odbc/src/api/functions.rs
```diff
@@ -2013,8 +2005,10 @@ unsafe fn sql_get_infow_helper(
                 | InfoType::SQL_CONVERT_VARBINARY
                 | InfoType::SQL_CONVERT_LONGVARBINARY
                 | InfoType::SQL_CONVERT_GUID => {
-                    // MongoSQL does not support CONVERT.
-                    i16_len::set_output_fixed_data(&SQL_U32_ZERO, info_value_ptr, string_length_ptr)
+                    // MongoSQL does not support the CONVERT scalar function, but clients also use
+                    // this to apply the CAST syntactic construct. The value we return for
+                    // SQL_CONVERT_FUNCTIONS alerts the client that we expect CAST and not CONVERT.
+                    i16_len::set_output_fixed_data(&SQL_U32_ALL, info_value_ptr, string_length_ptr)
```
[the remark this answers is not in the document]
pmeredit: It also bothers me because it only covers the to type and not the from type.

odbc/src/api/functions.rs
pmeredit: Since always? Their mashup logs very specifically say SQLGetInfo, as well. If I don't put SQLGetInfo in here it simply doesn't work.

odbc/src/api/functions.rs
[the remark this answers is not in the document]
nbagnard: I believe SQLFreeHandle used Info level for a specific reason. Can you double-check with @terakilobyte ?

Otherwise, yes, I am fine with the idea to log queries in debug and document that. We can look at obfuscating sensitive data later if needed. As far as I know Postgres, Mysql, etc... log their queries. And so does mongod, right?

odbc/src/api/functions.rs
nbagnard: This is not what I have been seeing.
Here are the old traces, for the beginning of the work with ODBC: https://docs.google.com/document/d/1eMnHvYwhLymsOvepjB7GgM9saN6asMl55Xr33448xes/edit#heading=h.gh965js96g5g
It calls SQLGetInfoW.
How would users be able to user (and Alexi) the latest driver if it was not working and calling SQLGetInfo?

Just a thought here. Last time you were using an old version of PowerBi. Is this one the newest one?

odbc/src/api/get_info_tests.rs
nbagnard: What matters is that adding "extra behaviour" is hard to track. It is also not expected by tools/users.
Not supporting extra behaviour means that you don't have to think twice or more every time you are deviating from the spec to make sure that it won't break anything.
In this case for example, I don't see any harm. But I am not the end user and if they rely on the call to fail if the connection was not made successfully, then they would be right, even if it seems very silly and they probably should rely on something else like the status of the their SQLxxConnect call.

odbc/src/api/functions.rs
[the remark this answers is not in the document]
terakilobyte: We have to log everything up through SQLDriverConnectW as Info because we can't know what their desired log level is, so the SQLFreeHandle call was left at Info to "complete" the log and not leave an impression that something was wrong from the user's perspective if they didn't see it in the log.

odbc/src/api/get_info_tests.rs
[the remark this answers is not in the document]
nbagnard: Yes, I don't know either. I have a feeling that the DM does that but I did not have time to look into it.
Can you can call SQLGetInfo before connection asking for the DB name and the DM let's you do that? If not, then it is very likely tracking states itself.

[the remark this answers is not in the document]
pmeredit: To clarify, this is a conversation Ryan and I had over a year ago. We considered even removing the state machines, but we haven't decided 100%

odbc/src/api/definitions.rs
pmeredit: Actually, that's a good point. Let me remove that.

odbc/src/api/functions.rs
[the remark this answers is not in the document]
nbagnard: umh, I am not sure that I follow the last comment.
It does cover both the to and from types. The "from" type is the `InfoType` and the to type is the list of supported types reported back via the bitmask, no?
That allows to build a matrix of support SQL to SQL conversion.

odbc/src/api/functions.rs
pmeredit: What I meant by "doesn't work" is that it wouldn't Direct Query anything that relied on SQLGetInfo output, i.e., any Cast. Should have been more clear there.

odbc/src/api/functions.rs
```diff
@@ -2013,8 +2004,10 @@ unsafe fn sql_get_infow_helper(
                 | InfoType::SQL_CONVERT_VARBINARY
                 | InfoType::SQL_CONVERT_LONGVARBINARY
                 | InfoType::SQL_CONVERT_GUID => {
-                    // MongoSQL does not support CONVERT.
-                    i16_len::set_output_fixed_data(&SQL_U32_ZERO, info_value_ptr, string_length_ptr)
+                    // MongoSQL does not support the CONVERT scalar function, but clients also use
+                    // this to apply the CAST syntactic construct. The value we return for
+                    // SQL_CONVERT_FUNCTIONS alerts the client that we expect CAST and not CONVERT.
+                    i16_len::set_output_fixed_data(&MONGO_CAST_SUPPORT, info_value_ptr, string_length_ptr)
```
nbagnard, line 2010: I wonder if we should advertise cast which are not supported since the data will always be Null.
Things like BOOL to TIMESTAMP or TIMESTAMP to INT for example.
Or is it better since then it will fold no matter what?

odbc/src/api/functions.rs
```diff
@@ -2211,12 +2204,92 @@ unsafe fn sql_get_infow_helper(
...
+                    if attributes.current_catalog.is_some() {
+                        i16_len::set_output_wstring_as_bytes(
+                            attributes.current_catalog.as_ref().unwrap().as_str(),
+                            info_value_ptr,
+                            buffer_length as usize,
+                            string_length_ptr,
+                        )
+                    } else {
+                        err = Some(ODBCError::ConnectionNotOpen);
+                        SqlReturn::ERROR
```
nbagnard, line 2277: Just a note here
I checked and this should be pre-covered by the DM.

08003 | Connection not open | (DM) The type of information requested in InfoType requires  an open connection. Of the information types reserved by ODBC, only  SQL_ODBC_VER can be returned without an open connection.

nbagnard (changes requested):
Can you revert back the log level for SQLFreeHandle?
And then I think you are all set.

odbc/src/api/functions.rs
pmeredit: We decided to reevaluate this during Direct Query

odbc/src/api/functions.rs
pmeredit: Sounds good, I like keeping this since rust forces us to have an else block, anyway. I guess I could add unreachable, but who knows if one of the other DMs is wrong (looking at you iodbc :)).

nbagnard (approved):
LGTM

pmeredit:
evergreen merge

PR #146, opened 2023-07-14
SQL-1500: Add "simple_types_only" mode to the ODBC Driver

EthanHardyMongo:
I'll be on vacation next week, so I'm going to describe what else needs to be done on this PR here for when I get back.

1. I need to define `core/src/bson_type_info/simple_type_info.rs` to reflect its purpose. It was basically just acting as a place holder for me while I did all the necessary refactoring. So far, all I have done is change some unknown types to `VARCHAR` or `FLOAT` as specified by the ticket. However, I have not taken fixed/unfixed size information into account. This information will be needed to use as little memory as possible.  (fyi `bson_type_info.rs` is now `standard_type_info.rs`).
2. I need to make more tests to test the `simple_types_only` mode functionality. On the bright side, all existing unit tests are passing! I think all the major refactoring is done, but the new tests I create will reveal if that's really the case.


odbc/src/api/data_tests.rs
```diff
@@ -2798,194 +3193,138 @@ mod unit {
...
             let _ = Box::from_raw(conn as *mut WChar);
             let _ = Box::from_raw(env as *mut WChar);
         }
     }
+
+    #[test]
+    fn sql_fetch_and_more_results_basic_functionality_test() {
+        sql_fetch_and_more_results_basic_functionality(SchemaMode::Standard);
+        sql_fetch_and_more_results_basic_functionality(SchemaMode::Simple);
+    }
```
EthanHardyMongo, line 3221: I did `cargo fmt` and everything got moved, so it looks like I made a lot of changes. All I really did though was make these new testing functions and add a parameter to the existing ones so that I could test both schema modes.

core/src/type_info/simple_bson_type_info.rs
```diff
@@ -0,0 +1,453 @@
...
+    type_info::{TypeInfoFields, SQL_PRED_BASIC, SQL_PRED_NONE, SQL_SEARCHABLE},
+};
+
+#[non_exhaustive]
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct SimpleBsonTypeInfo {
+    pub type_info_fields: TypeInfoFields,
+}
+
+impl SimpleBsonTypeInfo {
```
EthanHardyMongo, line 12: Almost all of the constants defined below are the same as `StandardBsonTypeInfo` with the only change being that the `sql_type` and `non_concise_type` is now `VARCHAR` in places where it was `UNKNOWN_TYPE`. Also, I changed `is_case_sensitive` to `true` for the types I changed to `VARCHAR`. However, some of them had more changes. These ones are `OBJECTID`, `NULL`, `UNDEFINED`, `TIMESTAMP`, `MINKEY`, `MAXKEY`,

resources/integration_test/tests/functions.yml
```diff
@@ -59,7 +59,7 @@ tests:
     - [ "array", 0, null, null, null, null, 1, 0, 0, null, 0, null, "array", null, null, 0, null, null, null ]
     - [ "bson", 0, null, null, null, null, 1, 0, 0, null, 0, null, "bson", null, null, 0, null, null, null ]
     - [ "dbPointer", 0, null, null, null, null, 1, 0, 2, null, 0, null, "dbPointer", null, null, 0, null, null, null ]
-    - [ "decimal", 0, null, null, null, null, 1, 0, 2, 0, 0, 0, "decimal", null, null, 0, null, null, null ]
```
EthanHardyMongo, line 62: This change has to do with the `// TODO SQL-1068: Change to SqlDataType::DECIMAL` that was in  the now deleted `bson_type_info.rs`.

core/src/bson_type_info.rs
```diff
@@ -19,6 +25,38 @@ pub struct BsonTypeInfo {
...
+
+    const fn default() -> Option<Self> {
+        Some(Self {
+            sql_type: SqlDataType::VARCHAR,
+            non_concise_type: SqlDataType::VARCHAR,
+            precision: None,
+            octet_length: None,
+            fixed_bytes_length: None,
+        })
+    }
```
EthanHardyMongo, line 59: Functions need to be constant to work in constant declarations, so I had to get rid of the `default()` trait. Instead, I just moved the `default()` function into the `impl SimpleTypeInfo`.

odbc/src/api/data_tests.rs
```diff
@@ -158,2196 +165,2239 @@ lazy_static! {
...
-            EnvState::ConnectionAllocated,
-            RefCell::new(None),
-        ))));
-        let conn = Box::into_raw(Box::new(MongoHandle::Connection(Connection::with_state(
-            env as *mut _,
-            ConnectionState::Connected,
-        ))));
-        let stmt = Statement::with_state(conn as *mut _, StatementState::Allocated);
-        *stmt.mongo_statement.write().unwrap() = Some(Box::new(MongoQuery::new(
+    static ref SIMPLE_BSON_TYPE_MQ: MongoQuery = MongoQuery::new(
```
EthanHardyMongo, line 301: I had to keep `SIMPLE_BSON_TYPE_MQ` for testing. The only difference between the two is the `TypeMode`.

core/src/primary_keys.rs
```diff
@@ -9,41 +9,47 @@ lazy_static! {
...
+            TypeMode::Standard,
             Nullability::NO_NULLS
         ),
         MongoColMetadata::new_metadata_from_bson_type_info(
             "",
             "".to_string(),
             "KEY_SEQ".to_string(),
             BsonTypeInfo::INT,
+            TypeMode::Standard,
             Nullability::NO_NULLS
```
EthanHardyMongo, line 45: There is no difference between the simple and regular versions of `BsonTypeInfo::INT` and `BsonTypeInfo::STRING`, so we can just set this to `TypeMode::Standard`.

terakilobyte (commented):
Overall this looks great! Your first iteration was good, and your refactoring to reduce a lot of code is also really well done.

I'll test this from an updated PowerBI connector soon to ensure it works as we expect for an e2e validation before approving this.

core/src/odbc_uri.rs
```diff
@@ -22,6 +22,7 @@ pub const UID: &str = "uid";
 pub const URI: &str = "uri";
 pub const APPNAME: &str = "appname";
 pub const LOGLEVEL: &str = "loglevel";
+pub const SIMPLE: &str = "simple_types_only";
```
pmeredit, line 25:
```suggestion
pub const SIMPLE_TYPES_ONLY: &str = "simple_types_only";
```
[nit] everywhere else the name we use for the constant is the same as the constant, so I'm suggesting this just for consistency.

core/src/odbc_uri.rs
```diff
@@ -32,7 +33,7 @@ const SERVER_KWS: &[&str] = &[SERVER];
 
 lazy_static! {
     static ref KEYWORDS: RegexSet = RegexSetBuilder::new(
-        [DATABASE, DRIVER, DSN, PASSWORD, PWD, SERVER, USER, UID, URI, APPNAME, LOGLEVEL]
+        [DATABASE, DRIVER, DSN, PASSWORD, PWD, SERVER, USER, UID, URI, APPNAME, LOGLEVEL, SIMPLE]
```
pmeredit, line 36:
```suggestion
        [DATABASE, DRIVER, DSN, PASSWORD, PWD, SERVER, USER, UID, URI, APPNAME, LOGLEVEL, SIMPLE_TYPES_ONLY]
```

pmeredit (changes requested):
Looks excellent, just one substantive change and one nit.

odbc/src/handles/definitions.rs
```diff
@@ -237,6 +238,8 @@ pub struct Connection {
     // all Statements allocated from this Connection
     pub statements: RwLock<HashSet<*mut MongoHandle>>,
     pub errors: RwLock<Vec<ODBCError>>,
+    // type_info represents whether we're using simple_type_info or standard_type_info
```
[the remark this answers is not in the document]
EthanHardyMongo: Good point. Maybe something like `type_mode indicates if BsonTypeInfo.simple_type_info will be utilized in place of standard BsonTypeInfo fields.`?

nbagnard (changes requested):
I did not find test for TypeMode::Simple.
I'd like to have one for the resultset return by sqlGetTypeInfo and for a query too.
Otherwise, it is looking much better this way for sure :)

core/src/bson_type_info.rs
```diff
@@ -331,6 +385,7 @@ impl BsonTypeInfo {
         is_auto_unique_value: None,
         is_unsigned: None,
         num_prec_radix: None,
+        simple_type_info: SimpleTypeInfo::new(68, 68, 68),
```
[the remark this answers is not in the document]
EthanHardyMongo: According to [this](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#extended-json-format), Timestamp should return `{"$timestamp": {"t": pos-integer, "i": pos-integer}}`. So I got the 68 from the length of `{"$timestamp": {"t": 9223372036854775807, "i": 9223372036854775807}}` where 9223372036854775807 is the 64-bit integer max.

core/src/collections.rs
```diff
@@ -22,34 +21,39 @@ lazy_static! {
             "".to_string(),
             "TABLE_CAT".to_string(),
             BsonTypeInfo::STRING,
+            TypeMode::Standard,
```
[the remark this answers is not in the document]
EthanHardyMongo: Good point. Is this something you want changed or just something to keep in mind for the future?

core/src/type_info.rs
```diff
@@ -38,132 +38,151 @@ lazy_static! {
             "".to_string(),
             "TYPE_NAME".to_string(),
             BsonTypeInfo::STRING,
+            TypeMode::Standard,
```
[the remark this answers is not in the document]
EthanHardyMongo: Yeah I think you are right. I'll make that change.

pmeredit (approved):
LGTM once we have the simple type tests!

core/src/type_info.rs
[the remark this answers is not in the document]
EthanHardyMongo: Is there a better way to handle this other than by creating a duplicate METADATA vector for TypeMode::Simple? This would basically be what I did before we decided to get rid of all the duplicate code.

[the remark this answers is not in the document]
terakilobyte: I might be misunderstanding here, but why would there be two types here? This is describing column metadata, which is statically known to be strings since it would never change on type, no?

core/src/bson_type_info.rs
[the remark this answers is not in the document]
nbagnard: ah, I see why I was confused.
Here is the query I used : `select CAST('2023-05-20 00:00:00.000' AS TIMESTAMP) from foo`
And the result was `{"$date": "2023-05-20T00:00:00Z"}`
But this is supposed to be `BSON_TIMESTAMP`.

core/src/collections.rs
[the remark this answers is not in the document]
nbagnard: I mean, if this is quick enough, I'd do it.
If it is a pain to track these, then it's Ok, you can keep it as is.
It is doing the exact same thing, only it's more verbose.

core/src/json_schema.rs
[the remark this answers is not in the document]
nbagnard: ah, I see. I was just curious.

core/src/type_info.rs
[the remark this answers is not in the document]
nbagnard: Yes, sorry, I mixed up things.
I spoke with Ethan.
What I noticed is the SqlGetTypeInfo returns the same data types information, no matter what the connection type_mode is.
Ethan is working on it.

resources/integration_test/tests/queries.yaml
```diff
@@ -227,6 +227,27 @@ tests:
           # 11: {"$undefined":true},
       ]
 
+  - description: query_types_other using simple_types_only mode
+    test_definition: SELECT * FROM types_other
+    db: integration_test
+    simple_type: true
+    expected_result:
```
EthanHardyMongo, line 234: I didn't change the expected results at all, and this test worked. Is this how it should be in simple_types_only mode?

integration_test/src/test_runner/test_generator_util.rs
```diff
@@ -88,6 +88,7 @@ pub fn generate_baseline_test_file(entry: &TestEntry, stmt: HStmt) -> Result<()>
     let test_entry = TestEntry {
         description: entry.description.clone(),
         db: entry.db.clone(),
+        simple_type: entry.simple_type,
```
nbagnard, line 91: [nit] For flags, use the form "is_xxx". It makes it very easy to identify the purpose this way.

resources/integration_test/tests/queries.yaml
nbagnard: This is because you don't have any metadata test.
You need `expected_sql_type`, `expected_precision, `expected_display_size` and maybe `expected_length` with the arrays set with the correct values.
I would add a test just for checking metadata in simple mode:
`query_types_other using simple_types_only_mode` (no need to check the expected result since we ask for the string representation in both cases) and maybe the same query as `select_star_unordered`.
Or even better, find a query which returns columns with complex and simple types and add only the metadata tests in both simple and standard mode. Making sure that in simple mode the complex types are returned as VARCHAR and the other types stay the same then for the standard mode.

nbagnard (changes requested):
This is looking very good.
The last remaining thing are the metadata for query tests.
I tried to explain what to do. If it is unclear, let me know and we can quickly clear that offline.

core/src/col_metadata.rs
```diff
@@ -67,39 +68,59 @@ impl MongoColMetadata {
             display_size: bson_type_info.fixed_bytes_length,
```
nbagnard, line 68: @EthanHardyMongo I think you need to update this line to get the display size to behave has expected.
```suggestion
            display_size: bson_type_info.fixed_bytes_length(type_mode),
```

nbagnard (approved):
LGTM!
If evergreen is happy, then we are good to go 🚀

resources/integration_test/tests/queries.yaml
```diff
@@ -67,6 +67,15 @@ tests:
...
 
+  - description: select_star_unordered_using_simple_types_only_mode
+    test_definition: SELECT * FROM class
+    is_simple_type: true
+    db: integration_test
+    expected_display_size: [ 4, 1, 0, 16, 4 ]
+    expected_length: [ 4, 1, 0, 16, 4 ]
+    expected_sql_type: [ 4, -7, -9, 93, 4 ]
+    expected_precision: [ 10, 1, 0, 23, 10 ]
+
```
EthanHardyMongo, line 78: The `simple_types_only` version of this test expects the same results as the regular version.

nbagnard: Yes, I saw the change.
It's not how I would have done it. Because, even though you know today that the types info did not change, nothing will show this in the long run.
Separate metadata tests would have been better. What we would really need is a complete coverage of all data types in 2 tests. 
However, the tests you added for complex types has a simple type column and we see it does not change.
And with the tests you added, we achieve the coverage, only not with targeted tests, which can make things harder to grasp when it fails.

EthanHardyMongo:
evergreen merge

PR #147, opened 2023-07-20
SQL-1374 & SQL-1541: update run_adf and tests

terakilobyte:
We're back to using the master branch for ADF. Using the latest tag doesn't give us anything much different, and potentially leads to problems like this PR addresses.

The script has been updated to start a compute node, a new requirement. Additionally, nullability has been updated due to a fix in ADF that bubbled down.


resources/run_adf.sh
```diff
@@ -280,6 +297,43 @@ else
   fi
 fi
 
+manage_compute_node() {
+  if [ $ARG = $START ]; then
+    check_compute_node $MONGO_DOWNLOAD_DIR
```
bucaojit, line 302: Here, we're checking if the compute node is running but don't check the return or change behavior based on that.  An example of how check_mongohoused is doing it [here](https://github.com/mongodb/mongo-odbc-driver/blob/eed6cce38e8bbbbc277ac6d916ba55f0350e76e8/resources/run_adf.sh#L284).

resources/run_adf.sh
```diff
@@ -307,14 +361,16 @@ if [[ $? -ne 0 ]]; then
         fi
         cd $MONGOHOUSE_DIR
 
-        # for now, we checkout a specific working commit of ADF
-        # TODO SQL-1374: Find a better long term solution
-        git checkout 247a246
 
         export GOPRIVATE=github.com/10gen
+        # make sure mod vendor is cleaned up
+        $GO mod vendor
```
bucaojit, line 367: We haven't had to add this vendoring step before.  Do you have this set in your env, something like `export GOFLAGS=-mod=vendor`?  If so, try unsetting that and see if run_adf works without this vendoring.

resources/run_adf.sh
```diff
@@ -122,6 +126,19 @@ check_version() {
   fi
 }
 
+check_compute_node() {
```
bucaojit, line 129: Just to double check, now ADF uses a completely separate mongod as well as the mongod where the collections are stored?

resources/run_adf.sh
```diff
@@ -280,6 +297,43 @@ else
...
+    fi
+
+    waitCounter=0
+    while : ; do
+        check_compute_node
+        if [[ $? -eq 0 ]]; then
+            break
+        fi
+        if [[ "$waitCounter" -gt $TIMEOUT ]]; then
+            echo "ERROR: Local mongod did not start under $TIMEOUT seconds"
```
bucaojit, line 322:
```suggestion
            echo "ERROR: Local compute node mongod did not start under $TIMEOUT seconds"
```
Differentiating which mongod is failing

resources/integration_test/tests/columns_all_filter.yml
```diff
@@ -3,15 +3,15 @@ tests:
     test_definition: [ "sqlcolumnsw", "integration_test", 0, "", 0, "%o%", 3, "%i%", 3 ]
     db: integration_test
     expected_result:
-      - ["integration_test", null, "a_non_lexicographic_field_order", "_id", "4", "int", 10, "4", 0, 10, 1, "", null, "4" , null , 0, 4, "YES"]
```
bucaojit, line 6: Which change in ADF caused these values to get updated?

bucaojit (changes requested):
Nice work on the run_adf script, it'll be great to be testing on the latest ADF/mongosql again.

resources/run_adf.sh
terakilobyte: This is incompatible with the setup instructions for [sqlproxy](https://github.com/10gen/sqlproxy#unix), so it seems easy enough (and works) to just run it as part of the script.

resources/run_adf.sh
terakilobyte: This is correct.

resources/integration_test/tests/columns_all_filter.yml
terakilobyte: https://github.com/10gen/mongohouse/commit/d768d7aef9772a7926dd01cd61508e633dcd79af

resources/run_adf.sh
```diff
@@ -77,22 +81,22 @@ MONGOSH_DOWNLOAD_BASE=https://downloads.mongodb.com/compass
 
 ## Linux
 # Ubuntu 22.04
-MONGO_DOWNLOAD_UBUNTU=mongodb-linux-x86_64-ubuntu2204-6.0.4.tgz
+MONGO_DOWNLOAD_UBUNTU=mongodb-linux-x86_64-ubuntu2204-7.0.0-rc6.tgz
```
jchemburkar, line 84: personal question, but wondering the cadence for when we change these. I.e. is it just whenever we see a need (useful features, major release, etc)? and are there any particular tests to run, or more just eyeballing if existing tests work?

resources/run_adf.sh
```diff
@@ -53,11 +56,12 @@ LOGS_PATH=$LOCAL_INSTALL_DIR/logs
 DB_CONFIG_PATH=$(pwd)/resources/integration_test/testdata/adf_db_config.json
 MONGOD_PORT=28017
 MONGOHOUSED_PORT=27017
+export COMPUTE_MODE_MONGODB_PORT=47017
 START="start"
 STOP="stop"
 MONGOD="mongod"
 MONGOHOUSED="mongohoused"
-TENANT_CONFIG="./testdata/config/mongodb_local/tenant-config.json"
+TENANT_CONFIG="./testdata/config/inline_local/tenant-config.json"
```
jchemburkar, line 64: I see they still have mongodb_local/inline_local, was poking around the datasets but wondering what precipitated this change?

jchemburkar (approved):
Looks good to me pending Oliver's comments, and just had a couple of extra questions

resources/run_adf.sh
bucaojit: I created this ticket to address possibly automating updating the versions.  We haven't really discussed the cadence, so far it's been when someone brings it up or makes a major change to run_adf. 
https://jira.mongodb.org/browse/SQL-981

resources/run_adf.sh
terakilobyte: I updated this based off of their README. Locally, the `inline_local` always works to start it up, but having issues with `mongodb_local`.

bucaojit (approved):
LGTM!

PR #148, opened 2023-08-10
SQL-1557: Update SQLDriverConnectW to just return SUCCESS rather than…

pmeredit:
… SUCCESS_WITH_INFO when the out_connection_string is NULL or buffer_len is 0

This is passing locally. The resultset tests show that we no longer return SUCCESS_WITH_INFO when a null buffer/0 buffer length are passed.

### DIFF
--- a/integration_test/tests/common/mod.rs
+++ b/integration_test/tests/common/mod.rs
@@ -153,6 +153,15 @@ pub fn connect_with_conn_string(env_handle: HEnv, in_connection_string: String) 
             // Originally, this would return SUCCESS_WITH_INFO since we pass null_mut() as
             // out_connection_string and 0 as buffer size. Now, this should always return SUCCESS.
             SqlReturn::SUCCESS => (),
+            // TODO SQL-1568: Windows DM is still changing SUCCESS to SUCCESS_WITH_INFO
+            SqlReturn::SUCCESS_WITH_INFO => {
+                if !cfg!(windows) {
+                    return Err(Error::DriverConnect(
+                        sql_return_to_string(SqlReturn::SUCCESS_WITH_INFO),
+                        get_sql_diagnostics(HandleType::Dbc, dbc),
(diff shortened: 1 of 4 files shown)


terakilobyte (commented):
This looks like the ticket. I'm going to checkout the pr in ubuntu and test it with `iusql`

terakilobyte (approved):
LGTM. Confirmed this works locally.

pmeredit:
evergreen merge

odbc/src/api/functions.rs
```diff
@@ -930,6 +930,7 @@ pub unsafe extern "C" fn SQLDriverConnectW(
     string_length_2: *mut SmallInt,
     driver_completion: DriverConnectOption,
 ) -> SqlReturn {
+    dbg!("!!!!");
```
nbagnard: Left over debug.

integration_test/tests/common/mod.rs
```diff
@@ -150,7 +150,18 @@ pub fn connect_with_conn_string(env_handle: HEnv, in_connection_string: String)
             str_len_ptr,
             DriverConnectOption::NoPrompt,
         ) {
-            SqlReturn::SUCCESS | SqlReturn::SUCCESS_WITH_INFO => (),
+            // Originally, this would return SUCCESS_WITH_INFO since we pass null_mut() as
```
nbagnard, line 153: I don't think that there is much value in keep some kind of historical comment on a test.
If you want to track the changes, I'd suggest to tag the comment in the feature change itself (in function.rs) with the JIRA ticket which introduced the change.
This help track changes overtime and now which version it was introduced into.

pmeredit:
evergreen push

pmeredit:
evergreen merge

pmeredit:
> evergreen push

Heh, where was my mind?

PR #151, opened 2023-08-18
log in correct instance

terakilobyte:
I don't know if I wasn't feeling well and wasn't feeling it, or I'm just an idiot and forgot to include these changes.

This will log whatever is connecting if it supplies an ODBC URI APPNAME key. Previously, it would only be if there was a mongodb uri appName.


nbagnard (commented):
I tried it out and I am wondering if you could use trace_odbc instead of a log line with no handle info.
Right now the log line is context less and I am thinking that this could be a pain/not that useful if we have multiple connections happening at once with different connector versions. We wouldn't always know which one is which connection. I don't know if using the gateway could be a possible way of getting there. I think the connector and driver are only installed once, but I am not 100% sure.

terakilobyte:
Yes, entirely possible. I'll just have to pipe it up and do a check for it in sql_driver_connect. I'll think about the cleanest way to do that.

PR #153, opened 2023-10-20
SQL-1706: update ODBC driver to handle new result set metadata order

jchemburkar:
One open question, discussed in standup 10/20, is what to do with the test failures. This is caused by changes being in mongosql master, but not the latest release (nor in ADF). To test this, the master branch of mongosql-rs was build, and `mongosql.dll` was copied into `local_adf/mongohouse/artifacts` so adf would point at the latest code. All tests are passing locally with these changes


jchemburkar:
@nbagnard @pmeredit in working on JDBC I'm seeing some odd behavior in testing, I think relevant here too. Converting to draft for now while I sort it out -- everything in here should still be relevant, but may be a couple more changes. Will un-draft and ping once that's done!

nbagnard:
> I'm seeing some odd behavior in testing

Oh, and what's this odd behavior?

jchemburkar:
> > I'm seeing some odd behavior in testing
> 
> Oh, and what's this odd behavior?

I realized there was a specific test that I should have updated but was still working -- which led me to realize I had an issue in how I was running the latest mongosql. Once I fixed that, I realized the mongosql code needs a small update. When we serialize the successful translation, [this](https://github.com/10gen/mongosql-rs/blob/2cfca64bd7db1d7ebbd44a02f56b1a1db554487f/mongosql-c/src/lib.rs#L133) line is still converting the properties list to a document, and as a result, "select a, b from foo" will have the shape `[("", "a"), ("", "b")]` but will get serialized as `{"": ["b"]}`

core/src/col_metadata.rs
```diff
@@ -251,7 +237,10 @@ impl SqlGetSchemaResponse {
             .properties
             .clone()
             .into_iter()
-            .sorted_by(|a, b| Ord::cmp(&a.0, &b.0))
+            .sorted_by(|a, b| match sort_metadata {
+                true => Ord::cmp(&a.0, &b.0),
+                false => Ord::cmp(&a.0, &a.0),
```
pmeredit, line 242: Could just replace this with `Ordering::Equal`

PR #154, opened 2023-11-02
SQL-1751, SQL-1752: Add necessary constants to expose new mongosql functionality, and update simple_types to use EXT_W_VARCHAR

pmeredit:


pmeredit:
Closing this PR to avoid this accidentally getting into a release. This will be moved to a branch called topic/direct_query

bucaojit:
I see this is closed now, wanted to add this here so it's not lost:

I was looking at the test failure and the integration tests will need to be updated with the new buffer length:
https://github.com/mongodb/mongo-odbc-driver/blob/23f19cf81089236949c8e50c7e9424eea53e3ef1/resources/integration_test/tests/columns_all_filter.yml#L6
First row, Column 8 is expecting `65535`.  It's likely there are more areas but the test stopped after the first error. 
And [column 8 is](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlcolumns-function?view=sql-server-ver16):  BUFFER_LENGTH (ODBC 1.0) 	8 	Integer

PR #157, opened 2023-11-09
SQL-1706: order result set metadata by select list order

jchemburkar:
To test, I ran ADF with the [SQL-1773](https://github.com/10gen/mongosql-rs/pull/636) branch of mongosql, and all of unit, integration, and result set tests pass. 

Result set tests will fail for the time being, because select order =/= alphabetical order for `inner_join` and `select_as_confirm_column_label`.


core/src/col_metadata.rs
```diff
@@ -172,31 +176,53 @@ impl SqlGetSchemaResponse {
...
+                        .map(|col| (vec![col.table_name.clone(), col.col_name.clone()], col)))
+                })
+                // flatten the key-value pairs representing the metadata into a single vector,
+                // then finally convert to a BTree
+                .flatten_ok()
+                .collect::<Result<HashMap<Vec<String>, MongoColMetadata>>>()?;
+
+        Ok(match self.select_order {
+            // in the select list order is None, for example if using an older adf version, sort
+            None => processed_result_set_metadata
```
jchemburkar, line 209: in theory, this shouldn't be hit when mongosql + ADL changes have been released. It was a bit more natural in JDBC, which was a nice side effect. If its considered overkill we could just error if select order is empty

pmeredit: I actually like leaving this

pmeredit (approved):
LGTM, see if it works with that clone removed

core/src/col_metadata.rs
```diff
@@ -172,31 +177,51 @@ impl SqlGetSchemaResponse {
...
+                .into_values()
+                .sorted_by(|a, b| match Ord::cmp(&a.table_name, &b.table_name) {
+                    core::cmp::Ordering::Equal => Ord::cmp(&a.col_name, &b.col_name),
+                    v => v,
+                })
+                .collect(),
+            // given a select order, convert the values of the map into an ordered vector
+            _ => self
+                .select_order
+                .clone()
```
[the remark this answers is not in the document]
jchemburkar: 👍 I think you're right, will verify but may now be able to use `as_ref() ... iter()` instead

resources/integration_test/tests/queries.yaml
```diff
@@ -80,21 +80,21 @@ tests:
...
-      - [ true, Mike, 303 ]
-      - [ false, Mary, 204323 ]
-      - [ false, Pete, 10 ]
-    expected_bson_type: ["bool", "string", "int"]
+      - [ true, 10329, John ]
+      - [ false, 342, Jane  ]
+      - [ true, 303, Mike ]
+      - [ false, 204323, Mary ]
+      - [ false, 10, Pete ]
+    expected_bson_type: ["bool", "int", "string"]
```
bucaojit, line 88: The test updates look correct, but the result-set test is failing because it is expecting an `int` column but got a `string`.
Not sure why, I tried running the test generator to confirm the values but that's hitting a different error entirely when building. 

```
[2023/11/13 20:20:24.522]    --> integration_test\src\test_runner\test_generator_util.rs:159:53
[2023/11/13 20:20:24.522] 159 |                 SqlDataType::EXT_GUID => CDataType::Guid,
[2023/11/13 20:20:24.522]     |                                                     ^^^^ variant or associated item not found in `CDataType`
```

bucaojit (changes requested):
Changes look good, I'm curious to see why the result set test is failing.

resources/integration_test/tests/queries.yaml
jchemburkar: this test is expected to fail still until mongosql gets released, and then subsequent changes are made in monoghouse. So for now, I have been testing with a local fork of both, which shows these tests passing

bucaojit: Ah gotcha.  Thanks for explaining 👍

bucaojit (approved):
LGTM

nbagnard (approved):
LGTM!

PR #158, opened 2023-11-14
SQL-1265: Remove workaround for driver cleanup

mattChiaravalloti:
This PR removes the 200ms sleep after dropping the client. The ticket suggests this should be safe now that we've upgraded to rust driver 2.7.

### DIFF
--- a/odbc/src/api/functions.rs
+++ b/odbc/src/api/functions.rs
@@ -852,6 +852,14 @@ pub unsafe extern "C" fn SQLDisconnect(connection_handle: HDbc) -> SqlReturn {
             let conn = must_be_valid!((*conn_handle).as_connection());
             // set the mongo_connection to None. This will cause the previous mongo_connection
             // to drop and disconnect.
+            conn.mongo_connection
+                .write()
+                .unwrap()
+                .as_ref()
+                .unwrap()
+                .client
+                .clone()
+                .shutdown();
             *conn.mongo_connection.write().unwrap() = None;
             SqlReturn::SUCCESS
         },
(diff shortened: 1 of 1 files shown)


odbc/src/api/functions.rs
```diff
@@ -853,11 +853,6 @@ pub unsafe extern "C" fn SQLDisconnect(connection_handle: HDbc) -> SqlReturn {
             // set the mongo_connection to None. This will cause the previous mongo_connection
             // to drop and disconnect.
             *conn.mongo_connection.write().unwrap() = None;
-            // Temporary workaround for https://jira.mongodb.org/browse/RUST-1099
-            // This allows time for the underlying async runtime to clean up all of its
-            // resources before we report back success and subsequently drop the connection
-            // handle entirely
-            std::thread::sleep(std::time::Duration::from_millis(200));
```
terakilobyte, line 860: One small change request. It appears the driver team introduced the [shutdown](https://docs.rs/mongodb/2.7.1/mongodb/sync/struct.Client.html#method.shutdown) method in this [PR](https://github.com/mongodb/mongo-rust-driver/pull/920), and I think it would be best if we called that to ensure things are properly shut down.

terakilobyte (changes requested):

odbc/src/api/functions.rs
mattChiaravalloti: Done

odbc/src/api/functions.rs
```diff
@@ -852,12 +852,15 @@ pub unsafe extern "C" fn SQLDisconnect(connection_handle: HDbc) -> SqlReturn {
...
             // set the mongo_connection to None. This will cause the previous mongo_connection
             // to drop and disconnect.
+            conn.mongo_connection
+                .write()
+                .unwrap()
+                .as_ref()
+                .unwrap()
+                .client
+                .clone()
+                .shutdown();
```
terakilobyte, line 862: 🤼 😅

terakilobyte (approved):
lgtm!

PR #159, opened 2023-11-16
SQL-1788: Give Decimal128 a simple_type

pmeredit:


nbagnard:
As mentioned on the JIRA ticket, I don't know if we need to do this. It looks we thought we could support Bson::Decimal128 to SQL_DECIMAL.
Maybe before mapping it to VARCHAR we could get a better understanding of the actual issue here.

pmeredit:
> As mentioned on the JIRA ticket, I don't know if we need to do this. It looks we thought we could support Bson::Decimal128 to SQL_DECIMAL. Maybe before mapping it to VARCHAR we could get a better understanding of the actual issue here.

They really aren't compatible. If we just return this as a simple_type, a user can insert a cast to double.

core/src/bson_type_info.rs
nbagnard, line 1: You need to revert Ethan's change and set`sql_type` and `non_concise_type` to UNKOWN as before.
The commit id is d5ce256.
This is the commit which turn Decimal128 from unknown to SQL_DECIMAL.

nbagnard (changes requested):

terakilobyte:
I'm a little worried about usability here. In speaking with Alexi in the past, she was okay with loss of precision. Users _can_ cast, yes, but it'd be far more usability for the drag and drop folks to just return it as a number.

pmeredit:
> I'm a little worried about usability here. In speaking with Alexi in the past, she was okay with loss of precision. Users _can_ cast, yes, but it'd be far more usability for the drag and drop folks to just return it as a number.

Silently converting data with a _huge_ loss of percentage with no escape hatch seems strictly worse

terakilobyte:
> > I'm a little worried about usability here. In speaking with Alexi in the past, she was okay with loss of precision. Users can cast, yes, but it'd be far more usability for the drag and drop folks to just return it as a number.

> Silently converting data with a huge loss of percentage with no escape hatch seems strictly worse

Hmm.. I agree. I was thinking about opting for returning as a string "123.512341234123" so that it'd be easier for users to click transform in the power bi ui, but not going to bikeshed this any longer.

terakilobyte (approved):
:shipit:

pmeredit:
> > > I'm a little worried about usability here. In speaking with Alexi in the past, she was okay with loss of precision. Users can cast, yes, but it'd be far more usability for the drag and drop folks to just return it as a number.
> 
> > Silently converting data with a huge loss of percentage with no escape hatch seems strictly worse
> 
> Hmm.. I agree. I was thinking about opting for returning as a string "123.512341234123" so that it'd be easier for users to click transform in the power bi ui, but not going to bikeshed this any longer.

Ohhh. Yea, that's a harder call

nbagnard (approved):
LGTM!

PR #160, opened 2023-11-20
SQL-1717: Static logger with driver level initialization

nbagnard:


shared_sql_utils/src/odbcinst.rs
```diff
@@ -31,3 +70,128 @@ extern "C" {
...
 }
+
+#[derive(Debug, Default)]
+pub struct DriverSettings {
+    pub driver: String,
+    pub setup: String,
+    pub log_level: String,
+}
+
+impl DriverSettings {
```
terakilobyte, line 81: Not strictly necessary, but wdyt about moving the `DriverSettings` and associated additions its own file? I had initially thought to keep this file pure as the interface for the odbcinst functions.

terakilobyte (changes requested):
🔥 This is awesome! I have a few change requests but nothing major.

bucaojit (changes requested):
Nice work with the static logger and the tests 💯 

I had a few suggestions with the evergreen file.

shared_sql_utils/src/odbcinst.rs
nbagnard: To be honest, I don't like the current name because I don't exactly it's supposed to be representing. odbcinst is the name of the initialization file/section for drivers and it looks like you thought of it as the API for drivers and DSNs configurations.
What about I rename this file `odbc_configuration.rs` and have functions from the installer DDL API in it. This matches the [description](https://learn.microsoft.com/en-us/sql/odbc/reference/install/installer-dll?view=sql-server-ver16).
And I move DriverSettings in `driver_settings.rs`?

terakilobyte: I'm fine with that, though the header files are called `odbcinst` or have it in their name in 2/3 instances
```
#[cfg_attr(target_os = "linux", link(name = "odbcinst", kind = "dylib"))]
#[cfg_attr(target_os = "macos", link(name = "iodbcinst", kind = "dylib"))]
#[cfg_attr(target_os = "windows", link(name = "odbccp32", kind = "raw-dylib"))]
```

pmeredit: Windows version was first though :P

nbagnard: oh, I see, you use the name of the h files, sort of.

terakilobyte: We can call it `file_of_utility_functions_that_all_behave_differently.rs` if you want! 😭

nbagnard: I like that! Let me update the file name ;)

bucaojit (approved):
LGTM

terakilobyte (approved):
lgtm

PR #161, opened 2023-11-27
SQL-1548: re-enable macos11 arm tests

terakilobyte:
As advertised on the tin!

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -1627,15 +1627,15 @@ buildvariants:
       - name: integration-test
       - name: result-set-test
 
-  # - name: macos-arm
-  #   display_name: "macOS 11.0 ARM 64"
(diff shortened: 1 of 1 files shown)


nbagnard:
Do you mind checking MacOS 13 arm? I was never able to get it to work and reproduced the issue on Evergreen.
I just don't want us to close this ticket and feel like we are ready to support Mac, only to discover that we aren't.
If it fails on Mac 13 arm (Evergreen only have arm version right now), just create a ticket for it for now.

nbagnard (approved):
Code change looks good.
Just worried we might miss issues with MacOs 13. It would be better to know and track.

terakilobyte:
Recording here as well - MacOS 13 will require some work to get working in evergreen. A ticket has been created to track this work.

PR #164, opened 2023-11-30
SQL-1684: Add enums to support ODBC 2.x

jchemburkar:
- Turns sql states into a list of length 2 to store both odbc 2 and odbc 3 sql states
- add odbc 2 to odbc version enum
- error if odbc 2 is set in SQLSetEnvAttr

### DIFF
--- a/constants/src/lib.rs
+++ b/constants/src/lib.rs
@@ -25,7 +25,7 @@ lazy_static! {
 
 // SQL states, stored as [ODBC2 state, ODBC3 state]
 pub const NOT_IMPLEMENTED: [&str; 2] = ["S1C00", "HYC00"];
-pub const TIMEOUT_EXPIRED:[&str; 2] = ["S1T00", "HYT00"];
+pub const TIMEOUT_EXPIRED: [&str; 2] = ["S1T00", "HYT00"];
 pub const GENERAL_ERROR: [&str; 2] = ["S1000", "HY000"];
 pub const PROGRAM_TYPE_OUT_OF_RANGE: [&str; 2] = ["S1003", "HY003"];
 pub const INVALID_SQL_TYPE: [&str; 2] = ["S1004", "HY004"];
--- a/constants/src/lib.rs
+++ b/constants/src/lib.rs
@@ -23,32 +23,112 @@ lazy_static! {
     pub static ref DRIVER_ODBC_VERSION: String = format_driver_version();
 }
 
-// SQL states, stored as [ODBC2 state, ODBC3 state]
-pub const NOT_IMPLEMENTED: [&str; 2] = ["S1C00", "HYC00"];
-pub const TIMEOUT_EXPIRED: [&str; 2] = ["S1T00", "HYT00"];
-pub const GENERAL_ERROR: [&str; 2] = ["S1000", "HY000"];
-pub const PROGRAM_TYPE_OUT_OF_RANGE: [&str; 2] = ["S1003", "HY003"];
-pub const INVALID_SQL_TYPE: [&str; 2] = ["S1004", "HY004"];
-pub const INVALID_ATTR_VALUE: [&str; 2] = ["S1009", "HY024"];
-pub const INVALID_INFO_TYPE_VALUE: [&str; 2] = ["S1096", "HY096"];
-pub const NO_DSN_OR_DRIVER: [&str; 2] = ["IM007", "IM007"];
-pub const GENERAL_WARNING: [&str; 2] = ["01000", "01000"];
-pub const RIGHT_TRUNCATED: [&str; 2] = ["01004", "01004"];
-pub const OPTION_CHANGED: [&str; 2] = ["01S02", "01S02"];
-pub const FRACTIONAL_TRUNCATION: [&str; 2] = ["01S07", "01S07"];
-pub const UNABLE_TO_CONNECT: [&str; 2] = ["08001", "08001"];
-pub const INVALID_DESCRIPTOR_INDEX: [&str; 2] = ["07009", "07009"];
-pub const NO_RESULTSET: [&str; 2] = ["24000", "07005"];
-pub const RESTRICTED_DATATYPE: [&str; 2] = ["07006", "07006"];
-pub const INVALID_CURSOR_STATE: [&str; 2] = ["24000", "24000"];
-pub const FUNCTION_SEQUENCE_ERROR: [&str; 2] = ["S1010", "HY010"];
-pub const UNSUPPORTED_FIELD_DESCRIPTOR: [&str; 2] = ["S1091", "HY091"];
-pub const INVALID_ATTRIBUTE_OR_OPTION_IDENTIFIER: [&str; 2] = ["S1092", "HY092"];
-pub const INDICATOR_VARIABLE_REQUIRED: [&str; 2] = ["22002", "22002"];
-pub const INTEGRAL_TRUNCATION: [&str; 2] = ["22003", "22003"];
-pub const INVALID_DATETIME_FORMAT: [&str; 2] = ["22008", "22007"];
-pub const INVALID_CHARACTER_VALUE: [&str; 2] = ["22005", "22018"];
-pub const CONNECTION_NOT_OPEN: [&str; 2] = ["08003", "08003"];
+pub struct OdbcState<'a> {
+    pub odbc_2_state: &'a str,
+    pub odbc_3_state: &'a str,
+}
+
+pub const NOT_IMPLEMENTED: OdbcState<'static> = OdbcState {
+    odbc_2_state: "S1C00",
+    odbc_3_state: "HYC00",
+};
+pub const TIMEOUT_EXPIRED: OdbcState<'static> = OdbcState {
+    odbc_2_state: "S1T00",
+    odbc_3_state: "HYT00",
+};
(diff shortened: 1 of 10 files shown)


constants/src/lib.rs
```diff
@@ -23,32 +23,32 @@ lazy_static! {
...
+pub const RESTRICTED_DATATYPE: [&str; 2] = ["07006", "07006"];
+pub const INVALID_CURSOR_STATE: [&str; 2] = ["24000", "24000"];
+pub const FUNCTION_SEQUENCE_ERROR: [&str; 2] = ["S1010", "HY010"];
+pub const UNSUPPORTED_FIELD_DESCRIPTOR: [&str; 2] = ["S1091", "HY091"];
+pub const INVALID_ATTRIBUTE_OR_OPTION_IDENTIFIER: [&str; 2] = ["S1092", "HY092"];
+pub const INDICATOR_VARIABLE_REQUIRED: [&str; 2] = ["22002", "22002"];
+pub const INTEGRAL_TRUNCATION: [&str; 2] = ["22003", "22003"];
+pub const INVALID_DATETIME_FORMAT: [&str; 2] = ["22008", "22007"];
+pub const INVALID_CHARACTER_VALUE: [&str; 2] = ["22005", "22018"];
+pub const CONNECTION_NOT_OPEN: [&str; 2] = ["08003", "08003"];
```
terakilobyte, line 51: What do you think about expanding these into const structs? It might look something like:

```suggestion
pub struct OdbcVersion<'a> {
    pub not_implemented: &'a str,
    pub timeout_expired: &'a str,
    pub general_error: &'a str,
    pub program_type_out_of_range: &'a str,
    pub invalid_sql_type: &'a str,
    pub invalid_attr_value: &'a str,
    pub invalid_info_type_value: &'a str,
    pub no_dsn_or_driver: &'a str,
    pub general_warning: &'a str,
    pub right_truncated: &'a str,
    pub option_changed: &'a str,
    pub fractional_truncation: &'a str,
    pub unable_to_connect: &'a str,
    pub invalid_descriptor_index: &'a str,
    pub no_resultset: &'a str,
    pub restricted_datatype: &'a str,
    pub invalid_cursor_state: &'a str,
    pub function_sequence_error: &'a str,
    pub unsupported_field_descriptor: &'a str,
    pub invalid_attribute_or_option_identifier: &'a str,
    pub indicator_variable_required: &'a str,
    pub integral_truncation: &'a str,
    pub invalid_datetime_format: &'a str,
    pub invalid_character_value: &'a str,
    pub connection_not_open: &'a str,
}

pub const ODBC3_STATES: OdbcVersion<'static> = OdbcVersion {
    not_implemented: "HYC00",
    timeout_expired: "HYT00",
    general_error: "HY000",
    program_type_out_of_range: "HY003",
    invalid_sql_type: "HY004",
    invalid_attr_value: "HY024",
    invalid_info_type_value: "HY096",
    no_dsn_or_driver: "IM007",
    general_warning: "01000",
    right_truncated: "01004",
    option_changed: "01S02",
    fractional_truncation: "01S07",
    unable_to_connect: "08001",
    invalid_descriptor_index: "07009",
    no_resultset: "07005",
    restricted_datatype: "07006",
    invalid_cursor_state: "24000",
    function_sequence_error: "HY010",
    unsupported_field_descriptor: "HY091",
    invalid_attribute_or_option_identifier: "HY092",
    indicator_variable_required: "22002",
    integral_truncation: "22003",
    invalid_datetime_format: "22007",
    invalid_character_value: "22018",
    connection_not_open: "08003",
};

pub const ODBC2_STATES: OdbcVersion<'static> = OdbcVersion {
    not_implemented: "S1C00",
    timeout_expired: "S1T00",
    general_error: "S1000",
    program_type_out_of_range: "S1003",
    invalid_sql_type: "S1004",
    invalid_attr_value: "S1009",
    invalid_info_type_value: "S1096",
    no_dsn_or_driver: "IM007",
    general_warning: "01000",
    right_truncated: "01004",
    option_changed: "01S02",
    fractional_truncation: "01S07",
    unable_to_connect: "08001",
    invalid_descriptor_index: "07009",
    no_resultset: "24000",
    restricted_datatype: "07006",
    invalid_cursor_state: "24000",
    function_sequence_error: "S1010",
    unsupported_field_descriptor: "S1091",
    invalid_attribute_or_option_identifier: "S1092",
    indicator_variable_required: "22002",
    integral_truncation: "22003",
    invalid_datetime_format: "22008",
    invalid_character_value: "22005",
    connection_not_open: "08003",
};
```

Then, where they're used, it's something like
```
            | Error::QueryExecutionFailed(err) => {
                if matches!(err.kind.as_ref(), ErrorKind::Io(ref io_err) if io_err.kind() == std::io::ErrorKind::TimedOut)
                {
                    return ODBC3_STATES.timeout_expired;
                }
                ODBC3_STATES.general_error
            }
```

terakilobyte (approved):
LGTM, just a question/suggestion that might help avoid future mistakes.

constants/src/lib.rs
jchemburkar: agreed, I think this will look a lot cleaner!

constants/src/lib.rs
```diff
@@ -23,32 +23,112 @@ lazy_static! {
...
-pub const INVALID_ATTRIBUTE_OR_OPTION_IDENTIFIER: &str = "HY092";
-pub const INDICATOR_VARIABLE_REQUIRED: &str = "22002";
-pub const INTEGRAL_TRUNCATION: &str = "22003";
-pub const INVALID_DATETIME_FORMAT: &str = "22007";
-pub const INVALID_CHARACTER_VALUE: &str = "22018";
-pub const CONNECTION_NOT_OPEN: &str = "08003";
+pub struct OdbcState<'a> {
+    pub odbc_2_state: &'a str,
+    pub odbc_3_state: &'a str,
+}
```
terakilobyte, line 29: 🦸

constants/src/lib.rs
```diff
@@ -23,32 +23,112 @@ lazy_static! {
...
+pub const FRACTIONAL_TRUNCATION: OdbcState<'static> = OdbcState {
+    odbc_2_state: "01S07",
+    odbc_3_state: "01S07",
+};
+pub const UNABLE_TO_CONNECT: OdbcState<'static> = OdbcState {
+    odbc_2_state: "08001",
+    odbc_3_state: "08001",
+};
+pub const INVALID_DESCRIPTOR_INDEX: OdbcState<'static> = OdbcState {
+    odbc_2_state: "07009",
```
nbagnard, line 85: This one is going to be tricky. You might want to have 2 flavors of it.

> S1002  <-> 07009 	: ODBC 2.x SQLSTATE S1002 is mapped to ODBC 3.x SQLSTATE 07009 if the underlying function is SQLBindCol, SQLColAttribute, SQLExtendedFetch, SQLFetch, SQLFetchScroll, or SQLGetData.

We support at least SQLGetData, SQLFetch and SQLColAttribute.

nbagnard (changes requested):
The mapping for 07009 is not as straightforward and should be split. Then, you'll also need to update the functions to return the correct flavor 😥
I don't know why they made it so complicated.

constants/src/lib.rs
jchemburkar: Ah good catch. There's also

> ODBC 3.x SQLSTATE 07009 is mapped to ODBC 2.x SQLSTATE S1093 if the underlying function is SQLBindParameter or SQLDescribeParam.

Which doesn't apply to us, I missed that this wasn't the only special case for `07009`, thanks for flagging

constants/src/lib.rs
```diff
@@ -82,6 +82,10 @@ pub const UNABLE_TO_CONNECT: OdbcState<'static> = OdbcState {
     odbc_3_state: "08001",
 };
 pub const INVALID_DESCRIPTOR_INDEX: OdbcState<'static> = OdbcState {
+    odbc_2_state: "S1002",
+    odbc_3_state: "07009",
+};
+pub const INVALID_COLUMN_NUMBER: OdbcState<'static> = OdbcState {
```
jchemburkar, line 88: I split off a separate error, and made the existing error map to S1002 because it's used far more. This is what I found, let me know if this makes sense. INVALID_DESCRIPTOR_INDEX was used by two errors:

- `ColIndexOutOfBounds`: used in `get_value` functions for `MongoCollections`, `MongoDatabases`, `MongoFields`, `MongoTableTypes`, and `MongoTypesInfo`, which then only get called by `SQLGetData` (and thus should return `S1002`). Also used by `MongoStatement.get_col_metadata` which is called in `SQLColAttribute` and `SQLDescribeCol`, but I don't think the error is ever directly used
- `InvalidDescriptorIndex`: error that ends up getting thrown when `MongoStatement.get_col_metadata` does not return ok; updated to the new error for `SQLDescribeCol` so that it uses the 07009 sql state; use the existing error for `SQLColAttribute` so that the sql state is `S1002`

constants/src/lib.rs
nbagnard: For S1093, it is a DM error for SQLBindParameter but not always for SQLDescribeParam,.
I think that if we ever add support for parameters, we will miss this.
What about a comment on top of INVALID_COLUMN_NUMBER explaining that we are not a useless mapping to S1093, but that if SQLDescribeParam is added it would be needed.

PR #165, opened 2023-12-04
SQL-1685: add function to get odbc version from any mongo handle

jchemburkar:
This will enable functions to get the odbc version from any arbitrary handle passed in, i.e.,
```
let odbc_ver = mongo_handle.odbc_ver();
```

This can then be used to fork behavior based on odbc version -- namely in `SQLTables` and the `SQLGetDiag*` funcs.

### DIFF
--- a/odbc/src/handles/definitions.rs
+++ b/odbc/src/handles/definitions.rs
@@ -92,6 +92,31 @@ impl MongoHandle {
         }
     }
 
+    /// get the odbc_version from the underlying env handle, used to handle
+    /// behavior that is different between odbc versions properly
+    pub fn odbc_version(&mut self) -> OdbcVersion {
+        let env = match self {
+            MongoHandle::Env(_) => self,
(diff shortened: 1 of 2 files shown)


odbc/src/handles/definitions.rs
```diff
@@ -92,6 +92,31 @@ impl MongoHandle {
         }
     }
 
+    /// get the odbc_version from the underlying env handle, used to handle
+    /// behavior that is different between odbc versions properly
+    pub fn odbc_version(&mut self) -> OdbcVersion {
```
terakilobyte, line 97:
```suggestion
    pub fn get_odbc_version(&mut self) -> OdbcVersion {
```

terakilobyte (approved):
LGTM. Just had a thought about naming.

nbagnard (approved):
LGTM once Nathan's suggestion is incorporated :)

PR #166, opened 2023-12-04
SQL-1794: Enable ASAN for integration tests

bucaojit:
Using `export RUSTFLAGS="-Z sanitizer=leak"` to enable ASAN for integration tests.  
The flag `"-Z sanitizer=address"` is more comprehensive but it was not compatible with the `serde_derive` crate and was throwing the following error:
```
[2023/11/22 12:13:38.048] error: /data/mci/97459366b05ed4c1edbb3e6d812cc9a7/mongosql-odbc-driver/target/debug/deps/libserde_derive-991a11b9bfb51463.so: undefined symbol: __asan_option_detect_stack_use_after_return
```

~~Tested that it does fail on memory leak by introducing a leak in SQLTablesW:~~
```
let leaked_memory = mem::transmute::<Box<i32>, *mut i32>(Box::new(42));
// also trying
let _leaked_memory = Box::leak(Box::new(42));
```
~~And confirming that the test failed due to the memory leak.~~
https://spruce.mongodb.com/task/mongosql_odbc_driver_ubuntu2204_integration_test_patch_3aaaf53a8a48ccf277531f718698ab5f0afcd915_656a80dd2fbabecf068c0a05_23_12_02_00_57_19/logs?execution=0

---

After fixing the memory leaks in the integration tests, I was not able to capture the memory leak in the library.  The failures I was hitting earlier were from memory leaks in the `result-set tests` and the `connection_tests`.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -1445,8 +1445,10 @@ tasks:
         variants: [macos, macos-arm]
       - func: "run ubuntu integration tests"
         variants: [ubuntu2204]
-      - func: "run asan integration tests"
-        variants: [ ubuntu2204 ]
+      # Commenting out because the following task only detects 
+      # memory leaks in the tests
+      # - func: "run asan integration tests"
+      #   variants: [ ubuntu2204 ]
       - func: "run windows integration tests"
         variants: [windows-64]
       - func: "run macos integration tests"
@@ -1485,8 +1487,10 @@ tasks:
         variants: [ubuntu2204]
       - func: "run macos result-set tests"
         variants: [macos, macos-arm]
-      - func: "run asan result-set tests"
-        variants: [ ubuntu2204 ]
+      # Commenting out because the following task only detects 
+      # memory leaks in the tests
+      # - func: "run asan result-set tests"
+      #   variants: [ ubuntu2204 ]
 
   - name: sign
     depends_on:
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -286,30 +286,6 @@ functions:
       params:
         file: mongosql-odbc-driver/expansions.yml
 
-  "compile release with debug info and asan flag":
-    - command: shell.exec
-      params:
-        shell: bash
-        working_dir: mongosql-odbc-driver
-        script: |
-          ${prepare_shell}
-          export RUSTFLAGS="-Z sanitizer=leak"
-          echo "env $(env)"
-          
-          ~/.cargo/bin/rustup default nightly
-          ~/.cargo/bin/rustup target add x86_64-unknown-linux-gnu
-          
-          # Build a release driver with debug information
-          cargo build --release
-          export DRIVER_LIB_PATH=$PWD/target/release
-
-          cat <<EOT >> expansions.yml
-            export DRIVER_LIB_PATH="$DRIVER_LIB_PATH"
-          EOT
-    - command: expansions.update
-      params:
-        file: mongosql-odbc-driver/expansions.yml
-
   "compile macos release":
     - command: shell.exec
       type: test
@@ -1216,7 +1192,13 @@ functions:
           echo "ODBCSYSINI = $ODBCSYSINI"
           echo "----------------------------------------------"
 
+          export RUSTFLAGS="-Z sanitizer=leak"
           export RUST_BACKTRACE=1
+          echo "env $(env)"
+          
+          ~/.cargo/bin/rustup default nightly
+          ~/.cargo/bin/rustup target add x86_64-unknown-linux-gnu
+          
           # Start a local ADF
           ./resources/run_adf.sh start && cargo run --bin data_loader &&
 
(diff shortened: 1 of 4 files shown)


evergreen.yml
```diff
@@ -286,6 +286,30 @@ functions:
...
         file: mongosql-odbc-driver/expansions.yml
 
+  "compile release with debug info and asan flag":
+    - command: shell.exec
+      params:
+        shell: bash
+        working_dir: mongosql-odbc-driver
+        script: |
+          ${prepare_shell}
+          export RUSTFLAGS="-Z sanitizer=leak"
```
nbagnard, line 296: Did you try something else than "leak", we use "address" on unit tests and it seems that it would useful too.
From the [doc](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html), "leak" is already on by default on Linux.

bucaojit: I tried with `address` as we have in the unit tests, and with `memory`.  
They both fail, "address" causes an undefined symbol error for one of the crates that we import `serde_derive`, although it isn't imported for the unit tests which is why our unit tests don't hit the same issue. 
```
undefined symbol: __asan_option_detect_stack_use_after_return
```
The `memory` flag fails in another crate called `futures-task`:
   Compiling futures-task v0.3.28

Adding the `sanitizer=leak` flag did cause the integration tests to fail with the `ERROR: LeakSanitizer: detected memory leaks` errors which we don't see without the flag set so I would think it is testing something different than what is default.  

I updated the description to say that after fixing the integration tests, I was not able to get a test failure from my introduced memory leak.  I incorrectly thought the error was due to the changes in the library but it was in the tests themselves.  I confirmed that the code to add the leak is run by adding logging, it's not getting caught unfortunately.

EthanHardyMongo: > I was not able to get a test failure from my introduced memory leak.

Are you completely sure that you were actually causing a memory leak? I just want to make sure because that seems kind of odd haha. Also, in the description you wrote, `After fixing the memory leaks in the integration tests, I was not able to capture the memory leak in the library.` So how do we know there is a memory leak in the library if we can't capture it?

bucaojit: Yes, if I put the same leak code in the integration test code (integration_test/src/test_runner/mod.rs) or the connection test it will detect the memory leak.  The `Box::leak` method will take ownership of a Box and prevents it from being dropped.  I added that in a few functions in `odbc/src/api/functions.rs` and added logging before and after to confirm the leak code was being executed. 

>So how do we know there is a memory leak in the library if we can't capture it?
The integration tests have memory leaks that are fixed by this PR.  I injected the memory leak with `Box::leak` in the driver/library code and was running the test with the `RUSTFLAGS="-Z sanitizer=leak"` flag expecting for it to detect the memory leak but it didn't.  So we don't know if there is a memory leak in the existing driver code but we also can't detect it through the integration test.  The leak tester does run with the unit tests so that's how we will confirm there are no leaks.

evergreen.yml
```diff
@@ -1227,7 +1229,7 @@ functions:
 
           ~/.cargo/bin/rustup default nightly
           ~/.cargo/bin/rustup target add x86_64-unknown-linux-gnu
-          export RUSTFLAGS="-Z sanitizer=address"
+          export RUSTFLAGS="-Z sanitizer=leak"
```
nbagnard, line 1232: I think it is important to assess that we don't overflow buffer or do out of bound access just as much as making sure that we don't have memory leaks.
I thought that since leak is already on by default on Linux, setting the flag to "address" would mean that we cover both, but looking at the rust doc, it looks like only one sanitizer at a time is possible.
Sorry to be a pain, but could we have 2 tasks?
- Run-time memory leak detection (aka run result set test with sanitizer=leak)
- Memory error detection (only need to compile the driver if I understand correctly)
To be fair, we might even need 3 with memory for even better coverage but this is outside of the scope for this and I think we can cover this better when dealing with SSDLC

nbagnard (changes requested):
Thank you for looking into the leaks in the integration tests, no leaks there is nice :)
I have a request for also making sure that we have a good coverage catching memory related error as early as possible.

evergreen.yml
EthanHardyMongo: Currently, both ASAN functions (result set and integration) are commented out because they only test for leaks in the tests. Are you saying that we uncomment those and then add another task that just does memory error detection?

nbagnard: Leak is a runtime detection. It should be used in conjunction with the unit tests.
For Address it is a compile time detection, so we just need to build the code with the flag.
So, I think we want to change the sanitizier to leak on "run asan unit tests" and have a new task in "Static code analysis" which build the driver with sanitizer=address.
Doe that make more sense?

EthanHardyMongo: Yes sounds good!

bucaojit: Okay, if I understand correctly we want to run the unit test with `sanitizer=leak` and add a task that builds the driver with `sanitizer=address`.  I added `asan-compile` and it runs cargo build with `sanitizer=address` with the debug build.  I can add --release if we think it will help.  I was able to get it to compile without the error I mentioned in the other thread by adding `--target x86_64-unknown-linux-gnu` to the build command.

nbagnard: Yes, you need `--target x86_64-unknown-linux-gnu` just a funny quirk of Rust. I think it's because it's still a nigthtly feature

PR #169, opened 2023-12-07
SQL-1807: Comment bson_type_info, fix two issues with double

pmeredit:


core/src/bson_type_info.rs
```diff
@@ -70,7 +97,7 @@ impl BsonTypeInfo {
         non_concise_type: SqlDataType::DOUBLE,
         searchable: SQL_PRED_BASIC,
         is_case_sensitive: false,
-        fixed_prec_scale: true,
```
pmeredit, line 73: fixed_prec_scale definitely needs to be false for DOUBLE

core/src/bson_type_info.rs
```diff
@@ -80,7 +107,7 @@ impl BsonTypeInfo {
         sql_code: None,
         is_auto_unique_value: Some(false),
         is_unsigned: Some(false),
-        num_prec_radix: Some(10),
+        num_prec_radix: Some(2),
```
pmeredit, line 110: num_prec_radix is supposed to be 2 for DOUBLES as I read the spec. However, the way we are listing precision here is in number of decimal digits, so I'm not 100% sure on this change, unlike the other

pmeredit: The correct answer may actually be to change the precision to be in bits

nbagnard: I think 15 for the precision is correct here.
See SQL_DOUBLE in [https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/column-size?view=sql-server-ver16](https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/column-size?view=sql-server-ver16)

nbagnard (approved):
LGTM :)

jchemburkar (approved):
lgtm!

PR #172, opened 2023-12-15
SQL-1690: enable odbc 2 to be set in SQLSetEnvAttr

jchemburkar:
Followed a similar structure to power bi's flow. This is lightly borrowed but not directly taken from any particular odbc 2 flow, because they tend to do a lot of stuff in the middle (i.e., alloc the env handle, then make calls like SQLDrivers, SQLDataSources, SQLFunctions, etc) that weren't what we want to test here.

I instead tried hitting on the core odbc 2 behaviors (sql tables, sql get type info, connecting + getting data), with the SQLColumns to come as part of SQL-1688

### DIFF
--- a/integration_test/tests/odbc_2_tests.rs
+++ b/integration_test/tests/odbc_2_tests.rs
@@ -5,10 +5,10 @@ mod integration {
         fetch_and_get_data, generate_default_connection_str, get_sql_diagnostics, BUFFER_LENGTH,
     };
     use odbc_sys::{
-        CDataType, DriverConnectOption, EnvironmentAttribute, HDbc, HEnv, HStmt, Handle,
-        HandleType, Len, Pointer, SQLAllocHandle, SQLDriverConnectW, SQLExecDirectW, SQLFetch,
-        SQLGetData, SQLGetTypeInfo, SQLSetEnvAttr, SQLTablesW, SmallInt, SqlDataType, SqlReturn,
-        NTS,
+        CDataType, ConnectionAttribute, DriverConnectOption, EnvironmentAttribute, HDbc, HEnv,
+        HStmt, Handle, HandleType, Len, Pointer, SQLAllocHandle, SQLDriverConnectW, SQLExecDirectW,
+        SQLFetch, SQLGetData, SQLGetTypeInfo, SQLSetConnectAttrW, SQLSetEnvAttr, SQLTablesW,
+        SmallInt, SqlDataType, SqlReturn, NTS,
     };
 
     use cstr::WideChar;
@@ -56,6 +56,22 @@ mod integration {
                     &mut dbc as *mut Handle
                 )
             );
+
+            // Causes iODBC to hang on `SetConnectOptionW` call
+            #[cfg(not(target_os = "macos"))]
+            {
+                // Set the login timeout
+                let login_timeout = 15;
+                assert_eq!(
+                    SqlReturn::SUCCESS,
+                    SQLSetConnectAttrW(
+                        dbc as HDbc,
+                        ConnectionAttribute::LoginTimeout,
+                        login_timeout as Pointer,
+                        0,
+                    )
+                );
+            }
 
             in_connection_string = generate_default_connection_str();
             let in_connection_string_encoded = cstr::to_widechar_vec(&in_connection_string);
@@ -120,8 +136,6 @@ mod integration {
     #[test]
     fn test_list_tables() {
         let env_handle = setup();
-        connect(env_handle);
-        let env_handle: HEnv = setup();
         let (conn_handle, _, _, _) = connect(env_handle);
         let mut stmt: Handle = null_mut();
 
@@ -195,8 +209,6 @@ mod integration {
     #[test]
     fn test_type_listing() {
         let env_handle = setup();
-        connect(env_handle);
-        let env_handle: HEnv = setup();
         let (conn_handle, _, _, _) = connect(env_handle);
         let mut stmt: Handle = null_mut();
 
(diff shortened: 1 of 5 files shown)


terakilobyte:
The code overall looks fine to me, but it's troubling that it's timing out specifically on macos

jchemburkar:
@terakilobyte agreed, but seems power bi ran into similar issues. This has a similar (but pared down) structure, will double check if there's anything we might need to copy over from that

integration_test/tests/odbc_2_tests.rs
```diff
@@ -0,0 +1,318 @@
...
+                    HandleType::Stmt,
+                    conn_handle as *mut _,
+                    &mut stmt as *mut Handle
+                )
+            );
+
+            // check that when requesting all types, both odbc 2 and 3 timestamp types are returned
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLGetTypeInfo(stmt as HStmt, SqlDataType::UNKNOWN_TYPE)
```
nbagnard, line 230: Can you please add SQL_ALL_TYPES in definition or in common?
That both UNKNOWN_TYPE and SQL_ALL_TYPES identifiers are `0` is good and probably a little bit on purpose, but having the type called SQL_ALL_TYPES is much clearer.

Note : It is also defined in sql.h. Both are.

integration_test/tests/odbc_2_tests.rs
```diff
@@ -0,0 +1,318 @@
...
+        unsafe {
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLAllocHandle(
+                    HandleType::Stmt,
+                    conn_handle as *mut _,
+                    &mut stmt as *mut Handle
+                )
+            );
+            assert_eq!(
```
nbagnard, line 294: Why do you check that you can retrieve all type info again in here if what you are trying to validate is the type mapping in SQLGetData?

integration_test/tests/odbc_2_tests.rs
```diff
@@ -0,0 +1,318 @@
...
+                "{}",
+                get_sql_diagnostics(HandleType::Stmt, stmt as Handle)
+            );
+
+            // fetch date field as odbc 2.x TimeStamp
+            fetch_and_get_data(
+                stmt,
+                Some(5),
+                vec![SqlReturn::SUCCESS; 1],
+                vec![CDataType::TimeStamp],
```
nbagnard, line 314: The C type didn't change between version right?
I think what you want to check is that you are getting the correct SQL type for the column in the resultset by calling SQLColAttribute with SQL_DESC_TYPE

nbagnard (changes requested):
Just a couple of things, otherwise it's in good shape

integration_test/tests/odbc_2_tests.rs
```diff
@@ -0,0 +1,318 @@
...
+                "{}",
+                get_sql_diagnostics(HandleType::Env, env_handle as Handle)
+            );
+
+            // assert all tables are returned from the previous SQLTables call
+            for _ in 0..14 {
+                assert_eq!(SqlReturn::SUCCESS, SQLFetch(stmt as HStmt));
+            }
+
+            assert_eq!(SqlReturn::NO_DATA, SQLFetch(stmt as HStmt))
```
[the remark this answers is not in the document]
jchemburkar: Yes, I can explicitly free here. However I remember this coming up in the power bi tests as well, and we found it wasn't complaining / we decided not to free there. If I add this helper do you think I should add it to those functions as well?

integration_test/tests/odbc_2_tests.rs
```diff
@@ -0,0 +1,318 @@
...
+    #[test]
+    fn test_connect() {
+        let env_handle = setup();
+        connect(env_handle);
+    }
+
+    /// Test list_tables, which should yield all tables
+    #[test]
+    fn test_list_tables() {
+        let env_handle = setup();
```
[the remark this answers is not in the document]
jchemburkar: yes! thanks for pointing that out.

integration_test/tests/odbc_2_tests.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    /// Test Setup flow
+    #[test]
+    fn test_setup() {
+        setup();
+    }
+
+    /// Test connect
+    #[test]
+    fn test_connect() {
```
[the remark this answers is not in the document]
jchemburkar: This was just testing that creating the connection handle doesn't panic basically. But I agree, maybe we can merge `test_setup` into this because they are so overlapping

integration_test/tests/odbc_2_tests.rs
jchemburkar: Just a bad copy/paste of the previous function when I was building off of the statement allocation 😓

integration_test/tests/odbc_2_tests.rs
jchemburkar: yes, you are right about sqlcolumns, and that's partially why this pr went up, so SQLColumns can be added as part of 1688. I can update the comment to be a little clearer

terakilobyte (approved):
LGTM. We'll figure out the iODBC stuff soon enough. This will help unblock me on 1688

integration_test/tests/odbc_2_tests.rs
nbagnard: What I mean is that this test is kind of useless because there was not change affecting getData as far as I can tell. We go straight from bson data to C data and we skip any kind of SQL mapping.

jchemburkar: I see what you mean now, that makes sense about the bson -> c conversion, and you are right that we didn't update the c types. I can remove it then

nbagnard: Cool, thank you!

nbagnard (approved):
LGTM!

PR #173, opened 2024-01-08
SQL-1688: ODBC2 behavior support for SQLColumns, SQLDescribeCol

terakilobyte:
This adds odbc2 behavioral support for select functions. I used the relevant queries that show the date type being converted properly.


core/src/fields.rs
```diff
@@ -459,12 +459,23 @@ pub struct MongoFields {
...
 // Statement related to a SQLTables call.
 // The Resultset columns are hard-coded and follow the ODBC resultset for SQLColumns :
 // TABLE_CAT, TABLE_SCHEMA, TABLE_NAME, TABLE_TYPE.
 impl MongoFields {
+    /// Whether to map the TIMESTAMP type (93) to EXT_TIMESTAMP (11) for odbc 2. Maps the type if `odbc_3_types` is `false`
+    /// AND the data_type is SqlDataType::TIMESTAMP, otherwise, this is an identity function.
+    /// See https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/datetime-data-type-changes?view=sql-server-ver16 for more information.
+    pub fn map_type_for_odbc_version(odbc_3_types: bool, data_type: SqlDataType) -> SqlDataType {
+        match (odbc_3_types, data_type) {
+            (false, SqlDataType::TIMESTAMP) => SqlDataType::EXT_TIMESTAMP,
```
pmeredit, line 475: I like this

integration_test/src/test_runner/test_generator_util.rs
```diff
@@ -156,7 +156,7 @@ fn get_expected_data_type(sql_type: &Value) -> CDataType {
                 SqlDataType::EXT_W_CHAR => CDataType::WChar,
                 SqlDataType::EXT_W_VARCHAR => CDataType::WChar,
                 SqlDataType::EXT_W_LONG_VARCHAR => CDataType::WChar,
-                SqlDataType::EXT_GUID => CDataType::Guid,
```
pmeredit, line 159: If we don't want this here, we should probably delete the line.

pmeredit (approved):
LGTM, probably delete that comment line

integration_test/src/test_runner/test_generator_util.rs
terakilobyte: I think we want it

terakilobyte:
@jchemburkar forgot to ping you on this review!

odbc/src/api/col_attr_describe_tests.rs
```diff
@@ -8,17 +8,30 @@ use std::sync::RwLock;
 
 mod unit {
```
terakilobyte, line 9: this file is mostly just updates to provide an env and conn handle so we don't panic

integration_test/tests/connection_tests.rs
```diff
@@ -107,6 +107,7 @@ mod integration {
     // At connection time, when the connection string is processed, if there is a log level specified
     // the logger log level is updated to the connection log level.
     #[test]
+    #[ignore]
```
terakilobyte, line 110: we're skipping this test in evergreen, so I added a skip here as well. I think we should probably open a ticket to look into this, or see if we even want to try to test this going forward.

terakilobyte (commented):
@pmeredit requesting again since there are more changes.

odbc/src/api/col_attr_describe_tests.rs
```diff
@@ -8,17 +8,30 @@ use std::sync::RwLock;
...
+    use crate::{
+        handles::definitions::{Connection, ConnectionState, Env, EnvState},
+        SQLSetEnvAttr,
+    };
 
     use super::*;
     // test unallocated_statement tests SQLColAttributeW when the mongo_statement inside
     // of the statement handle has not been allocated (before an execute or tables function
     // has been called).
+
```
pmeredit, line 22: Remove this new line since the comment is actually for the test. I'm not sure what happened to suggestion but it has disappeared from the UI.

pmeredit (changes requested):
```[2024/01/10 16:14:09.425] called `Result::unwrap()` on an `Err` value: DriverConnect("SUCCESS_WITH_INFO", "[unixODBC][Driver Manager]Driver does not support the requested version")```
Which version is it requesting here? It's working fine on windows but not on ubuntu. Maybe the ubuntu DM doesn't support 2.0? If it doesn't, we'll need to clean up the tests to not run 2.0 tests on Ubuntu.

**Comment by jchemburkar** [comment]
@pmeredit noted this in my 1850 pr, but looks like the periodic build has been failing since just before new years (with no code changes) leading to that error message for result set tests on ubuntu (doesn't explain integration tests however). I also spun up a new ubuntu box and couldn't repro 🤷

**Comment by terakilobyte** [comment]
@jchemburkar @pmeredit I see nothing in the [release notes](https://www.unixodbc.org/) that would suddenly cause this error

File: odbc/src/api/col_attr_describe_tests.rs [hunk 6]
```diff
@@ -422,19 +513,183 @@ mod unit {
...
             }
         }
+        unsafe {
+            let _ = Box::from_raw(conn as *mut WChar);
+            let _ = Box::from_raw(env as *mut WChar);
+        }
+    }
+
+    #[test]
+    fn test_numeric_field_attributes_odbc_2() {
```
jchemburkar, line 534: similar q -- if odbc 2 only effects datetime types + this has the same expectations, we might be able to remove this test case

jchemburkar (changes requested):
Main logic is looking good! had a couple questions on testing

odbc/src/api/functions.rs
```diff
@@ -1309,6 +1313,7 @@ fn sql_free_handle(handle_type: HandleType, handle: *mut MongoHandle) -> Result<
 #[named]
 #[no_mangle]
 pub unsafe extern "C" fn SQLFreeStmt(statement_handle: HStmt, _option: SmallInt) -> SqlReturn {
+    error!("option is {_option}");
```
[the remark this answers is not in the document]
terakilobyte: Oh, it's not, good catch!

odbc/src/api/col_attr_describe_tests.rs
```diff
@@ -422,19 +513,183 @@ mod unit {
...
+            );
+            let _ = Box::from_raw(name_buffer as *mut WChar);
+            let _ = Box::from_raw(conn as *mut WChar);
+            let _ = Box::from_raw(env as *mut WChar);
+        }
+    }
+
+    // check the describe output for odbc version 2
+    #[test]
+    fn test_describe_col_odbc_2() {
```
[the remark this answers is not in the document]
terakilobyte: I'm happy removing these. I was iffy on adding them, they were more to show that odbc 2 behavior does *not* impact.

odbc/src/api/col_attr_describe_tests.rs
terakilobyte: ditto, will remove

odbc/src/api/col_attr_describe_tests.rs
```diff
@@ -65,14 +77,24 @@ mod unit {
                 let _ = Box::from_raw(char_buffer as *mut WChar);
             }
         }
+        unsafe {
+            let _ = Box::from_raw(conn as *mut WChar);
+            let _ = Box::from_raw(env as *mut WChar);
```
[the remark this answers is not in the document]
terakilobyte: I don't think we need to do this with the statement since it isn't an allocated raw pointer, but perhaps I'm thinking of something else.

terakilobyte:
result-set-tests are disabled on macos, and the odbc2 result-set-test is disabled on ubuntu until we can figure out what's going on with evergreen unixodbc

odbc/src/api/col_attr_describe_tests.rs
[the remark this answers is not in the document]
jchemburkar: I think youre right (we know this wont leak / rust will clean it up). I think we do waffle a little bit on explicit freeing of handles / cleanup, but i'm ok with leaving as is 👍

jchemburkar (approved):
changes look good to me. Nice work on the updated resultset runner!

pmeredit (approved):
LGTM

terakilobyte:
Merging with failing test as it's unrelated to the changes.

PR #174, opened 2024-01-09
SQL-1850: add LogLevel to dsn iterator

jchemburkar:
This allows the log level to be written / read properly, so that the odbc configuration (set via dsn, or registry file) will be used by the driver.

Ubuntu tests are failing, but appears to not be related -- periodic build started failing 12/29: https://spruce.mongodb.com/variant-history/mongosql-odbc-driver/ubuntu2204?selectedCommit=521

### DIFF
--- a/shared_sql_utils/src/dsn.rs
+++ b/shared_sql_utils/src/dsn.rs
@@ -13,7 +13,6 @@ const SERVER: &str = "server";
 const UID: &str = "uid";
 const URI: &str = "uri";
 const USER: &str = "user";
-const LOGLEVEL: &str = "loglevel";
 // SQL-1281
 // const LOGPATH: &str = "LOGPATH";
 
@@ -54,7 +53,6 @@ pub struct Dsn {
     pub user: String,
     pub server: String,
     pub driver_name: String,
-    pub log_level: Option<String>,
 }
 
 impl Dsn {
@@ -78,7 +76,6 @@ impl Dsn {
                 user: args.user.into(),
                 server: args.server.into(),
                 driver_name: args.driver_name.into(),
-                log_level: None,
             })
         } else if !validation[1] {
             Err(DsnError::Dsn(args.dsn.into()))
@@ -207,7 +204,6 @@ impl Dsn {
             URI => self.uri = value.to_string(),
             USER => self.user = value.to_string(),
             UID => self.user = value.to_string(),
-            LOGLEVEL => self.log_level = Some(value.to_string()),
             // SQL-1281
             // LOGPATH => self.logpath = value.to_string(),
             _ => {}
--- a/shared_sql_utils/src/dsn.rs
+++ b/shared_sql_utils/src/dsn.rs
@@ -249,6 +249,7 @@ impl<'a> DSNIterator<'a> {
                 ("Password", &dsn_opts.password),
                 ("Uri", &dsn_opts.uri),
                 ("User", &dsn_opts.user),
+                ("LogLevel", &dsn_opts.log_level),
                 // SQL-1281
                 // ("Logpath", &dsn_opts.logpath),
             ],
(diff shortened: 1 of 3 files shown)


terakilobyte (approved):
LGTM. Will dsn loglevel override driver specified loglevel?

jchemburkar:
> LGTM. Will dsn loglevel override driver specified loglevel?

It should. The way I tested was 
1. just set up the DSN using the registry file (so no log level will be set) and connect to tool. This just gave info level logs
2. update the dsn via the odbc configuration gui to set log level to debug
3. re-connect to the tool -- we then got debug logs as expected

We have already seen that odbc uri works for setting debug logs so we know that should take precedence over both

shared_sql_utils/src/dsn.rs
```diff
@@ -249,6 +249,7 @@ impl<'a> DSNIterator<'a> {
                 ("Password", &dsn_opts.password),
                 ("Uri", &dsn_opts.uri),
                 ("User", &dsn_opts.user),
+                ("LogLevel", &dsn_opts.log_level),
```
nbagnard, line 252: Did you consider removing the log level from the DSN configuration GUI instead?

I didn't realize/forgot we have a field for log level in the DSN config GUI.
The only reason I didn't remove it from the URI support is because I don't want to break older PowerBI or even ODBC driver installation.
However, I am not a big fan of DSN/connection level configuration which are overriding other connections properties. It makes things hard to document and frustrating for users.
Also, since we would only allow Driver level logging for PowerBI, it might be better to standardize this.
Why was setting the driver log level at the Driver level not enough?

jchemburkar: I had not considered removing it, just noticed this was not working when doing compatibility testing so checked out out and found it was a simple fix. If it would be more sustainable / a better long term solution however, I'd be happy to do what you are suggesting

terakilobyte: I'm curious, what tools support multiple different simultaneous DSN configurations? If this is a thing, we may want to support adding the dsn name as a key in the log.

nbagnard: I'd expect most tools support multiple connections. ETL or SQL IDEs will for sure.  I haven't tested the tools. You guys should have a better idea for the tools we tested. Generally, if the tools also you to have multiple source for your report, then it has to allow multiple connection.

We have [SQL-1795](https://jira.mongodb.org/browse/SQL-1795) for supporting connection log level properly. I think it might be better to just tackle it at that time.

Side question for @terakilobyte : What would add the dsn key in the log do? How/where do you think about using it? Do you mean logging the DSN name for each line if we connected through a DSN?

terakilobyte: @nbagnard Yeah that's my thought, though the limited subset of tools I tested didn't have simultaneous connections. However, if this is common, it may be something to think about.

jchemburkar:
I think this achieves what we've talked about in pieces, removing the log level from the gui and just allowing it to be set via driver settings or connection string. In testing it seems to behave as expected, with the driver settings being the default value, and connection strings able to update/override this value

nbagnard (approved):
LGTM

shared_sql_utils/src/dsn.rs
```diff
@@ -55,7 +54,7 @@ pub struct Dsn {
     pub user: String,
     pub server: String,
     pub driver_name: String,
-    pub log_level: String,
+    pub log_level: Option<String>,
```
terakilobyte, line 57: Hmm.. do we need any notion of loglevel in dsn now?

terakilobyte (commented):
LGTM, but leaving as a comment to first ask if we even need loglevel support in the dsn?

shared_sql_utils/src/dsn.rs
jchemburkar: That's a good point -- I wasn't sure since the two are interlinked in some places, but seems like setting through the odbc uri works fine even if its removed from dsn. Made this change 👍

terakilobyte (approved):

PR #175, opened 2024-01-18
SQL-1859: Fix ubuntu issue by giving all enums with variants well defined sizes.

pmeredit:
The DM passes uninitialzed memory to functions like SQLGetEnvAttr, so we need to get the size correct in order to not have tainted results.

### DIFF
--- a/integration_test/tests/power_bi_tests.rs
+++ b/integration_test/tests/power_bi_tests.rs
@@ -144,7 +144,7 @@ mod integration {
             in_connection_string_encoded.push(0);
 
             let str_len_ptr = &mut 0;
-            const BUFFER_LENGTH: SmallInt = 300;
+            const BUFFER_LENGTH: SmallInt = 1024;
             let mut out_connection_string_buff: [WideChar; BUFFER_LENGTH as usize - 1] =
                 [0; (BUFFER_LENGTH as usize - 1)];
             let out_connection_string_buff = &mut out_connection_string_buff as *mut WideChar;
--- a/odbc/src/api/definitions.rs
+++ b/odbc/src/api/definitions.rs
@@ -66,7 +66,7 @@ pub enum Concurrency {
 }
 
 #[derive(Clone, Copy, Debug, FromPrimitive)]
-#[repr(usize)]
+#[repr(i32)]
 pub enum CursorType {
     ForwardOnly = 0,
     KeysetDriven = -1,
--- a/odbc/src/api/definitions.rs
+++ b/odbc/src/api/definitions.rs
@@ -26,7 +26,7 @@ pub enum ConnectionPooling {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, FromPrimitive)]
-#[repr(i32)]
+#[repr(u32)]
 pub enum CpMatch {
     Strict = 0,
     Relaxed,
--- a/odbc/src/api/definitions.rs
+++ b/odbc/src/api/definitions.rs
@@ -50,7 +50,7 @@ pub enum CursorSensitivity {
 }
 
 #[derive(Clone, Copy, Debug)]
-#[repr(i32)]
+#[repr(u32)]
 pub enum AsyncEnable {
     Off = 0,
     On,
--- a/odbc/src/api/definitions.rs
+++ b/odbc/src/api/definitions.rs
@@ -82,7 +82,7 @@ pub enum NoScan {
 }
 
 #[derive(Clone, Copy, Debug)]
-#[repr(i32)]
+#[repr(u32)]
 pub enum BindType {
     BindByColumn = 0,
 }
--- a/odbc/src/api/definitions.rs
+++ b/odbc/src/api/definitions.rs
@@ -66,12 +66,12 @@ pub enum Concurrency {
 }
 
 #[derive(Clone, Copy, Debug, FromPrimitive)]
-#[repr(i32)]
+#[repr(u32)]
 pub enum CursorType {
     ForwardOnly = 0,
-    KeysetDriven = -1,
-    Dynamic = -2,
-    Static = -3,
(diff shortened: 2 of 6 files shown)


odbc/src/api/functions.rs
```diff
@@ -956,14 +956,15 @@ pub unsafe extern "C" fn SQLDriverConnectW(
             let mongo_connection =
                 odbc_unwrap!(sql_driver_connect(conn, &odbc_uri_string), conn_handle);
             *conn.mongo_connection.write().unwrap() = Some(mongo_connection);
-            let buffer_len = usize::try_from(buffer_length).unwrap();
             // We know the mysql ODBC driver returns SUCCESS if the out_connection_string is NULL.
             // We can also just return SUCCESS if the buffer_len is 0. Likely, users are not
             // expecting to get back a warning when they pass an empty buffer to this, especially
             // given that we only currently support DriverConnectOption::NoPrompt.
```
pmeredit, line 962: This is just a defensive programming change that is worth keeping.

integration_test/tests/power_bi_tests.rs
```diff
@@ -144,7 +144,7 @@ mod integration {
             in_connection_string_encoded.push(0);
 
             let str_len_ptr = &mut 0;
```
pmeredit, line 146: Documentation suggests the buffer size should be 1024 minimum, shouldn't have actually be a problem, however.

odbc/src/api/definitions.rs
```diff
@@ -1,20 +1,23 @@
...
 #[derive(Copy, Clone, Debug, PartialEq, Eq, FromPrimitive)]
+#[repr(i32)]
 pub enum SqlBool {
     False = 0,
     True,
 }
 
 // Environment attributes
 #[derive(Copy, Clone, Debug, PartialEq, Eq, FromPrimitive)]
+#[repr(i32)]
```
nbagnard, line 12: Not that it matter size wise, but in the ODBC headers, the values are all unsigned long. Maybe to match, we should map it to u32 (the spec says i32, but the code in this case seems a better of truth to me)?

odbc/src/api/definitions.rs
```diff
@@ -98,11 +110,13 @@ pub enum RetrieveData {
 pub enum RowOperationPtr {}
 
 #[derive(Clone, Copy, Debug)]
+#[repr(i32)]
 pub enum SimulateCursor {
     NonUnique = 0,
 }
 
 #[derive(Clone, Copy, Debug, FromPrimitive)]
+#[repr(i32)]
```
nbagnard, line 119: The specs says SQLULEN, but the values in the ODBC headers (Windows, unixodbc, iodbc) are unsigned int.
```suggestion
#[repr(u32)]
```

nbagnard (changes requested):
Thank you for figuring out the root cause of the issue and also for adding the buffer length safe guard in SQLDriverConnect!
However, I think that we realized that the correct mapping matters and I went over the spec and ODBC headers for all the attributes below and the mapping you picked for most of them don't match. How did you pick the mapping? 
The ones where you used i32 instead of usize would create the same kind of issue when the DM or application would read it.

odbc/src/api/definitions.rs
```diff
@@ -31,25 +35,29 @@ pub enum CpMatch {
...
 pub enum CursorSensitivity {
     Unspecified = 0,
     Insensitive,
     Sensitive,
 }
 
 #[derive(Clone, Copy, Debug)]
+#[repr(i32)]
```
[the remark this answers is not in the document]
pmeredit: I think we want to be consistent with the headers over the spec. As far as I am concerned, the headers are the actual spec.

odbc/src/api/definitions.rs
pmeredit: Agreed, the headers are safer to follow.

pmeredit, line 12:
```suggestion
#[repr(u32)]
```

odbc/src/api/definitions.rs
```diff
@@ -66,7 +66,7 @@ pub enum Concurrency {
 }
 
 #[derive(Clone, Copy, Debug, FromPrimitive)]
-#[repr(usize)]
+#[repr(i32)]
```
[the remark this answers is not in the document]
nbagnard: The negative value is a mistake. Is should be one. 
I checked the headers and all values are prefixed with UL.
```
#define SQL_CURSOR_FORWARD_ONLY         0UL
#define SQL_CURSOR_KEYSET_DRIVEN        1UL
#define SQL_CURSOR_DYNAMIC              2UL
#define SQL_CURSOR_STATIC               3UL
```
Can you switch to u32? Once again, the spec and the headers don't match. Let's stick to the headers.

odbc/src/api/definitions.rs
```diff
@@ -58,6 +66,7 @@ pub enum Concurrency {
 }
 
 #[derive(Clone, Copy, Debug, FromPrimitive)]
+#[repr(i32)]
 pub enum CursorType {
     ForwardOnly = 0,
     KeysetDriven = -1,
```
nbagnard, line 72:
```suggestion
    KeysetDriven = 1
```

nbagnard (changes requested):

odbc/src/api/definitions.rs
```diff
@@ -1,20 +1,23 @@
 use num_derive::FromPrimitive;
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, FromPrimitive)]
+#[repr(i32)]
```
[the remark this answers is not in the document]
nbagnard: Created [SQL-1860](https://jira.mongodb.org/browse/SQL-1860) to address this.

nbagnard (approved):
LGTM!

PR #176, opened 2024-01-19
SQL-1817: consolidate definitions in odbc driver

jchemburkar:
This unifies the definitions that were in core, in odbc, and in our fork of odbc-sys into a single crate, "defintions". This keeps the convention of using enums with spec-defined naming conventions as opposed to const structs to define these values.

Most of the changes in this PR are mechanical -- changing imports, and updating things formerly relying on odbc-sys to use the correct naming


terakilobyte (commented):
Overall LGTM, nice work! :fire:

Not approving as I have a general question. We're putting a lot of info together in rust enums. Some are grouped as they are in the c headers under enums, others aren't. Are we comfortable with this?

definitions/src/attributes.rs
```diff
@@ -1,33 +1,32 @@
...
-    // nor does this attribute exist in unixODBC or iODBC implementations. This attribute exists in
-    // Microsoft implementation only and it's usage is unclear.
-    // For private driver manager
-    // SQL_ATTR_APPLICATION_KEY = 203,
-    OutputNts = 10001,
+    SQL_ATTR_ODBC_VERSION = 200,
+    SQL_ATTR_CONNECTION_POOLING = 201,
+    SQL_ATTR_CP_MATCH = 202,
+    SQL_ATTR_OUTPUT_NTS = 10001,
+    SQL_ATTR_DRIVER_UNICODE_TYPE = 1065,
```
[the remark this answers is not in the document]
nbagnard: @terakilobyte 
This is an iODBC specific attribute. See [here](https://github.com/openlink/iODBC/blob/a928c0f37a373dd71c55f5df0ece27b4a05d6de2/include/iodbcext.h#L216)
The DM uses this to check what encoding we are using. Technically this is because iODBC claims that they can do encoding conversion between the app and the driver if needed. In practice, we had no luck so far getting it working correctly I think when we tried to report UTF-16 and that's why we added the feature = "utf32" in cstr.

We should keep it though. The logic is handled correctly in the driver. We might want/be able to use UTF-16 with iODBC one day this way and avoid extra encoding conversion.

I agree that it's missing a comment or two which would explain what the attribute it and what it is used for.
@jchemburkar Would you mind adding a comment?
The section "Unicode Application with a Unicode Driver" on [this page](https://www.progress.com/tutorials/odbc/unicode) describe the idea fairly well, even though this is for the Progress DM. I looked at the iODBC code quite a lot when we tried to get it working, and the feature is the same.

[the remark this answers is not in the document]
jchemburkar: yes, I can add a comment to this effect!

terakilobyte (approved):
LGTM. We spoke offline about follow-up tickets to get rid of some of the features we don't need or that are default.

definitions/src/lib.rs
```diff
@@ -82,6 +87,7 @@ pub const SQLSTATE_SIZE: usize = 5;
 pub const SQLSTATE_SIZEW: usize = 10;
```
nbagnard, line 87: This is unused and not defined anywhere in the spec.
I'm Ok to remove.

definitions/src/lib.rs
```diff
@@ -90,40 +96,42 @@ pub enum FreeStmtOption {
...
-    Complete = 1,
-    Prompt = 2,
-    CompleteRequired = 3,
+    SQL_DRIVER_NO_PROMPT = 0,
+    SQL_DRIVER_COMPLETE = 1,
+    SQL_DRIVER_PROMPT = 2,
+    SQL_DRIVER_COMPLETE_REQUIRED = 3,
 }
 
 // Attribute for string lengths
```
nbagnard, line 137: Could you add the SQL prefix and remove the comment for IS_POINTER to IS_SMALLINT?
The values are in sqlext.h

definitions/src/lib.rs
```diff
@@ -374,3 +340,98 @@ pub struct Numeric {
     pub sign: Char,
     pub val: [Char; MAX_NUMERIC_LEN],
 }
+
+#[allow(non_camel_case_types)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, FromPrimitive)]
+pub enum SqlBool {
+    SQL_FALSE = 0,
+    SQL_TRUE,
```
nbagnard, line 348: [nit] I'm thinking that if we use value very specifically instead of relying on the auto increment, it might help validating the correctness with a script/tool.
But if we managed to get all our values from the C headers translated, this might be throw away work.

I think there aren't that many to change. If it is not too time consuming, can you update the enums?

definitions/src/sql_data_type.rs
```diff
@@ -0,0 +1,49 @@
...
+    SQL_SMALLINT = 5,
+    SQL_FLOAT = 6,
+    SQL_REAL = 7,
+    SQL_DOUBLE = 8,
+    SQL_DATETIME = 9,
+    SQL_VARCHAR = 12,
+    SQL_TYPE_DATE = 91,
+    SQL_TYPE_TIME = 92,
+    SQL_TYPE_TIMESTAMP = 93,
+    SQL_TIME_OR_INTERVAL = 10,
```
nbagnard, line 21: Looking at the headers, the values differs depending on the odbc version.
Since we are an ODBC 3.0+ driver, we should be able to safely say that this is SQL_INTERVAL or if you think that it best to highlight that the value is used under 2 different "aliases" then just a comment on top would do.

nbagnard (changes requested):
Wow, that was not a small PR to review.
It is looking good.
I have comments, but mainly there are related to make sure that we use the same convention everywhere.
Thank you for the refactor :)

definitions/src/sql_data_type.rs
jchemburkar: @terakilobyte and I had talked about this a little bit -- I think keeping odbc 3 conventions makes sense, and so SQL_INTERVAL here makes sense to me. A followup ticket that more generally considers odbc version / feature flags / etc now that we are no longer tied to odbc-sys likely makes sense, too

definitions/src/lib.rs
jchemburkar: by this do you mean explicitly adding `= 1` to `SQL_TRUE`? if so, it won't be a big change :)

definitions/src/attributes.rs
```diff
@@ -112,90 +113,51 @@ impl From<AttrCpMatch> for Pointer {
...
-    LengthExceptionBehaviour = 33,
-    /// SQL_ATTR_METADATA_ID
-    MetadataId = 10014,
+    SQL_ATTR_CURSOR_SCROLLABLE = -1,
+    SQL_ATTR_CURSOR_SENSITIVITY = -2,
+    SQL_ATTR_QUERY_TIMEOUT = 0,
+    SQL_ATTR_MAX_ROWS = 1,
+    SQL_ATTR_NOSCAN = 2,
+    SQL_ATTR_MAX_LENGTH = 3,
+    SQL_ATTR_ASYNC_ENABLE = 4,
```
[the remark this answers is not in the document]
jchemburkar: I think defining the values (and moving ConnectionAttribute) make sense. Though I'm thinking maybe we can just define the constants the two enums use in this file itself, if they are specific to the enums. wdyt?

core/src/col_metadata.rs
```diff
@@ -44,7 +43,7 @@ pub struct MongoColMetadata {
     // non-concise SqlDataType
     pub non_concise_type: SqlDataType,
     // sql_code, always NULL or SQL_CODE_TIMESTAMP (3) for our types
-    // odbc_sys does not define this enum yet, so we just use an i32.
+    //   does not define this enum yet, so we just use an i32.
```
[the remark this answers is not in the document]
jchemburkar: bleh good catch, I think this was just search and replace issue

constants/src/lib.rs
```diff
@@ -143,6 +143,190 @@ pub const SQL_ALL_TABLE_TYPES: &str = "%";
 pub const SQL_ALL_CATALOGS: &str = "%";
 pub const SQL_ALL_SCHEMAS: &str = "%";
 
+pub const SQL_CB_NULL: u16 = 0x0000;
+pub const SQL_U16_ZERO: u16 = 0x0000;
+pub const SQL_CL_START: u16 = 0x0001;
+pub const SQL_U32_ZERO: u32 = 0x0;
+pub const SQL_OIC_CORE: u32 = 0x00000001;
+pub const SQL_SC_SQL92_ENTRY: u32 = 0x00000001;
+pub const SQL_INFO_Y: &str = "Y";
```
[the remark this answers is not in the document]
jchemburkar: Yeah, I think using SQL only for things coming from the header makes sense, I will change this

constants/src/lib.rs
```diff
@@ -143,6 +143,190 @@ pub const SQL_ALL_TABLE_TYPES: &str = "%";
 pub const SQL_ALL_CATALOGS: &str = "%";
 pub const SQL_ALL_SCHEMAS: &str = "%";
 
+pub const SQL_CB_NULL: u16 = 0x0000;
+pub const SQL_U16_ZERO: u16 = 0x0000;
```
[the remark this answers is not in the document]
jchemburkar: same as the last one, I agree we can use a different convention if it isn't from the headers. Maybe something like `MAX_COLUMNS_U16_ZERO` or something may fit (happy to switch it to a better name too)

core/src/col_metadata.rs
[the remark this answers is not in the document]
jchemburkar: I just added the enum because it seemed straightforward

definitions/src/attributes.rs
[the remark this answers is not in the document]
nbagnard: I'm Ok to define the constants here or in another file. Whichever works best. Both approach have pros and cons and none is inherently bad.

definitions/src/lib.rs
nbagnard: Yes, this is what I mean. 
There are other enums with this pattern in that file. And I didn't check closely the other files.

definitions/src/sql_data_type.rs
nbagnard: Works for me 👍

definitions/src/lib.rs
```diff
@@ -390,19 +359,21 @@ pub enum CursorType {
...
 #[allow(non_camel_case_types)]
 #[derive(Clone, Copy, Debug)]
+#[repr(u32)]
 pub enum BindType {
     SQL_BIND_BY_COLUMN = 0,
 }
 
 #[derive(Clone, Copy, Debug)]
-pub enum ParamOperationPtr {}
+pub enum ParamfOperationPtr {}
```
nbagnard, line 376: Looks like an unwanted 'f' appeared here :D
But, shouldn't we just delete these empty enums? I don't see them used anywhere in the code

nbagnard (approved):
LGTM!

PR #178, opened 2024-02-03
SQL-1866: Implement SQLFetchScroll

jchemburkar:


constants/src/lib.rs
```diff
@@ -118,6 +118,10 @@ pub const INVALID_ATTRIBUTE_OR_OPTION_IDENTIFIER: OdbcState<'static> = OdbcState
     odbc_2_state: "S1092",
     odbc_3_state: "HY092",
 };
+pub const FETCH_TYPE_OUT_OF_RANGE: OdbcState<'static> = OdbcState {
```
pmeredit, line 121: I didn't see how we did odbc_2_state until now. I really like this solution

pmeredit (approved):
Looks great!

EthanHardyMongo (approved):
LGTM! Just one question: do we need to add a unit test for `SQLFetchScroll` now that it's implemented?

jchemburkar:
> LGTM! Just one question: do we need to add a unit test for `SQLFetchScroll` now that it's implemented?

been thinking about this a bit, and we should definitely test this. Its a bit tricky since we'd only need to test the case that a fetch orientation we don't accept gets passed. I think maybe this is a good candidate to add when we add more integration testing / flows based on the functions were implementing as part of this epic, but maybe not as part of this ticket

PR #184, opened 2024-02-22
SQL-1950: Merge beta changes back into master

pmeredit:
Third attempt at a PR is the charm


core/src/type_info.rs
```diff
@@ -31,7 +31,8 @@ const LEGACY_DATE: BsonTypeInfo = BsonTypeInfo {
 
 // order of array is by SqlDataType, since that is the ordering of the
 // SQLGetTypeInfo result set according to the spec
-const DATA_TYPES: [BsonTypeInfo; 22] = [
```
pmeredit, line 34: PowerBI requires VarChar be in the types and have type 12 to work

integration_test/src/test_runner/mod.rs
```diff
@@ -133,7 +133,7 @@ pub fn resultset_tests() -> Result<()> {
     run_resultset_tests(false)
 }
 
-#[cfg(target_os = "windows")]
```
pmeredit, line 136: I'm not sure why we didn't want this to run on linux, it works.

odbc/src/api/functions.rs
```diff
@@ -2006,6 +2018,8 @@ macro_rules! sql_get_info_helper {
                         | SQL_FN_NUM_DEGREES
```
pmeredit, line 2018: We supported more functions that are used by Direct Query.

constants/src/lib.rs
```diff
@@ -239,6 +240,26 @@ pub const SQL_FN_NUM_POWER: u32 = 0x00100000;
 pub const SQL_FN_NUM_RADIANS: u32 = 0x00200000;
 pub const SQL_FN_NUM_ROUND: u32 = 0x00400000;
```
pmeredit, line 242: The way I create these is actually pasting the defines from the header file and modifying them with regular expressions:
/'<,'>s/#define //
/'<,'>s/ /: u32 =/
/'<,'>s/^/pub const /
/'<,'>s/$/;/

core/src/type_info.rs
pmeredit: (see comment in the code above)

evergreen.yml
```diff
@@ -177,7 +177,7 @@ functions:
             cargo set-version $RELEASE_VERSION
           fi
           cargo install cargo-get
```
pmeredit, line 179: cargo-get had backward breaking Ux changes :(

evergreen.yml
```diff
@@ -369,7 +369,7 @@ functions:
               MINOR_VERSION="0.1"
               VERSION_LABEL="0.1.0"
           else
-              MINOR_VERSION=$(echo "$RELEASE_VERSION" | sed 's|\([0-9]\+[.][0-9]\+\)[.][0-9]\+|\1|')
+              MINOR_VERSION=$(echo "$RELEASE_VERSION" | sed 's|\([0-9]\+[.][0-9]\+\)[.][0-9]\+.*|\1|')
```
pmeredit, line 372: support `-beta`, etc.

evergreen.yml
```diff
@@ -1223,14 +1223,15 @@ functions:
 
           ~/.cargo/bin/rustup default nightly
           ~/.cargo/bin/rustup target add x86_64-unknown-linux-gnu
```
pmeredit, line 1225: This looks like it was removed. Why?

integration_test/src/test_runner/mod.rs
pmeredit: (I added it specifically so I could fix the tests locally)

core/src/type_info.rs
```diff
@@ -31,7 +31,8 @@ const LEGACY_DATE: BsonTypeInfo = BsonTypeInfo {
...
+const DATA_TYPES: [BsonTypeInfo; 23] = [
+    BsonTypeInfo::VARCHAR,             // SqlDataType(12)
```
jchemburkar, line 35: this should be in order of sql data type, so I think this needs to come after `LEGACY_DATE`

integration_test/tests/odbc_2_tests.rs
```diff
@@ -87,7 +87,8 @@ mod integration {
         let _ = unsafe { Box::from_raw(env_handle) };
     }
 
-    const EXPECTED_DATATYPES: [SqlDataType; 22] = [
+    const EXPECTED_DATATYPES: [SqlDataType; 23] = [
+        SqlDataType::SQL_VARCHAR,
```
jchemburkar, line 91: this will need to move in accordance with the prev comment

resources/integration_test/tests/functions.yml
```diff
@@ -45,14 +45,15 @@ tests:
...
-    expected_precision: [ 0, 10, 10, 0, 0, 0, 10, 10, 10, 10, 10, 10, 0, 10, 10, 10, 10, 10, 10 ]
+    expected_precision: [ 65535, 10, 10, 65535, 65535, 65535, 10, 10, 10, 10, 10, 10, 65535, 10, 10, 10, 10, 10, 10 ]
     expected_scale: [ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]
     expected_nullability: [ 0, 0, 1, 1, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1, 1, 0, 1, 1, 1 ]
   - description: SQLGetTypeInfo function get all types
     test_definition: [ "sqlgettypeinfo", 0 ]
     db: integration_test
     expected_result:
-    - [ "string", -9, null, "'", "'", null, 1, 1, 3, null, 0, null, "string", null, null, -9, null, null, null ]
+    - [ "varchar", 12, 65535, "'", "'", null, 1, 1, 3, null, 0, null, "varchar", null, null, 12, null, null, null ]
```
jchemburkar, line 55: same as prev

jchemburkar (commented):
looking good, just had a couple nits!

evergreen.yml
```diff
@@ -1346,6 +1347,9 @@ functions:
         working_dir: mongosql-odbc-driver/resources/download-center/
         script: |
           ${prepare_shell}
+          if [[ $(git branch --show-current) != 'master' ]]; then
```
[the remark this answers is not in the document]
pmeredit: I don't suspect we will be changing the name

nbagnard (changes requested):

evergreen.yml
pmeredit: sed is far easier for me to read because I don't know what cut is or does :joy:

evergreen.yml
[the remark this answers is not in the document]
pmeredit: sgtm

odbc/src/api/col_attr_describe_tests.rs
```diff
@@ -484,8 +484,8 @@ mod unit {
             (Desc::SQL_DESC_FIXED_PREC_SCALE, 0),
             (Desc::SQL_DESC_LENGTH, 0),
             (Desc::SQL_DESC_NULLABLE, 0),
-            (Desc::SQL_DESC_OCTET_LENGTH, 0),
-            (Desc::SQL_DESC_PRECISION, 0),
+            (Desc::SQL_DESC_OCTET_LENGTH, 65535),
```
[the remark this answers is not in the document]
pmeredit: https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/transfer-octet-length?view=sql-server-ver16

OCTENT LENGTH is max size. We choose 2^16 to be sort of reasonable. If this is 0, direct query does not work for several conversions to VARCHAR/STRING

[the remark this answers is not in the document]
pmeredit: https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/column-size-decimal-digits-transfer-octet-length-and-display-size?view=sql-server-ver16

```Similarly, the values for transfer octet length do not come from SQL_DESC_LENGTH. They come from the SQL_DESC_OCTET_LENGTH of a field of a descriptor for all character and binary types. There is no descriptor field that holds this information for other types.```

File: odbc/src/api/functions.rs [hunk 13]
```diff
@@ -2179,7 +2207,8 @@ macro_rules! sql_get_info_helper {
                     // TIMEDATE_DIFF, so this value will not be used. For the
                     // MongoSQL DATEADD and DATEDIFF functions, we support the
                     // following intervals.
-                    const TIMEDATE_INTERVALS: u32 = SQL_FN_TSI_SECOND
+                    const TIMEDATE_INTERVALS: u32 = SQL_FN_TSI_FRAC_SECOND
```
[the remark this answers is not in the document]
pmeredit: We do, and Direct Query is using this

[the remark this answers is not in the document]
pmeredit: https://github.com/10gen/mongosql-rs/blob/master/mongosql/src/air/definitions.rs#L634

resources/integration_test/odbc2/queries.yaml
```diff
@@ -17,7 +17,7 @@ tests:
     expected_is_searchable: [2, 2, 3, 2, 2]
     expected_is_unsigned: [0, 1, 1, 1, 0]
     expected_sql_type: [4, -7, -9, 11, 4]
-    expected_precision: [10, 1, 0, 23, 10]
+    expected_precision: [10, 1, 65535, 23, 10]
```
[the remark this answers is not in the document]
pmeredit: PRECISION is maximum size, not current size

odbc/src/api/col_attr_describe_tests.rs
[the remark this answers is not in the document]
pmeredit: Explained in other thread where you asked the same question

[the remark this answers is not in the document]
pmeredit: The ODBC documentation is terrible.

https://www.ibm.com/docs/en/db2-for-zos/11?topic=attributes-precision-sql-data-types

evergreen.yml
pmeredit: ok, this cut command works for me, I like it!

resources/integration_test/odbc2/queries.yaml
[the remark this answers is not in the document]
pmeredit: Correct, but in mongo we have no way to know what maximum for the result set is other than maximum size for all strings

odbc/src/api/col_attr_describe_tests.rs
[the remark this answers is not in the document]
pmeredit: Right, if you look at what ibm did in db2 they said precision for char types to be length. I'm comfortable doing what they do since the ODBC spec seems to be silent. (this link https://www.ibm.com/docs/en/db2-for-zos/11?topic=attributes-precision-sql-data-types)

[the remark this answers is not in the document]
pmeredit: I think the issue here is that simple types needs to set precision to MAX_STRING_SIZE, also, when it's not a fixed size (like ObjectID)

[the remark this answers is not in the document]
pmeredit: I made that change and am also leaving OCTENT_LENGTH as 0. I tested to make sure that didn't break Direct Query. Direct Query is relying only on PRECISION for strings, thankfully.

nbagnard (approved):
LGTM!

PR #185, opened 2024-02-23
SQL-1864: Implement SQLFreeStmt

mattChiaravalloti:
This PR implements `SQLFreeStmt` according to the guidelines in the design doc.

### DIFF
--- a/core/src/mock_query.rs
+++ b/core/src/mock_query.rs
@@ -55,4 +55,11 @@ impl MongoStatement for MongoQuery {
     fn get_resultset_metadata(&self) -> &Vec<MongoColMetadata> {
         &self.resultset_metadata
     }
+
+    // Close the cursor by setting the current value and cursor to None.
+    fn close_cursor(&mut self) -> Result<()> {
+        self.current = None;
+        self.resultset = vec![];
+        Ok(())
+    }
 }
--- a/core/src/mock_query.rs
+++ b/core/src/mock_query.rs
@@ -57,9 +57,8 @@ impl MongoStatement for MongoQuery {
     }
 
     // Close the cursor by setting the current value and cursor to None.
-    fn close_cursor(&mut self) -> Result<()> {
+    fn close_cursor(&mut self) {
         self.current = None;
         self.resultset = vec![];
-        Ok(())
     }
 }
--- a/core/src/stmt.rs
+++ b/core/src/stmt.rs
@@ -30,9 +30,7 @@ pub trait MongoStatement: Debug {
     }
     // Closes the cursor.
     // Only MongoQuery supports this workflow. The other statements don't.
-    fn close_cursor(&mut self) -> Result<()> {
-        Err(Error::UnsupportedOperation("close_cursor"))
-    }
+    fn close_cursor(&mut self) {}
 }
 
 #[derive(Debug)]
--- a/core/src/stmt.rs
+++ b/core/src/stmt.rs
@@ -28,6 +28,11 @@ pub trait MongoStatement: Debug {
     fn execute(&mut self, _: &MongoConnection) -> Result<bool> {
         Err(Error::UnsupportedOperation("execute"))
     }
+    // Closes the cursor.
+    // Only MongoQuery supports this workflow. The other statements don't.
+    fn close_cursor(&mut self) -> Result<()> {
+        Err(Error::UnsupportedOperation("close_cursor"))
+    }
 }
 
 #[derive(Debug)]
--- a/odbc/src/api/free_stmt_tests.rs
+++ b/odbc/src/api/free_stmt_tests.rs
@@ -83,7 +83,7 @@ mod unit {
             // Here, we create a MockQuery with nonsense dummy data since the
             // values themselves do not matter.
             let mock_query = &mut MongoQuery::new(
-                vec![doc! {"x": "y"}],
+                vec![doc! {"x": "y"}, doc! {"x": "z"}],
                 vec![MongoColMetadata::new(
                     "test_db",
                     "dn".to_string(),
@@ -106,8 +106,7 @@ mod unit {
                 SQLFreeStmt(stmt as *mut _, FreeStmtOption::SQL_CLOSE as i16)
             );
 
-            // Assert that the mongo_statement closed the cursor (no resultset, no current)
-            // but retained the metadata.
+            // Assert that the mongo_statement closed the cursor (no resultset, no current).
             match s
                 .mongo_statement
                 .read()
@@ -129,6 +128,7 @@ mod unit {
                 .unwrap()
                 .next(None)
             {
+                // we expect false since there should be no data to iterate after SQLFreeStmt
                 Ok((false, _)) => {}
                 _ => panic!("cursor not closed -- able to call next()"),
             }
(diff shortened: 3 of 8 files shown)


core/src/mock_query.rs
```diff
@@ -55,4 +55,11 @@ impl MongoStatement for MongoQuery {
     fn get_resultset_metadata(&self) -> &Vec<MongoColMetadata> {
         &self.resultset_metadata
     }
+
+    // Close the cursor by setting the current value and cursor to None.
```
mattChiaravalloti, line 59: Decided to implement this for `MockQuery` as well so I could use this type in the unit tests -- turned out to be easier that way.

core/src/stmt.rs
```diff
@@ -28,6 +28,11 @@ pub trait MongoStatement: Debug {
     fn execute(&mut self, _: &MongoConnection) -> Result<bool> {
         Err(Error::UnsupportedOperation("execute"))
     }
+    // Closes the cursor.
+    // Only MongoQuery supports this workflow. The other statements don't.
+    fn close_cursor(&mut self) -> Result<()> {
```
mattChiaravalloti, line 33: Adding this to the trait seemed the most straightforward way to close the cursor when in the context of `SQLFreeStmt` since all we have access to at that point is the `mongo_statement` value which is typed as the trait `MongoStatement`.

odbc/src/api/free_stmt_tests.rs
```diff
@@ -0,0 +1,137 @@
...
+            )
+        }
+    }
+
+    #[test]
+    fn test_free_stmt_unbind() {
+        let env = &mut MongoHandle::Env(Env::with_state(EnvState::Allocated));
+        let conn =
+            &mut MongoHandle::Connection(Connection::with_state(env, ConnectionState::Allocated));
+        let stmt: *mut _ =
```
mattChiaravalloti, line 49: Since I manipulate the `stmt` value in the `unsafe` block below, I am unable to use the helper function here (or in the next test). This is ok since these are just unit tests and this is only 4 lines of "repeated" code.

odbc/src/api/functions.rs
```diff
@@ -1348,8 +1348,46 @@ fn sql_free_handle(handle_type: HandleType, handle: *mut MongoHandle) -> Result<
...
+                }
+                // We do not implement SQLBindParameter, so this is a no-op.
+                Some(FreeStmtOption::SQL_RESET_PARAMS) => SqlReturn::SUCCESS,
+                _ => {
+                    // TODO: This is handled by the DM, so really this is an unreachable branch.
+                    //       Should we include the error in diagnostics or is it not worth?
+                    // add_diag_info!(
+                    //     mongo_handle,
+                    //     ODBCError::InvalidAttrIdentifier(option as i32)
+                    // );
```
mattChiaravalloti, line 1384: Thoughts on this? Jonathan suggested we _may_ be ok to not include this error here since this is not a case we should ever hit since the DM should ensure the function call is rerouted if an invalid option comes in. I'm happy to remove it but figured I'd put it into review for confirmation.

odbc/src/handles/definitions.rs
mattChiaravalloti, line 1: All the changes in this file will overlap with Ethan's changes for implementing SQLBindCol. I took this all straight from the mock branch.

odbc/src/api/free_stmt_tests.rs
```diff
@@ -0,0 +1,137 @@
...
+            // Set the mongo_statement
+            let s = (*stmt).as_statement().unwrap();
+            *s.mongo_statement.write().unwrap() = Some(Box::new(mock_query.clone()));
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt as *mut _, FreeStmtOption::SQL_CLOSE as i16)
+            );
+
+            // Assert that the mongo_statement closed the cursor (no resultset, no current).
```
mattChiaravalloti, line 109: I didn't want to make the `MockQuery` fields public, so instead I am indirectly testing the "closed cursor" state here by first trying to `get_data` (which, if not for `SQLFreeStmt`, would be valid since we already called `next` earlier) and then trying to call `next` (which, again, if not for `SQLFreeStmt`, would be valid).

odbc/src/api/functions.rs
jchemburkar: looking at the code now I think I agreew with what I said earlier, I dont know if we should do additional work to ultimately return an error. my mental rust compiler is a little weak but could we put unreachable here with a coment? or does it have to be an error to satisfy the return type.

We should, however, probably handle `None` separately then the default error here. I think we need to return `HY092` as the state with the message `Option type out of range` if we get a value that isn't one of these 4. We'll likely need to create a new error for this

odbc/src/api/free_stmt_tests.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn create_stmt_handle() -> *mut MongoHandle {
+        let env = &mut MongoHandle::Env(Env::with_state(EnvState::Allocated));
+        let conn =
+            &mut MongoHandle::Connection(Connection::with_state(env, ConnectionState::Allocated));
+        let stmt: *mut _ =
+            &mut MongoHandle::Statement(Statement::with_state(conn, StatementState::Allocated));
+
+        stmt
+    }
+
```
jchemburkar, line 26: related to the question of the default mongostatement.close_cursor, maybe we should add a test where we have a statement that isn't a MongoQuery and call close?

jchemburkar (changes requested):
the trait ended up being really nice! I just had a couple thoughts about error handling/testing but the function looks good

core/src/stmt.rs
```diff
@@ -28,6 +28,11 @@ pub trait MongoStatement: Debug {
...
+        Err(Error::UnsupportedOperation("close_cursor"))
```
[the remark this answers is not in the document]
mattChiaravalloti: That sounds good. I started implementing it with Ok(()) but decided to go with an Error to mirror `execute`.

I'm going to make this change, but I'm going to go a bit further and not return anything since there is no way for this to return an error _other than_ the current default implementation. That will mean the default implementation is truly a no-op.

odbc/src/api/free_stmt_tests.rs
mattChiaravalloti: done!

odbc/src/api/functions.rs
mattChiaravalloti: Oh wait I totally didn't notice the "(DM)" next to the "Option type out of range" error on the docs page before. I thought that the only driver manager errors were the "Function sequence" and "Driver does not support this function" errors. Given that, do we need any actual true error handling here? Could/should we just do a wildcard for everything other than the 3 real options, and have that wildcard be either `unreachable!()` or `SqlReturn::ERROR`?

jchemburkar: ah good catch. In that case I think we might be good then to just wildcard it (sorry for the misdirect). I feel like `SqlReturn::ERROR` might be more consistent with what we do in other places in the driver but I could see it going either way!

jchemburkar (approved):
Lgtm!

pmeredit (approved):
LGTM!

PR #186, opened 2024-02-23
SQL-1865: Implement SQLCancel

jchemburkar:
Implementation for SQLCancel; we associate a uuid with each statement handle, and pass that to any executing queries as a comment. Additionally, we have an intermediate statement state that indicates the statement handle is running a query.

When SQLCancel is called, it first checks if the statement handle is running a query (otherwise, we do nothing and return sqlreturn success). If there is a query running, we use current op to get all running operations, and filter the comments of these operations to see if any are associated with the statement handle, cancelling if they are.


odbc/src/api/cancel_tests.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        // StatementState::Allocated is the only valid state other than the state indicating a query is executing.
+        // SQLCancel should be a no op in this case; verify we get a success.
+        let stmt: *mut _ =
+            &mut MongoHandle::Statement(Statement::with_state(conn, StatementState::Allocated));
+        unsafe { assert_eq!(SQLCancel(stmt as *mut _), SqlReturn::SUCCESS) }
+    }
+
+    #[test]
+    fn test_cancel_running_query_not_executing() {
```
nbagnard, line 123: Maybe add a comment here to say that this test is checking that if at the time Cancel is called the query is still executing, but when killOp is called, the execution is over, then we still gracefully return SUCCESS.

odbc/src/api/functions.rs
```diff
@@ -414,7 +414,38 @@ pub unsafe extern "C" fn SQLBulkOperations(
 #[named]
 #[no_mangle]
 pub unsafe extern "C" fn SQLCancel(statement_handle: HStmt) -> SqlReturn {
-    unimpl!(statement_handle);
+    panic_safe_exec_keep_diagnostics!(
```
nbagnard, line 417: I thought we were going to drop the mongo_statement for some reason.
I checked the code, I am not sure how we re-use statements (or if we actually can, I'd have to dig a little more), but if we are in the middle of retrieving data and cancel is called, we should close the cursor or drop the mongo_statement somehow if it's possible.

nbagnard (changes requested):
It's looking good already :)

odbc/src/api/functions.rs
jchemburkar: I agree this is a bit unclear, but the case that motivates my uncertainty would be: if you issue a query using SQLPrepare followed by SQLExecute, then call SQLCancel, should you be able to call SQLExecute again? If the answer is yes, I believe we need to keep the `mongo_statement` in tact. 

I was thinking the answer to this might be yes, since other behaviors we don't support for SQLCancel behave this way -- namely [cancelling functions that need data](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlcancel-function?view=sql-server-ver16#canceling-functions-that-need-data). So My thought was to just allow the program to behave as if execution failed and let the thread calling SQLExecute figure out how to proceed

core/src/conn.rs
```diff
@@ -64,6 +64,36 @@ impl MongoConnection {
             bson::from_document(cmd_res).map_err(Error::DatabaseVersionDeserialization)?;
         Ok(build_info.data_lake.version)
     }
+
+    /// cancels all queries for a given statement id
+    pub fn cancel_queries_for_statement(&self, statement_id: Bson) -> Result<bool> {
+        // use $currentOp to list all queries currently running
+        let admin_db = self.client.database("admin");
+        let current_ops_pipeline = vec![doc! {"$currentOp": {}}];
```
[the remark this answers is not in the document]
jchemburkar: I wasn't able to get this to work through the driver, but it does seem to work ok in my smaller testapp. I'll look into this, it should work and make the code much simpler 😄

odbc/src/api/cancel_tests.rs
```diff
@@ -0,0 +1,140 @@
+use crate::{handles::definitions::*, SQLCancel};
+use definitions::SqlReturn;
+use mongo_odbc_core::MongoConnection;
+use mongodb::sync::Client;
+use std::env;
+
+mod integration {
```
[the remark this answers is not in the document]
jchemburkar: Yes, I agree a full integration test would be useful, though I'm a bit unsure of if/how it would actually be possible if we can't use the same handle amongst multiple threads. I had tried a couple ways but couldn't quite get it to work. Do you have a certain test setup in mind?

odbc/src/api/functions.rs
nbagnard: Yes, that seems like the right thing to do.
If found this in the spec, which corroborate your thinking. "To close a cursor, an application should call SQLCloseCursor, not SQLCancel." 👍

odbc/src/api/cancel_tests.rs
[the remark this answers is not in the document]
nbagnard: uhm, no. I quickly looked into it and it looks like, we would need some kind of thread-safe wrapper around the handle. I am not a Rust multi-thread expert at all unfortunately.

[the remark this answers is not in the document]
jchemburkar: Likewise, though I'm not sure it will be trivial. Maybe we can file a ticket and it could be a good skunkworks project? I do think its nice to have and worth implementing, but none of the tools we're currently focussed on rely on this functionality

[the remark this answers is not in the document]
nbagnard: Yes, this can be a follow up item.

core/src/conn.rs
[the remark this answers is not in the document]
jchemburkar: It was possible, but requires a small update to the adf config. I added a config of our own and pointed the run adf script at that for now, but open to other ways to structure this since I know we usually want our adf scripts to stay in sync.

core/src/conn.rs
```diff
@@ -64,6 +64,33 @@ impl MongoConnection {
...
+        let admin_db = self.client.database("admin");
+        let current_ops_pipeline = vec![
+            doc! {"$currentOp": {}},
+            doc! {"$match": {"command.comment": statement_id}},
+        ];
+        let mut cursor = admin_db
+            .aggregate(current_ops_pipeline, None)
+            .map_err(Error::QueryExecutionFailed)?;
+
+        // iterate through all running operations, looking for commands
```
nbagnard, line 80:
```suggestion
        // iterate through the results and kill the operations
```

odbc/src/api/cancel_tests.rs
```diff
@@ -0,0 +1,142 @@
...
+    use super::*;
+
+    fn generate_connection_uri() -> String {
+        let user_name = env::var("ADF_TEST_LOCAL_USER").expect("ADF_TEST_LOCAL_USER is not set");
+        let password = env::var("ADF_TEST_LOCAL_PWD").expect("ADF_TEST_LOCAL_PWD is not set");
+        let host = env::var("ADF_TEST_LOCAL_HOST").expect("ADF_TEST_LOCAL_HOST is not set");
+        format!("mongodb://{user_name}:{password}@{host}")
+    }
+
+    // skip-reason: https://jira.mongodb.org/browse/MHOUSE-9953
```
nbagnard, line 18: Can you had a ticket linked to this MHOUSE one in our Epic to track this.

nbagnard (approved):
LGTM!

Just 2 very small suggestion/follow-up. No need to re-tag me.

core/src/query.rs
```diff
@@ -123,24 +127,26 @@ impl MongoStatement for MongoQuery {
...
-            _ => db
-                .aggregate(pipeline, None)
-                .map_err(Error::QueryExecutionFailed)?,
+        let opt = AggregateOptions::builder().comment_bson(Some(stmt_id));
+        // If the query timeout is 0, it means "no timeout"
+        let options = if self.query_timeout.is_some_and(|timeout| timeout > 0) {
+            opt.max_time(Duration::from_millis(self.query_timeout.unwrap() as u64))
+                .build()
+        } else {
+            opt.build()
```
EthanHardyMongo, line 136: Nice refactor here! I've never seen `is_some_and` used before lol.

odbc/src/api/functions.rs
```diff
@@ -1091,22 +1128,26 @@ pub unsafe extern "C" fn SQLExecute(statement_handle: HStmt) -> SqlReturn {
             let mongo_handle = MongoHandleRef::from(statement_handle);
             let stmt = must_be_valid!(mongo_handle.as_statement());
             let connection = must_be_valid!((*stmt.connection).as_connection());
+            // set the statment state to executing so SQLCancel knows to search the op log for hanging queries
+            *stmt.state.write().unwrap() = StatementState::SynchronousQueryExecuting;
             odbc_unwrap!(sql_execute(stmt, connection), mongo_handle);
+            *stmt.state.write().unwrap() = StatementState::Allocated;
```
EthanHardyMongo, line 1134:
```suggestion
            *stmt.state.write().unwrap() = StatementState::SynchronousQueryExecuting;
            odbc_unwrap!(sql_execute(stmt, connection), mongo_handle);
            // return the statement state to its original value
            *stmt.state.write().unwrap() = StatementState::Allocated;
```
[optional] This would keep consistency with the comments above.

core/src/conn.rs
```diff
@@ -64,6 +64,33 @@ impl MongoConnection {
...
+    pub fn cancel_queries_for_statement(&self, statement_id: Bson) -> Result<bool> {
+        // use $currentOp and match the comment field to identify any queries issued by the current statement
+        let admin_db = self.client.database("admin");
+        let current_ops_pipeline = vec![
+            doc! {"$currentOp": {}},
+            doc! {"$match": {"command.comment": statement_id}},
+        ];
+        let mut cursor = admin_db
+            .aggregate(current_ops_pipeline, None)
+            .map_err(Error::QueryExecutionFailed)?;
```
EthanHardyMongo, line 78:
```suggestion
        // use $currentOp and match the comment field to identify any queries issued by the current statement
        let current_ops_pipeline = vec![
            doc! {"$currentOp": {}},
            doc! {"$match": {"command.comment": statement_id}},
        ];
        let admin_db = self.client.database("admin");
        let mut cursor = admin_db
            .aggregate(current_ops_pipeline, None)
            .map_err(Error::QueryExecutionFailed)?;
```
[optional] This is a way to make the comment closer to what it's describing and move `admin_db` to closer to its use.

odbc/src/api/cancel_tests.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    fn generate_connection_uri() -> String {
+        let user_name = env::var("ADF_TEST_LOCAL_USER").expect("ADF_TEST_LOCAL_USER is not set");
+        let password = env::var("ADF_TEST_LOCAL_PWD").expect("ADF_TEST_LOCAL_PWD is not set");
+        let host = env::var("ADF_TEST_LOCAL_HOST").expect("ADF_TEST_LOCAL_HOST is not set");
+        format!("mongodb://{user_name}:{password}@{host}")
+    }
+
+    // skip-reason: SQL-1959
+    // the local adf currently can't be configured to handle killop commands; it is on be default in production
```
EthanHardyMongo, line 19:
```suggestion
    // the local adf currently can't be configured to handle killop commands; it is on by default in production
```
nit

EthanHardyMongo (approved):
LGTM! I just had 2 optional suggestions regarding comments and one typo nit.

odbc/src/api/cancel_tests.rs
jchemburkar: thank you for catching this, i fear I am quite prone to typos/misspelling in my comments 😔

core/src/query.rs
jchemburkar: ++ it came from @nbagnard 's suggestion, I think it came out very clean 😄

PR #187, opened 2024-03-05
SQL-1863: Implement Column Wise binding of a single row 

EthanHardyMongo:
This PR implements `SQLBindCol` to allow column-wise binding of one row at a time. Any other type of binding (e.g., row-wise) will cause an error.


mattChiaravalloti:
Reading through the README, I just realized we still have all the newly supported functions listed as "unsupported". Would you mind updating the README as part of this PR to remove those? (`SQLFreeStmt`, `SQLCancel`, `SQLFetchScroll`, and of course `SQLBindCol`. I think that's everything.)

odbc/src/api/bind_col_tests.rs
```diff
@@ -0,0 +1,319 @@
+mod unit {
```
EthanHardyMongo, line 1: A lot of my tests use repeated code; however, when I tried to refactor and condense, I was getting some lifetime related errors, so I decided not to take the time to condense everything yet. However, I can spend more time on this and condense it if that's preferable.

odbc/src/api/bind_col_tests.rs
```diff
@@ -0,0 +1,319 @@
...
+            let indicator: *mut Len = null_mut();
+
+            // Assert that SQLBindCol is successful
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLBindCol(
+                    stmt as *mut _,
+                    1,
+                    CDataType::SQL_C_SLONG as SmallInt,
+                    null_mut(),
```
mattChiaravalloti, line 172:
```suggestion
                    null_mut(), // Note that when TargetValuePtr is null, the driver unbinds the data buffer for the column
```
Let's include a comment explaining that _this_ is the most consequential part of the test.

odbc/src/api/bind_col_tests.rs
```diff
@@ -0,0 +1,319 @@
...
+            &mut MongoHandle::Connection(Connection::with_state(env, ConnectionState::Allocated));
+        let stmt: *mut _ =
+            &mut MongoHandle::Statement(Statement::with_state(conn, StatementState::Allocated));
+
+        unsafe {
+            // Get Statement
+            let s = (*stmt).as_statement().unwrap();
+
+            // Set the bound_cols to non-None initially.
+            *s.bound_cols.write().unwrap() = Some(map! {
```
mattChiaravalloti, line 118: For this test, we might benefit from having `bound_cols` include more than one mapping. That way, the test can show with more certainty that only the referenced column is unbound, as opposed to the whole map being emptied. Would you update this map to include a second binding, and then update the final assertion to show that that second binding still remains even after unbinding `1`?

odbc/src/api/bind_col_tests.rs
```diff
@@ -0,0 +1,319 @@
...
+
+            // set all statement attributes to the correct values.
+            s.attributes.write().unwrap().row_bind_offset_ptr = null_mut();
+            s.attributes.write().unwrap().row_array_size = 1;
+            s.attributes.write().unwrap().row_bind_type = BindType::SQL_BIND_BY_COLUMN as usize;
+
+            // Give bound_cols a column binding with an invalid target_type. The target_type check is in SQLFetch, so if
+            // there is an attempt to create a column binding with an invalid target_type, it will be added to bind_cols,
+            // and SQLBindCol will return SQLReturn::Success. Then, everything will operate as if nothing is wrong until
+            // SQLFetch accesses bound_cols and checks for an invalid target_type. At this point, the error will occur.
```
mattChiaravalloti, line 274: Why is this the case. I see the following in the `SQLBindCol` docs: 

SQLSTATE | Error | Description
-- | -- | --
HY003 | Invalid application buffer type | The argument TargetType was neither a valid data type nor SQL_C_DEFAULT.

Doesn't this mean we should have returned an error from `SQLBindCol` if an invalid target type was provided?

odbc/src/api/functions.rs
```diff
@@ -337,13 +337,82 @@ fn sql_alloc_handle(
...
+                || !stmt
+                    .attributes
+                    .read()
+                    .unwrap()
+                    .row_bind_offset_ptr
+                    .is_null()
+                || stmt.attributes.read().unwrap().row_bind_type
+                    != (BindType::SQL_BIND_BY_COLUMN as usize)
+            {
+                unsupported_function!(hstmt)
```
mattChiaravalloti, line 363: Is it possible/worth providing the context that leads to this panic? As in, which of the three conditions actually triggered this?

mattChiaravalloti (changes requested):
Nice job so far!

odbc/src/api/functions.rs
```diff
@@ -1207,7 +1276,41 @@ unsafe fn sql_fetch_helper(statement_handle: HStmt, function_name: &str) -> SqlR
 
         *stmt.var_data_cache.write().unwrap() = Some(HashMap::new());
 
-        if !warnings_opt.is_empty() {
+        let mut success_with_info_encountered = false;
+
+        if stmt.bound_cols.read().unwrap().is_some() {
+            let mongo_handle_for_sql_get_data_helper = MongoHandleRef::from(statement_handle);
+
+            for (col, bound_col_info) in stmt.bound_cols.read().unwrap().as_ref().unwrap().iter() {
```
mattChiaravalloti, line 1284:
```suggestion
        if let Some(ref bound_cols) = stmt.bound_cols.read().unwrap().as_ref() {
            let mongo_handle_for_sql_get_data_helper = MongoHandleRef::from(statement_handle);

            for (col, bound_col_info) in bound_cols.iter() {
```
I didn't try this against the compiler, but is something like this possible? To reduce the duplicate `stmt.bound_cols.read().unwrap()` calls?

odbc/src/api/bind_col_tests.rs
EthanHardyMongo: Yes you're right. I was thinking It didn't matter where the error was returned as long as it got returned and returning the error in SQLFetch worked smoother with the code; however, I'll go ahead and move it.

odbc/src/api/functions.rs
EthanHardyMongo: Yes, I think that would be helpful! Originally, Natacha told me to just return `unsupported_function!(hstmt)` if one of these conditions was encountered; however, it would be helpful if the user knew what the actually issue is. Also, I just noticed that `unsupported_function!(hstmt)` will run this:
```
let name = function_name!();
add_diag_info!(mongo_handle, ODBCError::Unimplemented(name));
```
So the user will get a message essentially saying `SQLBindCol` is unimplemented. 

I could do something like this instead:
```
add_diag_info!(mongo_handle, ODBCError::Unimplemented("<describe the incorrect setting>"));
return SqlReturn::ERROR;
```
This way the user knows what went wrong.

mattChiaravalloti: Nice!

odbc/src/api/functions.rs
```diff
@@ -337,13 +337,82 @@ fn sql_alloc_handle(
...
+            // Unbind column if target_value is null
+            if target_value.is_null() {
+                stmt.bound_cols
+                    .write()
+                    .unwrap()
+                    .as_mut()
+                    .unwrap()
+                    .remove(&col_number);
+            }
+            // Bind column
```
jchemburkar, line 395: I believe this works for the rebinding case as well, since insert if a key exists should overwrite the older buffer info, maybe worth calling out in this comment

odbc/src/api/functions.rs
```diff
@@ -1207,7 +1276,41 @@ unsafe fn sql_fetch_helper(statement_handle: HStmt, function_name: &str) -> SqlR
 
         *stmt.var_data_cache.write().unwrap() = Some(HashMap::new());
 
-        if !warnings_opt.is_empty() {
+        let mut success_with_info_encountered = false;
+
+        if let Some(ref bound_cols) = stmt.bound_cols.read().unwrap().as_ref() {
```
jchemburkar, line 1281: also a small nit but maybe a comment at the top of this block would be helpful -- jsut noting we copy the data from the result set into the bound buffers if we have bound columns

odbc/src/api/bind_col_tests.rs
```diff
@@ -0,0 +1,333 @@
+mod unit {
+    use crate::{
```
jchemburkar, line 2: related to earlier comment, maybe we should test rebinding a column works. E.g. call bind by column twice with the same number and check that the buffer ptr updates properly?

jchemburkar (changes requested):
looks really good already, nice work. Just had a couple nits in addition to matt's comments, and one bigger question

odbc/src/api/functions.rs
```diff
@@ -337,13 +337,82 @@ fn sql_alloc_handle(
...
 ) -> SqlReturn {
-    unimpl!(hstmt);
+    panic_safe_exec_keep_diagnostics!(
+        debug,
+        || {
+            let mongo_handle = MongoHandleRef::from(hstmt);
+            let stmt = must_be_valid!((*mongo_handle).as_statement());
+
+            // Currently, we only support column binding of one row at a time.
+            if stmt.attributes.read().unwrap().row_array_size != 1
```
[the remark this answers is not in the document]
nbagnard: uhm, yes, from the SSIS traces, the tool is binding a rowset, not a single row.
Also, if we want to start supporting only one row at a time, we need to add the protection on the SQLSetStmtAttribute, not just on SQLBind. It's good to have it, but it is better to fail earlier so that it is clear we don't support rowset binding (yet)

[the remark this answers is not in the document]
EthanHardyMongo: I'll add the checks to `SQLSetStmtAttrW` in this pr.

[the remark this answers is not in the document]
EthanHardyMongo: I was adding checks to `SQLSetStmtAttrW`, but I'm not sure what to do about `SQL_ATTR_ROW_ARRAY_SIZE` because it modifies the same field as `SQL_ROWSET_SIZE`. Do we need to add another field to `StatementAttributes` and break these 2 up? I think we should add another field called `rowset_size` or something like that.

```
        StatementAttribute::SQL_ATTR_ROW_ARRAY_SIZE | StatementAttribute::SQL_ROWSET_SIZE => {
            match FromPrimitive::from_i32(value_ptr as i32) {
                Some(ras) => {
                    stmt.attributes.write().unwrap().row_array_size = ras;
                    SqlReturn::SUCCESS
                }
                None => {
                    stmt_handle
                        .add_diag_info(ODBCError::InvalidAttrValue("SQL_ATTR_ROW_ARRAY_SIZE"));
                    SqlReturn::ERROR
                }
            }
        }
```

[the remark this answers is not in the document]
nbagnard: I think both have the same meaning.
SQLExtendedFetch uses SQL_ROWSET_SIZE and SQLFetchScroll uses SQL_ATTR_ROW_ARRAY_SIZE, but SQLFetchScroll is essentially the ODBC 3.x version of SQLExtendedFetch.
https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlextendedfetch-function?view=sql-server-ver16

It seems fine to keep it as is. We don't support SQLExtendedFetch and if we were to I think we would be Ok with the same internal variable for the row set size. 
A note on that : The signature is not in functions.rs and that could be potentially be an issue when the DM tries to pass the call to the driver.

mattChiaravalloti (approved):

jchemburkar (approved):
lgtm!

odbc/src/api/functions.rs
```diff
@@ -337,13 +337,113 @@ fn sql_alloc_handle(
...
+            if stmt.attributes.read().unwrap().row_array_size != 1 {
+                let mongo_handle = MongoHandleRef::from(hstmt);
+                add_diag_info!(
+                    mongo_handle,
+                    ODBCError::Unimplemented("`column binding with arrays`")
+                );
+                return SqlReturn::ERROR;
+            }
+
+            // makes sure that col_number is in bounds
```
nbagnard, line 390:
```suggestion
            // makes sure that a query was executed/prepared and the number of columns for the resultset is known
```

odbc/src/api/functions.rs
```diff
@@ -337,13 +337,113 @@ fn sql_alloc_handle(
...
-    unimpl!(hstmt);
+    panic_safe_exec_keep_diagnostics!(
+        debug,
+        || {
+            let mongo_handle = MongoHandleRef::from(hstmt);
+            let stmt = must_be_valid!((*mongo_handle).as_statement());
+
+            // Currently, we only support column binding of one row at a time.
+            // Make sure that column-wise binding is being used.
+            if stmt.attributes.read().unwrap().row_bind_type != BindType::SQL_BIND_BY_COLUMN as ULen
```
nbagnard, line 354: [nit] I didn't see a check that the following 3 error paths are tested.

nbagnard (approved):
LGTM
Note:  Integration tests will be added via the upcoming tools flows tickets

EthanHardyMongo:
The current evergreen failures are unrelated to this PR.

PR #188, opened 2024-03-05
SQL-1867: implement additional connection attributes

jchemburkar:
These connection and column attributes allow for SSMS to relay correct errors with a query (both from the application and driver side). Previously, it would output cached errors caused by these.

### DIFF
--- a/odbc/src/api/functions.rs
+++ b/odbc/src/api/functions.rs
@@ -3075,6 +3075,16 @@ unsafe fn set_connect_attrw_helper(
                 conn.attributes.write().unwrap().current_catalog = Some(current_db);
                 SqlReturn::SUCCESS
             }
+            // we use 0 (no timeout throughout the driver); only allow the user to set this value if they are setting to 0
+            ConnectionAttribute::SQL_ATTR_CONNECTION_TIMEOUT => match (value_ptr as u32) == 0 {
+                true => SqlReturn::SUCCESS,
+                false => {
+                    conn_handle.add_diag_info(ODBCError::GeneralWarning(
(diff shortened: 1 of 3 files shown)


nbagnard (approved):
LGTM

odbc/src/api/functions.rs
```diff
@@ -1454,6 +1456,14 @@ unsafe fn sql_get_connect_attrw_helper(
                 let login_timeout = attributes.login_timeout.unwrap_or(0);
                 i32_len::set_output_fixed_data(&login_timeout, value_ptr, string_length_ptr)
             }
+            // according to the spec, SQL_ATTR_CONNECTION_DEAD just returns the latest status of the connection, not the current status
+            ConnectionAttribute::SQL_ATTR_CONNECTION_DEAD => {
+                let connection_dead = match *conn.mongo_connection.write().unwrap() {
+                    Some(_) => SqlBool::SQL_FALSE,
+                    None => SqlBool::SQL_TRUE,
+                };
```
EthanHardyMongo, line 1464:
```suggestion
            ConnectionAttribute::SQL_ATTR_CONNECTION_DEAD => {
                let connection_dead = match *conn.mongo_connection.read().unwrap() {
                    Some(_) => SqlBool::SQL_FALSE,
                    None => SqlBool::SQL_TRUE,
                };
```
I don't think a write is necessary here

EthanHardyMongo (approved):
LGTM! I just had one, tiny nit.

odbc/src/api/functions.rs
jchemburkar: good catch. I initially was pinging the database with the connection (hence the write) but realized that was more then the spec was asking for

PR #190, opened 2024-03-12
SQL-1975: Set working ADF version in run_adf to pass integration tests

bucaojit:
Sets to an ADF version prior to the changes to require configuration values. 
Passing tests:
https://evergreen.mongodb.com/version/65f0858c1e2d1742684714e4

### DIFF
--- a/resources/run_adf.sh
+++ b/resources/run_adf.sh
@@ -364,6 +364,11 @@ if [ $ARG = $START ]; then
         fi
         cd $MONGOHOUSE_DIR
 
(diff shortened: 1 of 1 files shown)


mattChiaravalloti (approved):
Nice 👍 thanks for figuring out a version that works for us

pmeredit (approved):
LGTM, though this does mean we could miss a breaking change in ADF

bucaojit:
@pmeredit Yeah, we can schedule SQL-1976 at triage to address the actual ADF updates to not miss any breaking changes.

PR #192, opened 2024-03-13
SQL-1974: Support column wise binding of rowset in SQLBindCol

EthanHardyMongo:
This PR adds support for using `SQLBindCol` with rowsets with size > 1. The most important/complicated work was done in `sql_fetch_helper`, so that part needs to get reviewed the most.


core/src/query.rs
```diff
@@ -131,6 +137,14 @@ impl MongoStatement for MongoQuery {
...
         let options = if self.query_timeout.is_some_and(|timeout| timeout > 0) {
             opt.max_time(Duration::from_millis(self.query_timeout.unwrap() as u64))
+                .batch_size({
+                    // Natacha and I aren't actually sure what the default batch size is, but we think it might be like 101.
+                    if rowset_size > 101 {
+                        rowset_size as u32
+                    } else {
+                        101
+                    }
+                })
```
EthanHardyMongo, line 147: The idea here is to only update the batch size when `rowset_size` is greater than the default batch size. Otherwise, we may update the batch size to be a very small number.

nbagnard: Yes, although, I couldn't find that default anywhere in the code.
But here is the doc : [https://www.mongodb.com/docs/drivers/rust/current/fundamentals/crud/read-operations/retrieve/#modify-aggregation-behavior](url)
I think that not setting any value if it's below a threshold that we deem suitable, then it will be better.
I don't think that "forcing" the default when not needed is necessary. Especially since we are not sure the default value we use here is 100% correct.

odbc/src/api/functions.rs
```diff
@@ -1279,72 +1270,88 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+                    function_name.to_string()
                 );
+            });
+            if !has_next {
+                stmt_attrs.row_index_is_valid = false;
+                // No more rows
+                return SqlReturn::NO_DATA;
+            }
+            stmt_attrs.row_index_is_valid = true;
+
```
jchemburkar, line 1306:
```suggestion
            if !has_next {
                // No more rows
                return SqlReturn::NO_DATA;
            }

```

I quickly tried SSIS with your branch and things looked good 😄 I had just noticed that we were hanging at this point, and I dont see if/were we use these values elsewhere, so maybe we should just remove the setting of them? (we'll also need to update setting row array size, but otherwise I used the branch as is)

I didn't examine it closely, but it seems to work great. Would be happy to take a closer look if/when this goes into review

core/src/query.rs
EthanHardyMongo: Yeah good point. Do we know what the threshold should be?

odbc/src/api/functions.rs
EthanHardyMongo: The problem is probably due to this line `let mut stmt_attrs = stmt.attributes.write().unwrap();` and this line `*stmt.var_data_cache.write().unwrap() = Some(HashMap::new());` after setting `row_index_is_valid`. I think I create 2 write accesses which causes a dead lock.

core/src/query.rs
```diff
@@ -118,7 +119,12 @@ impl MongoStatement for MongoQuery {
     // Execute the $sql aggregation for the query and initialize the result set
     // cursor. If there is a timeout, the query must finish before the timeout
     // or an error is returned.
-    fn execute(&mut self, connection: &MongoConnection, stmt_id: Bson) -> Result<bool> {
+    fn execute(
+        &mut self,
+        connection: &MongoConnection,
+        stmt_id: Bson,
+        rowset_size: ULen,
```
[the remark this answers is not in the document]
EthanHardyMongo: I made the maximum `u32::MAX` for now.

odbc/src/api/bind_col_tests.rs
```diff
@@ -466,4 +403,832 @@ mod unit {
...
+                .unwrap()
+                .as_ref()
+                .unwrap()
+                .get(&2)
+                .unwrap()
+                .length_or_indicator;
+            assert_eq!(8, *indicator);
+            assert_eq!(8, *((indicator as ULen + 8) as *mut Len));
+            assert_eq!(8, *((indicator as ULen + 16) as *mut Len));
+            assert_eq!(8, *((indicator as ULen + 24) as *mut Len));
```
EthanHardyMongo, line 568: For some reason, I get `16` when running this locally, but I get `8` when running on evergreen. This also happens with this test: `api::get_info_tests::unit::sql_database_name`; however, I never modified anything that touches that test. I think a `WideChar` can either be 2 or 4 bytes, so that might be what's going on.

odbc/src/api/bind_col_tests.rs
```diff
@@ -466,4 +403,832 @@ mod unit {
             let _ = Box::from_raw(buffer as *mut WChar);
         }
     }
+
+    #[test]
+    fn test_binding_arrays_when_rowset_size_evenly_divides_resultset_size() {
+        // Set up MongoHandle
```
EthanHardyMongo, line 409: So far I've only added tests that successfully fetch data and bind columns. I didn't add any tests for error handling yet because there's a lot of possible errors that could come up so I wasn't sure what the best way to test that would be. For example, the cases where one row causes an error, every row is an error, a row can be fetched but then binding fails, etc.

odbc/src/api/functions.rs
```diff
@@ -1279,72 +1274,183 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+            global_warnings_opt.append(&mut warnings_opt);
+
+            if has_rows_fetched_buffer {
+                *stmt.attributes.write().unwrap().rows_fetched_ptr += 1;
+            }
+
+            *stmt.var_data_cache.write().unwrap() = Some(HashMap::new());
+
+            // If there are bound columns, then copy data from the result set into the bound buffers.
+            if let Some(bound_cols) = stmt.bound_cols.read().unwrap().as_ref() {
```
EthanHardyMongo, line 1369: Since `sql_fetch_helper` is getting pretty long, I could put the column binding part in a separate helper function if that would be preferable.

jchemburkar: I would be in favor of this with readability in mind!

jchemburkar (changes requested):
This feature is huge and looks like you've done a great job with it 💪 I especially appreciate all the comments, it made it easy to follow along.

I added some thoughts throughout, mainly focussed on breaking up some of the bigger code blocks like you suggested!

core/src/query.rs
```diff
@@ -130,10 +135,22 @@ impl MongoStatement for MongoQuery {
...
+                opt_with_timeout.build()
+            }
         } else {
-            opt.build()
+            // 100 is an arbitrary value
+            if rowset_size > 100 {
+                opt.batch_size(rowset_size).build()
+            } else {
+                opt.build()
+            }
```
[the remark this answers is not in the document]
nbagnard: You can actually do something like this:
```
        let mut max_time = None;
        if self.query_timeout.is_some_and(|timeout| timeout > 0) {
            max_time = Some(Duration::from_millis((nanos/10000) as u64))
        }
        let mut batch_size = None;
        if rowset_size > 100 {
            batch_size = Some(rowset_size)
        }
        let options = opt.max_time(max_time).batch_size(batch_size).build();
```

odbc/src/api/bind_col_tests.rs
```diff
@@ -466,4 +403,832 @@ mod unit {
...
+            assert_eq!(4, *indicator);
+            assert_eq!(4, *((indicator as ULen + 8) as *mut Len));
+            assert_eq!(4, *((indicator as ULen + 16) as *mut Len));
+            assert_eq!(4, *((indicator as ULen + 24) as *mut Len));
+
+            // assert that SQLFetch is successful. We are fetching the next 4 rows in the result set.
+            assert_eq!(SqlReturn::SUCCESS, SQLFetch(stmt as *mut _));
+
+            // assert that the last 4 values from the result set were put in the bound buffer array correctly for column 1
+            let bound_buffer = s
```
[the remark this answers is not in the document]
EthanHardyMongo: I can actually remove all the "getting" I do in all these tests if we would prefer it that way. The "getting" isn't actually necessary for any of the tests since I create all the references at the top of the tests. I just did that to check everything in the process; however, since the likelihood of this part actually catching an issue is so small, I can just remove it.

odbc/src/api/bind_col_tests.rs
```diff
@@ -466,4 +403,832 @@ mod unit {
...
+            s.attributes.write().unwrap().row_bind_type = BindType::SQL_BIND_BY_COLUMN as usize;
+
+            // row_status_ptr needs 4 slots since the rowset size is 4.
+            s.attributes.write().unwrap().row_status_ptr =
+                Box::into_raw(Box::new([0u16; 4])) as *mut USmallInt;
+            s.attributes.write().unwrap().rows_fetched_ptr =
+                Box::into_raw(Box::new(0usize)) as *mut ULen;
+
+            // create a mongo query with data that corresponds to the bound columns.
+            let mock_query = MongoQuery::new(
```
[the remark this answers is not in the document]
EthanHardyMongo: Yeah that's a good idea!

odbc/src/api/bind_col_tests.rs
```diff
@@ -466,4 +403,832 @@ mod unit {
...
+                .get(&1)
+                .unwrap()
+                .length_or_indicator;
+            assert_eq!(4, *indicator);
+            assert_eq!(4, *((indicator as ULen + 8) as *mut Len));
+            assert_eq!(4, *((indicator as ULen + 16) as *mut Len));
+            assert_eq!(4, *((indicator as ULen + 24) as *mut Len));
+
+            // assert that the first 4 values from the result set were put in the bound buffer array correctly for column 2.
+            let bound_buffer = s
```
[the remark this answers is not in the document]
EthanHardyMongo: My reply to the comment below talks more about this, but I can just remove all the "getting" I do and just use the reference I create at the beginning of the test.

definitions/src/row_status.rs
```diff
@@ -0,0 +1,6 @@
+use crate::USmallInt;
+
+pub const SQL_ROW_SUCCESS: USmallInt = 0;
```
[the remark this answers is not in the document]
EthanHardyMongo: I'm good with putting this wherever (also an enum or constant would work fine), but I'm confused on what our convention actually is haha. It seems like some things are constants, other things are enums, some things have their own file, and other things are in the `lib.rs`? But I can't tell when we should do one or the other haha.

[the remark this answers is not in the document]
jchemburkar: yeah "convention" is probably too heavy handed a word lol, mainly because we forked from odbc-sys but are now handling it ourselves. I think, in general, a good convention that we can stick with / use when adding things is probably just all groups of values should be enums, and all singular values (or things like bitmasks) should be constants.

in this case, looking at the [spec](https://github.com/microsoft/ODBC-Specification/blob/4dda95986bda5d3b55d7749315d3e5a0951c1e50/Windows/inc/sqlext.h#L1970) headers, since this is clearly a group of values I think an enum would be good. We don't need to include all the values since we're defining them anyways (like `SQL_ROW_DELETE` doesn't need to get added eg), so something like the following might be the most consistent option:
```
#[allow(non_camel_case_types)]
#[derive(Debug, PartialEq, Eq, Clone, Copy)]
#[repr(u16)]
pub enum ParamType {
    SQL_ROW_SUCCESS = 0,
    SQL_ROW_NOROW = 3,
    SQL_ROW_ERROR = 5,
    SQL_ROW_SUCCESS_WITH_INFO = 6
}
```

odbc/src/api/bind_col_tests.rs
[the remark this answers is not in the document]
EthanHardyMongo: I condensed all the stuff that's easy to condense. Further condensing would require more work since I would need to make macros and deal with lifetime stuff. I can continue condensing if that's preferable, but since the test file is already much shorter, I figured I would stop here for now and check if we want to do that.

definitions/src/row_status.rs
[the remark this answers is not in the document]
EthanHardyMongo: Ok that makes sense haha. Yeah I agree, an enum would be good for those values.

odbc/src/api/functions.rs
```diff
@@ -1279,73 +1274,211 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+            SqlReturn::SUCCESS_WITH_INFO => {
+                encountered_success_with_info_during_col_binding = true;
+            }
+            _ => {}
+        }
     }
+
+    if !row_status_buffer.is_null() {
+        // It's possible that one binding causes SUCCESS_WITH_INFO and another causes ERROR, so we need to check for an error first
+        // and ignore SUCCESS_WITH_INFO if it also occurs.
```
EthanHardyMongo, line 1470: It's not actually specified anywhere what to do if both happen, but I assume this is correct. I would think ERROR would take precedence over SUCCESS_WITH_INFO.

core/src/query.rs
```diff
@@ -128,13 +133,19 @@ impl MongoStatement for MongoQuery {
...
-            opt.build()
-        };
+        if self.query_timeout.is_some_and(|timeout| timeout > 0) {
+            max_time = Some(Duration::from_millis(self.query_timeout.unwrap() as u64))
+        }
+
+        let mut batch_size = None;
+        // 100 is an arbitrary value
+        if rowset_size > 100 {
+            batch_size = Some(rowset_size)
```
EthanHardyMongo, line 146: We still need to figure out what value to use here. I'm just arbitrarily using 100 at the moment.

EthanHardyMongo:
I've been reviewing the logic, and the only issue I could find was with handling errors that apply to the entire function (e.g., SQLSTATE HYT00 (Timeout expired)) vs errors that only apply to one row (e.g.,  SQLSTATE 22012 (Division by zero)). 

From the SQLFetch docs:
> If an error applies to the entire function, such as SQLSTATE HYT00 (Timeout expired) or SQLSTATE 24000 (Invalid cursor state), SQLFetch returns SQL_ERROR and the applicable SQLSTATE. The contents of the rowset buffers are undefined and the cursor position is unchanged.

Currently, the error handling is set up to only return `SQL_ERROR` if every row has an error to match this part of the docs:
> SQLFetch continues fetching rows until it has fetched all the rows in the rowset. It returns SQL_SUCCESS_WITH_INFO unless an error occurs in every row of the rowset (not including rows with status SQL_ROW_NOROW), in which case it returns SQL_ERROR. In particular, if the rowset size is 1 and an error occurs in that row, SQLFetch returns SQL_ERROR.

Currently, SQLFetch is set up to loop `row_array_size` number of times, so to handle this, we could make a list of errors that require an immediate `SQL_ERROR` return, and then have a check for those errors in the function. Otherwise, `SQL_ERROR` may not get returned because the error may not happen on every row (unless I'm missing something here). Is this correct? If so, what all types of errors would I need to check for?

odbc/src/api/functions.rs
```diff
@@ -1279,73 +1274,218 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+                }
+
+                // if the cursor has already reached the end of the result set when SQLFetch is called, NO_DATA is returned.
+                // Additionally, rows_fetch_ptr is set to 0.
+                return SqlReturn::NO_DATA;
+            }
+
+            has_fetched_at_least_one_row = true;
+
+            if !row_status_buffer.is_null() {
```
nbagnard, line 1356: If you use a marker, here you can do 
```
if track_row_status {
   let row_status_buffer: *mut USmallInt = (stmt.attributes.read().unwrap().row_status_ptr as ULen + (index * size_of::<u16>())) as *mut USmallInt;
    *row_status_buffer = if warnings_opt.is_empty() {
        RowStatus::SQL_ROW_SUCCESS as USmallInt
    } else {
        RowStatus::SQL_ROW_SUCCESS_WITH_INFO as USmallInt
    };
}
```

odbc/src/api/functions.rs
```diff
@@ -1279,73 +1274,218 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+                    RowStatus::SQL_ROW_SUCCESS as USmallInt
+                } else {
+                    RowStatus::SQL_ROW_SUCCESS_WITH_INFO as USmallInt
+                };
+            }
+
+            // keep track of all warnings that occur throughout the entire function.
+            global_warnings_opt.append(&mut warnings_opt);
+
+            if has_rows_fetched_buffer {
```
nbagnard, line 1367: Does it make sense updating pointer as you go or could it wait until you get the SQL return part and you just do `*stmt.attributes.write().unwrap().rows_fetched_ptr = fetched_rows`?

odbc/src/api/functions.rs
```diff
@@ -1279,73 +1274,218 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+unsafe fn sql_fetch_column_binding_helper(
+    statement_handle: HStmt,
+    index: ULen,
+    row_status_buffer: *mut USmallInt,
+    bound_cols: &HashMap<USmallInt, BoundColInfo>,
+    function_name: &str,
+) -> (bool, bool) {
+    let mongo_handle_for_sql_get_data_helper = MongoHandleRef::from(statement_handle);
+
+    let mut encountered_error_during_col_binding = false;
```
nbagnard, line 1441:
```suggestion
    let mut encountered_error_getting_data_ = false;
```

odbc/src/api/functions.rs
```diff
@@ -1279,73 +1274,218 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+    let mut encountered_success_with_info_during_col_binding = false;
```
nbagnard, line 1442:
```suggestion
    let mut encountered_success_with_info_getting_data= false;
```

nbagnard (changes requested):

core/src/stmt.rs
```diff
@@ -25,7 +25,7 @@ pub trait MongoStatement: Debug {
     }
     // Executes a prepared statement.
     // Only MongoQuery supports this workflow. The other statements don't.
-    fn execute(&mut self, _: &MongoConnection, _: Bson) -> Result<bool> {
+    fn execute(&mut self, _: &MongoConnection, _: Bson, _: u32) -> Result<bool> {
```
nbagnard, line 28: Can we do this to properly keep track of what the arguments are, even here?
```suggestion
    fn execute(&mut self, _connection: &MongoConnection, _stmt_id: Bson, _rowset_size: u32) -> Result<bool> {
```

nbagnard:
> I've been reviewing the logic, and the only issue I could find was with handling errors that apply to the entire function (e.g., SQLSTATE HYT00 (Timeout expired)) vs errors that only apply to one row (e.g., SQLSTATE 22012 (Division by zero)).
> 
> From the SQLFetch docs:
> 
> > If an error applies to the entire function, such as SQLSTATE HYT00 (Timeout expired) or SQLSTATE 24000 (Invalid cursor state), SQLFetch returns SQL_ERROR and the applicable SQLSTATE. The contents of the rowset buffers are undefined and the cursor position is unchanged.
> 
> Currently, the error handling is set up to only return `SQL_ERROR` if every row has an error to match this part of the docs:
> 
> > SQLFetch continues fetching rows until it has fetched all the rows in the rowset. It returns SQL_SUCCESS_WITH_INFO unless an error occurs in every row of the rowset (not including rows with status SQL_ROW_NOROW), in which case it returns SQL_ERROR. In particular, if the rowset size is 1 and an error occurs in that row, SQLFetch returns SQL_ERROR.
> 
> Currently, SQLFetch is set up to loop `row_array_size` number of times, so to handle this, we could make a list of errors that require an immediate `SQL_ERROR` return, and then have a check for those errors in the function. Otherwise, `SQL_ERROR` may not get returned because the error may not happen on every row (unless I'm missing something here). Is this correct? If so, what all types of errors would I need to check for?

Yes, this makes sense.
You would break out of the loop when a "global" error like 24000 happens.
I think it also looks like that because we are running a loop for rowset, we should check for cancellation which I don't think we are doing.
```Or, the SQLFetch function was called, and before it completed execution, SQLCancel or SQLCancelHandle was called on the StatementHandle from a different thread in a multithread application.
```
And `Operation canceled` would be a general error.

odbc/src/api/functions.rs
```diff
@@ -1279,73 +1274,211 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+                    ODBCError::GeneralWarning(warning.to_string()),
+                    function_name.to_string()
+                );
+            });
+
+            if !has_next {
+                // There are no more rows, so set the rest of Row Status Array buffers to SQL_ROW_NOROW if the rowset end has not been reached.
+                if !row_status_buffer.is_null() {
+                    for row_status_index in index..rowset_size {
+                        let row_status_buffer = (stmt.attributes.read().unwrap().row_status_ptr
```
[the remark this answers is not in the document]
EthanHardyMongo: Natacha and I discussed this and decided that we have as much protection as we can get. Any memory issues that happen here (or with any of the other buffers that we use pointer arithmetic to iterate through) will occur due to user error and aren't our fault. It's the user's responsibility to ensure that they provide a buffer of sufficient size.

odbc/src/api/functions.rs
```diff
@@ -1279,73 +1274,211 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+                            as ULen
+                            + (row_status_index * size_of::<u16>()))
+                            as *mut USmallInt;
+                        *row_status_buffer = definitions::SQL_ROW_NOROW;
                     }
-                    _ => {}
+                }
+
+                // break here to prevent NO_DATA from being returned if the rowset had at least one row in it.
+                if has_fetched_at_least_one_row {
```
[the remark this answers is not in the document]
EthanHardyMongo: I need to break no matter what since there are no rows left. So it would have to be like this if we change it:
```
if !has_fetched_at_least_one_row {
return SqlReturn::NO_DATA;
}else{
  break;
}
```

odbc/src/api/functions.rs
```diff
@@ -1279,73 +1274,218 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+        if let Ok((has_next, mut warnings_opt)) = move_to_next_result {
+            warnings_opt.iter().for_each(|warning| {
+                add_diag_with_function!(
+                    MongoHandleRef::from(statement_handle),
+                    ODBCError::GeneralWarning(warning.to_string()),
+                    function_name.to_string()
+                );
+            });
+
+            if !has_next {
```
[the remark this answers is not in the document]
EthanHardyMongo: I have to be able to handle the case where the user is not using the `row status array`, so I don't think I can do this without making an `else if !has_next {...}` which would then cause code duplication.

odbc/src/api/functions.rs
EthanHardyMongo: I made a marker called `has_row_status_array`, but is it really preferable to define `row_status_buffer` within every `if has_row_status_array {...}` block rather than just defining it once like I am already doing. (i.e., 
```
        let row_status_buffer: *mut USmallInt =
            if has_row_status_array {
                (stmt.attributes.read().unwrap().row_status_ptr as ULen
                    + (index * size_of::<u16>())) as *mut USmallInt
            } else {
                null_mut()
            };
```
).
I'm open to doing it either way, but the current way avoids code duplication.

odbc/src/api/functions.rs
```diff
@@ -1279,73 +1274,218 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+                Some(mongo_stmt) => mongo_stmt
+                    .next(connection.mongo_connection.read().unwrap().as_ref())
+                    .map_err(|e| e.into()),
+                None => Err(ODBCError::InvalidCursorState),
+            }
+        };
+
+        // if row_status_ptr is null, it was not set by the user and can be ignored.
+        // Otherwise, we need to set the row_status_buffer to the address that corresponds with the current row.
+        let row_status_buffer: *mut USmallInt =
```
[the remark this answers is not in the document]
EthanHardyMongo: I made a marker called `has_row_status_array`, and I talk more about it in a reply to a comment below.

EthanHardyMongo:
> > I've been reviewing the logic, and the only issue I could find was with handling errors that apply to the entire function (e.g., SQLSTATE HYT00 (Timeout expired)) vs errors that only apply to one row (e.g., SQLSTATE 22012 (Division by zero)).
> > From the SQLFetch docs:
> > > If an error applies to the entire function, such as SQLSTATE HYT00 (Timeout expired) or SQLSTATE 24000 (Invalid cursor state), SQLFetch returns SQL_ERROR and the applicable SQLSTATE. The contents of the rowset buffers are undefined and the cursor position is unchanged.
> > 
> > 
> > Currently, the error handling is set up to only return `SQL_ERROR` if every row has an error to match this part of the docs:
> > > SQLFetch continues fetching rows until it has fetched all the rows in the rowset. It returns SQL_SUCCESS_WITH_INFO unless an error occurs in every row of the rowset (not including rows with status SQL_ROW_NOROW), in which case it returns SQL_ERROR. In particular, if the rowset size is 1 and an error occurs in that row, SQLFetch returns SQL_ERROR.
> > 
> > 
> > Currently, SQLFetch is set up to loop `row_array_size` number of times, so to handle this, we could make a list of errors that require an immediate `SQL_ERROR` return, and then have a check for those errors in the function. Otherwise, `SQL_ERROR` may not get returned because the error may not happen on every row (unless I'm missing something here). Is this correct? If so, what all types of errors would I need to check for?
> 
> Yes, this makes sense. You would break out of the loop when a "global" error like 24000 happens. I think it also looks like that because we are running a loop for rowset, we should check for cancellation which I don't think we are doing.
> 
> And `Operation canceled` would be a general error.

Why would `Operation canceled` be a general error rather than `ODBCError::Core(Error::QueryCancelled)`?

odbc/src/api/functions.rs
```diff
@@ -1312,73 +1307,235 @@ pub unsafe extern "C" fn SQLFetch(statement_handle: HStmt) -> SqlReturn {
...
+            let error = move_to_next_result.unwrap_err();
+
+            // Checks if there is an error that applies to the entire function instead of just one row.
+            // If there is, early exit with SqlReturn::Error.
+            if matches!(
+                error,
+                ODBCError::InvalidCursorState | ODBCError::Core(Error::QueryCancelled)
+            ) {
+                return SqlReturn::ERROR;
+            }
```
EthanHardyMongo, line 1449: This is what I made so far to check for errors that apply to the entire function and require an early exit. What other errors should I be checking for?

nbagnard:
> Why would `Operation canceled` be a general error rather than `ODBCError::Core(Error::QueryCancelled)`?

By general error I mean that it applies to the entire SQLFetch and not to a row. Sorry for the confusion here.
Yes, `ODBCError::Core(Error::QueryCancelled)` for the error itself is good.

odbc/src/api/functions.rs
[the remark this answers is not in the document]
nbagnard: oh, yes, I should open the file on the side to see the correct indentation.
Disregard this comment and the next.

odbc/src/api/bind_col_tests.rs
EthanHardyMongo: Ticket: https://jira.mongodb.org/browse/SQL-2010

nbagnard (approved):
LGTM!
Great job implementing rowset binding 💯 
Can you please just make sure that you create the follow up JIRAs for cancelling SQLFetch (plus reference it in the code) and adding test cases for failure cases.

jchemburkar (approved):
Great work implementing this + figuring out all the edge cases, looks good with the comments addressed!

I also merged your branch with nathan's async branch and tested it with SSIS that way, everything seems to be working good! 🥳

odbc/src/api/bind_col_tests.rs
```diff
@@ -466,4 +354,478 @@ mod unit {
             let _ = Box::from_raw(buffer as *mut WChar);
         }
     }
+
+    #[test]
+    fn test_binding_arrays_when_rowset_size_evenly_divides_resultset_size() {
```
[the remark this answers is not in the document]
EthanHardyMongo: I'll add this to my ticket regarding adding more test cases instead of doing it as part of this PR. I think it's better this way since there will probably be more refactor opportunities whenever I add more test cases anyways.

PR #193, opened 2024-03-19
SQL-1976: Update ADF config to add new required values

bucaojit:
Added  `currentOp:` and `query: killOp:` configuration  to the latest [ADF config](https://github.com/10gen/mongohouse/blob/master/testdata/config/inline_local/frontend-agent-backend.yaml) for inline_local. 

@pmeredit the ADF config for the ODBC repo has changes to enable match filters after currentOp that Jonathan added.  This config is different from the JDBC/Power BI repos so that could be one reason to keep the run_adf scripts separate.

### DIFF
--- a/resources/integration_test/config/adf_config.yaml
+++ b/resources/integration_test/config/adf_config.yaml
@@ -1,34 +1,33 @@
 hostProvider: local
 hostRegion: local
 environment: local
-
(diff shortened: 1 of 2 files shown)


pmeredit (approved):

jchemburkar (approved):
lgtm

PR #194, opened 2024-03-21
SQL-1962: Add integration tests for SSIS preview data flow

mattChiaravalloti:
This PR adds integration tests for the SSIS preview data flow. I broke it down into two tests since SSIS has a tendency to free all handles between discrete units of work. As discussed individually, I attempted to make these tests _slightly_ more general than the exact SSIS workflow. Instead of naming them "ssis_tests" I called the file "free_and_cancel_tests.rs" since this is the first set of integration tests that focus on the newly implemented `SQLFreeStmt` and `SQLCancel` functions. I am open to modifying the names/coverage as requested.

### DIFF
--- a/integration_test/src/test_runner/mod.rs
+++ b/integration_test/src/test_runner/mod.rs
@@ -177,7 +177,7 @@ pub fn run_resultset_tests(generate: bool) -> Result<()> {
 /// the test results should be written to a file for baseline test file
 /// generation, or be asserted for correctness.
 pub fn run_resultset_tests_odbc_2(generate: bool) -> Result<()> {
-    let env = allocate_env(AttrOdbcVersion::SQL_OV_ODBC3);
+    let env = allocate_env(AttrOdbcVersion::SQL_OV_ODBC2);
     unsafe {
         assert_eq!(
             SqlReturn::SUCCESS,
--- a/integration_test/src/test_runner/mod.rs
+++ b/integration_test/src/test_runner/mod.rs
@@ -2,8 +2,8 @@ mod test_generator_util;
 
 use cstr::WideChar;
 use definitions::{
-    CDataType, Desc, EnvironmentAttribute, HDbc, HStmt, Handle, HandleType, SmallInt, SqlReturn,
-    USmallInt,
+    AttrOdbcVersion, CDataType, Desc, EnvironmentAttribute, HDbc, HStmt, Handle, HandleType,
+    SmallInt, SqlReturn, USmallInt,
 };
 use serde::{Deserialize, Serialize};
 use serde_json::json;
@@ -141,10 +141,10 @@ pub fn odbc2_resultset_tests() -> Result<()> {
 }
 
 /// Run an integration test. The generate argument indicates whether
-/// the test results should written to a file for baseline test file
+/// the test results should be written to a file for baseline test file
 /// generation, or be asserted for correctness.
 pub fn run_resultset_tests(generate: bool) -> Result<()> {
-    let env = allocate_env().unwrap();
+    let env = allocate_env(AttrOdbcVersion::SQL_OV_ODBC3);
     let paths = load_file_paths(PathBuf::from(TEST_FILE_DIR)).unwrap();
     for path in paths {
         let yaml = parse_test_file_yaml(&path).unwrap();
@@ -174,10 +174,10 @@ pub fn run_resultset_tests(generate: bool) -> Result<()> {
 }
 
 /// Runs odbc 2 compatibility integration test. The generate argument indicates whether
-/// the test results should written to a file for baseline test file
+/// the test results should be written to a file for baseline test file
 /// generation, or be asserted for correctness.
 pub fn run_resultset_tests_odbc_2(generate: bool) -> Result<()> {
-    let env = allocate_env().unwrap();
+    let env = allocate_env(AttrOdbcVersion::SQL_OV_ODBC3);
     unsafe {
         assert_eq!(
             SqlReturn::SUCCESS,
--- a/integration_test/tests/common/mod.rs
+++ b/integration_test/tests/common/mod.rs
@@ -152,7 +152,9 @@ pub fn allocate_env(odbc_version: AttrOdbcVersion) -> HEnv {
                 HandleType::SQL_HANDLE_ENV,
                 null_mut(),
                 &mut env as *mut Handle,
-            )
+            ),
+            "{}",
+            get_sql_diagnostics(HandleType::SQL_HANDLE_ENV, env as Handle)
         );
         assert_eq!(
             SqlReturn::SUCCESS,
@@ -161,7 +163,9 @@ pub fn allocate_env(odbc_version: AttrOdbcVersion) -> HEnv {
                 EnvironmentAttribute::SQL_ATTR_ODBC_VERSION,
                 odbc_version.into(),
                 0
-            )
+            ),
+            "{}",
+            get_sql_diagnostics(HandleType::SQL_HANDLE_ENV, env as Handle)
         );
     }
     env as HEnv
--- a/integration_test/tests/common/mod.rs
+++ b/integration_test/tests/common/mod.rs
@@ -127,30 +127,50 @@ pub fn sql_return_to_string(return_code: SqlReturn) -> String {
 }
 
 #[allow(dead_code)]
-// Allocates new environment variable
-pub fn allocate_env() -> Result<HEnv> {
+/// Setup flow that connects and allocates a statement. This allocates a new
+/// environment handle, sets the ODBC_VERSION environment attribute, connects
+/// using the provided (or default) URI, and allocates a statement. The flow
+/// is:
+///     - SQLAllocHandle(SQL_HANDLE_ENV)
+///     - SQLSetEnvAttr(SQL_ATTR_ODBC_VERSION, SQL_OV_ODBC3)
+///     - SQLAllocHandle(SQL_HANDLE_DBC)
+///     - SQLDriverConnectW
+///     - SQLAllocHandle(SQL_HANDLE_STMT)
+pub fn default_setup_connect_and_alloc_stmt(
+    odbc_version_value: AttrOdbcVersion,
+) -> (HEnv, HDbc, HStmt) {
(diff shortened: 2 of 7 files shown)


integration_test/tests/free_and_cancel_tests.rs
```diff
@@ -0,0 +1,142 @@
...
+    ///     - <loop: for each column>
+    ///         - SQLDescribeColW
+    ///         - SQLColAttributeW(SQL_DESC_UNSIGNED)
+    ///         - SQLColAttributeW(SQL_DESC_UPDATABLE)
+    ///     - SQLFreeStmt(SQL_CLOSE)
+    ///     - SQLDisconnect
+    ///     - SQLFreeHandle(SQL_HANDLE_DBC)
+    ///     - SQLFreeHandle(SQL_HANDLE_ENV)
+    #[test]
+    fn test_free_stmt() {
```
mattChiaravalloti, line 54: The trace includes a few more Get/Set attribute calls that I omitted since they seemed inconsequential to the core of what we're trying to capture (especially all the GetInfo calls). The same idea applies to the next test.

integration_test/tests/free_and_cancel_tests.rs
```diff
@@ -0,0 +1,142 @@
...
+            let row_count_ptr = &mut 0;
+            assert_eq!(SqlReturn::SUCCESS, SQLRowCount(stmt_handle, row_count_ptr));
+
+            get_column_attributes(stmt_handle as Handle, 2, None, false);
+
+            fetch_and_get_data(
+                stmt_handle as Handle,
+                Some(3),
+                vec![SqlReturn::SUCCESS; 2],
+                vec![CDataType::SQL_C_SSHORT, CDataType::SQL_C_SLONG],
```
jchemburkar, line 136: I think this call is right, but also seems like we don't support `SQL_C_SSHORT` in `SQLGetData`. i imagine it wouldn't be _hard_, and if I'm understanding the spec right it seems like we should? looking at [this](https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/c-data-types?view=sql-server-ver16) doc it would seem the types we dont support that we probably could are `SQL_C_SSHORT`, `SQL_C_USHORT`, `SQL_C_STINYINT` and `SQL_C_UTINYINT`; and I imagine the bookmark types we would ignore.

Wondering if either you or @nbagnard have any context on whether there was any reason not to include them, or not implement them now? (current location is [here](https://github.com/mongodb/mongo-odbc-driver/blob/43bb8a95c1a4b1946daa57f1abf8f4dbd29facb3/odbc/src/api/data.rs#L788))

integration_test/tests/common/mod.rs
```diff
@@ -324,17 +357,25 @@ pub fn fetch_and_get_data(
...
 
 #[allow(dead_code)]
-///  Helper function for checking resultset metadata
+/// Helper function for checking result set metadata. It optionally takes a list
+/// of attributes to get. If none is provided, the following defaults will be
+/// checked. It also optionally invokes SQLDescribeColW.
 ///  - SQLNumResultCols()
 ///  - For columns 1 to {numCols}
+///      - (Optional) SQLDescribeColW
 ///      - SQLColAttributeW(SQL_DESC_CONCISE_TYPE)
```
jchemburkar, line 366: I think this could be clarified now with the new logic. Even something like:

`SQLColAttributeW for all attrs; if not attrs: [...list the defaults...]`

maybe?

integration_test/tests/common/mod.rs
```diff
@@ -324,17 +357,25 @@ pub fn fetch_and_get_data(
 }
 
 #[allow(dead_code)]
-///  Helper function for checking resultset metadata
+/// Helper function for checking result set metadata. It optionally takes a list
+/// of attributes to get. If none is provided, the following defaults will be
+/// checked. It also optionally invokes SQLDescribeColW.
```
jchemburkar, line 362: nit: might be bikeshedding and feel free to do whatever you think seems clearest; but could be clearer if this comment points out that this function should get all column metadata, and some tools do this by calling SQLColAttribute for every attribute, and some get the bulk of attributes through SQLDescribeCol then request addtional attributes ad-hoc

jchemburkar (commented):
The overall flow looks good to me, and the places you cut make sense 👍 just had a couple nits

integration_test/tests/free_and_cancel_tests.rs
```diff
@@ -0,0 +1,142 @@
...
+    ///         - SQLColAttributeW(SQL_COLUMN_NULLABLE)
+    ///         - SQLColAttributeW(SQL_DESC_TYPE_NAME)
+    ///         - SQLColAttributeW(SQL_COLUMN_LENGTH)
+    ///         - SQLColAttributeW(SQL_COLUMN_SCALE)
+    ///     - <loop: until SQLFetch return SQL_NO_DATA>
+    ///         - SQLFetch
+    ///         - SQLGetData
+    ///     - SQLCancel
+    #[test]
+    fn test_cancel() {
```
nbagnard, line 103:
```suggestion
    fn test_cancel_noop() {
```
Aka, cancel will be a no-op in our case because the odbc version set for the environment is ODBC 3 and the query is not still executing.

integration_test/tests/free_and_cancel_tests.rs
nbagnard: @jchemburkar 
If I remember correctly we didn't implement them because no BSON type maps to SQL_SMALLINT or SQL_TINYINT so we don't expect tools to want to get the data as C_xxSHORT or SQL_C_xxTINYINT because they would risk truncation in doing that.
We can support this yes and if tools need it, then we should.
As you figured out we "simply" need to add the support for the SQL_C_xxx type in the function you pointed to, map it to the correct rust type and then add to_i16, to_u16, etc....

nbagnard: I looked at the mapping, data in the `foo` collection are Bson int, Bson long, so you should get => SQL_LONG, SQL_BIGINT in the metadata => SQL_C_SLONG, SQL_C_SQL_C_SBIGINT  for the C default/direct mapping=> i32, i64 as Rust type.

integration_test/tests/free_and_cancel_tests.rs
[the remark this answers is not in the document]
mattChiaravalloti: No need for more discussion. I see what you're saying. My initial idea here was that this was still a flow that made actual use of the statement handle (getting metadata) before freeing, but you're right that just calling `SQLPrepareW` is enough to demonstrate that point. I ended up leaving the name as `test_free_stmt` since I cut all the metadata-gathering stuff. Now it just does Prepare then frees the statement.

integration_test/src/test_runner/mod.rs
```diff
@@ -174,10 +174,10 @@ pub fn run_resultset_tests(generate: bool) -> Result<()> {
 }
 
 /// Runs odbc 2 compatibility integration test. The generate argument indicates whether
-/// the test results should written to a file for baseline test file
+/// the test results should be written to a file for baseline test file
 /// generation, or be asserted for correctness.
 pub fn run_resultset_tests_odbc_2(generate: bool) -> Result<()> {
-    let env = allocate_env().unwrap();
+    let env = allocate_env(AttrOdbcVersion::SQL_OV_ODBC3);
```
nbagnard, line 180:
```suggestion
    let env = allocate_env(AttrOdbcVersion::SQL_OV_ODBC2);
```

integration_test/tests/common/mod.rs
```diff
@@ -126,30 +121,50 @@ pub fn sql_return_to_string(return_code: SqlReturn) -> String {
...
-        match SQLSetEnvAttr(
-            env as HEnv,
-            EnvironmentAttribute::SQL_ATTR_ODBC_VERSION,
-            AttrOdbcVersion::SQL_OV_ODBC3.into(),
-            0,
-        ) {
-            SqlReturn::SUCCESS => (),
-            sql_return => return Err(Error::SetEnvAttr(sql_return_to_string(sql_return))),
-        }
+        assert_eq!(
```
nbagnard, line 149: I think that the former construction of errors was done on purpose.
The idea I believe was to know what failed without having to look at the line number. With these asserts, we will only see something like 
```
assertion `left == right` failed
  left: true
 right: false
```
Even if rely on assert now, we should add a more descriptive message.

integration_test/tests/common/mod.rs
```diff
@@ -126,30 +121,50 @@ pub fn sql_return_to_string(return_code: SqlReturn) -> String {
...
+    let env_handle = allocate_env(odbc_version_value);
+    let (conn_handle, stmt_handle) = connect_and_allocate_statement(env_handle, None);
+
+    (env_handle, conn_handle, stmt_handle)
+}
+
+#[allow(dead_code)]
+/// Allocates new environment variable and sets the ODBC_VERSION environment
+/// attribute to the provided odbc_version.
+pub fn allocate_env(odbc_version: AttrOdbcVersion) -> HEnv {
```
nbagnard, line 145: [nit] We only to set the odbc version to ODBC 2.0 in one or 2 places.
I would have left the existing as-is instead of forcing us to pass in SQL_OV_ODBC3 in every call now.
What was bothering you with this function?

nbagnard (changes requested):
I like the clean refactoring per functionality!
It is generally looking good. The main thing I'd like to see if assert! with a dedicated more informative error message.

integration_test/src/test_runner/mod.rs
mattChiaravalloti: Whoops, thanks.

integration_test/tests/common/mod.rs
mattChiaravalloti: Updated all new asserts to include the same detailed error message as the existing ones 👍

integration_test/tests/common/mod.rs
nbagnard: ah, I see now, yes makes sense! 👍

nbagnard (approved):
LGTM! 💯

jchemburkar (approved):
Great work 🔥

PR #195, opened 2024-03-22
SQL-1991: remove all enums from odbc api parameters

jchemburkar:


mattChiaravalloti (approved):
lgtm 🔥

nbagnard (changes requested):
It's looking good.
I only found a few small mismatches.

definitions/src/functions.rs
```diff
@@ -582,7 +580,7 @@ extern "system" {
     pub fn SQLGetDescFieldW(
         descriptor_handle: HDesc,
         record_number: SmallInt,
-        field_identifier: Desc,
+        field_identifier: USmallInt,
```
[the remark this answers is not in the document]
jchemburkar: Ah, I had based these params off of the enum representations -- i.e. [here](https://github.com/mongodb/mongo-odbc-driver/blob/9adb634cf55f892d13400f21ca7b93e82cbefb2a/definitions/src/desc.rs#L3). Should I update the `repr` to be an `i16` eg? or is it ok that they diverge since we are representing the enum by its contained values (rather than the range of possible values in odbc)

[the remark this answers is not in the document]
nbagnard: uhm, we are using the same enum in SQLColAttribute and there FieldIdentifier is a SQLUSMALLINT.
Descriptor fields can be driver specific.
We don't have specific description fields at the moment.
If we wanted to be able to support negative driver specific descriptor fields, then it would be needed, but I don't think we will. There is more than enough possible values left that the ODBC API isn't already using.
There won't be any issues with converting i16 values to u16, none of the out-of-bound ones will map to ODBC defined values.
I think you can leave the enum  representation as is.

odbc/src/api/functions.rs
```diff
@@ -748,10 +748,7 @@ pub unsafe extern "C" fn SQLColAttributeW(
                 None => {
                     let mongo_handle = MongoHandleRef::from(statement_handle);
                     let _ = must_be_valid!((*mongo_handle).as_statement());
-                    add_diag_info!(
-                        mongo_handle,
-                        ODBCError::UnsupportedFieldDescriptor(format!("{field_identifier}"))
-                    );
+                    add_diag_info!(mongo_handle, ODBCError::DriverNotCapable);
```
nbagnard, line 751: You flipped the errors.
If the descriptor is valid (e.i in the list), but we don't support it, we should return HYC00-`DriverNotCapable`
If it is not a know value, we should return HY091 - `Invalid descriptor` (code assigned to UnsupportedFieldDescriptor)

nbagnard (changes requested):
One last little thing

odbc/src/api/col_attr_describe_tests.rs
```diff
@@ -566,7 +566,7 @@ mod unit {
                 .unwrap();
             assert_eq!(errors.len(), 1);
             assert_eq!(
-                "[MongoDB][API] Driver not capable".to_string(),
+                "[MongoDB][API] Invalid field descriptor value 4".to_string(),
```
nbagnard, line 569: 4 ?

jchemburkar: This is the case where the value passed in to descriptor doesn't map to any values in the enum (as opposed to a value in the enum that we dont support), and I thought it would be useful to include the value in the error message. This is a call that SSIS makes:
````
1ee0-1414	ENTER SQLColAttributeW 
		SQLHSTMT            0x00000223EE458400
		SQLSMALLINT                  1 
		SQLSMALLINT                  4 <unknown>
		SQLPOINTER         0x00000223EBAE6A40
		SQLSMALLINT               4096 
		SQLSMALLINT *       0x000000322297A868
		SQLPOINTER          0x000000322297A808
```

nbagnard: oh, right because we don't have a name for it, makes sense, sorry 👍

nbagnard (approved):
LGTM!

PR #196, opened 2024-03-26
SQL-1964: Add integration tests for SSIS query flow

mattChiaravalloti:
This PR adds an integration test for the SSIS query flow. ~This builds on top of #194. Feel free to ignore the first commit -- for convenience, I'll rebase it away once #194 is merged into master.~

I created just one test since really only one unique new behavior is done by running a query in SSIS. The test calls `SQLBindCol` after executing a query, and calls `SQLFreeStmt` after those calls to `SQLBindCol`. This is different from the previous `SQLFreeStmt` test since the previous one did not bind columns prior to freeing the statement.

This is optionally blocked on #192 since I set the `ROW_ARRAY_SIZE` attribute to 1000 (just like SSIS). Technically, I can omit that and we can observe this test passing now. I figure it is ok to leave this in review until after #192 is merged, though.

### DIFF
--- a/integration_test/tests/bind_cols_tests.rs
+++ b/integration_test/tests/bind_cols_tests.rs
@@ -29,9 +29,9 @@ mod integration {
             exec_direct_default_query(stmt_handle);
 
             let id_buffer = &mut [0u8; 4] as *mut _;
-            let id_indicator = &mut [0isize; 2] as *mut Len;
+            let id_indicator = &mut [0isize; 1] as *mut Len;
             let a_buffer = &mut [0u8; 8] as *mut _;
-            let a_indicator = &mut [0isize; 2] as *mut Len;
+            let a_indicator = &mut [0isize; 1] as *mut Len;
             bind_cols(
                 stmt_handle,
                 vec![
--- a/integration_test/tests/bind_cols_tests.rs
+++ b/integration_test/tests/bind_cols_tests.rs
@@ -1,0 +1,77 @@
+mod common;
+
+mod integration {
+    use crate::common::{
+        default_setup_connect_and_alloc_stmt, disconnect_and_free_dbc_and_env_handles,
+        exec_direct_default_query, fetch_and_bind_cols, get_sql_diagnostics,
+    };
+    use definitions::{
+        AttrOdbcVersion, BindType, CDataType, FreeStmtOption, Handle, HandleType, Pointer,
+        SQLFreeStmt, SQLSetStmtAttrW, SqlReturn, StatementAttribute,
+    };
+
+    /// This test is inspired by the SSIS Query flow. It is altered to be
+    /// more general than that specific flow, with a focus on freeing the
+    /// statement handle after calling SQLBindCol. This flow depends on
+    /// default_setup_connect_and_alloc_stmt.
+    /// After allocating a statement handle, the flow is:
+    ///     - SQLSetStmtAttrW(SQL_ATTR_ROW_BIND_TYPE, SQL_BIND_BY_COLUMN)
+    ///     - SQLSetStmtAttrW(SQL_ATTR_ROW_ARRAY_SIZE, 1000)
+    ///     - SQLExecDirectW(<query>)
+    ///     - <loop: until SQLFetchScroll returns SQL_NO_DATA>
+    ///         - SQLBindCol for each column
+    ///         - SQLFetchScroll
+    ///     - SQLFreeStmt(SQL_UNBIND)
+    ///     - SQLDisconnect
+    ///     - SQLFreeHandle(SQL_HANDLE_DBC)
+    ///     - SQLFreeHandle(SQL_HANDLE_ENV)
+    #[test]
+    fn test_free_stmt_after_bind_col() {
+        let (env_handle, conn_handle, stmt_handle) =
+            default_setup_connect_and_alloc_stmt(AttrOdbcVersion::SQL_OV_ODBC3);
+
+        unsafe {
+            let bind_type = BindType::SQL_BIND_BY_COLUMN as i32;
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLSetStmtAttrW(
+                    stmt_handle,
+                    StatementAttribute::SQL_ATTR_ROW_BIND_TYPE,
+                    bind_type as Pointer,
+                    0,
+                ),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, stmt_handle as Handle)
+            );
+
+            let row_array_size = 1000;
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLSetStmtAttrW(
+                    stmt_handle,
+                    StatementAttribute::SQL_ATTR_ROW_ARRAY_SIZE,
+                    row_array_size as Pointer,
+                    0
+                ),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, stmt_handle as Handle)
+            );
+
+            exec_direct_default_query(stmt_handle);
+
+            fetch_and_bind_cols(
+                stmt_handle,
+                vec![CDataType::SQL_C_SLONG, CDataType::SQL_C_SBIGINT],
+            );
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt_handle, FreeStmtOption::SQL_UNBIND),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, stmt_handle as Handle)
+            );
+
+            disconnect_and_free_dbc_and_env_handles(env_handle, conn_handle);
+        }
+    }
+}
--- a/integration_test/tests/bind_cols_tests.rs
+++ b/integration_test/tests/bind_cols_tests.rs
@@ -148,7 +148,7 @@ mod integration {
                 let (a1, a2) = expected_a_data[i];
                 assert_eq!(a1, *(a_buffer as *mut i64));
                 assert_eq!(a2, *((a_buffer as ULen + 8) as *mut i64));
-                assert_eq!(8, *id_indicator);
+                assert_eq!(8, *a_indicator);
 
                 i += 1;
             }
--- a/integration_test/tests/bind_cols_tests.rs
+++ b/integration_test/tests/bind_cols_tests.rs
@@ -2,8 +2,8 @@ mod common;
 
 mod integration {
     use crate::common::{
-        bind_cols, default_setup_connect_and_alloc_stmt, disconnect_and_free_dbc_and_env_handles,
-        exec_direct_default_query, get_sql_diagnostics,
+        bind_cols, default_setup_connect_and_alloc_stmt, exec_direct_default_query,
+        get_sql_diagnostics,
     };
     use definitions::{
         AttrOdbcVersion, CDataType, FetchOrientation, FreeStmtOption, Handle, HandleType, Integer,
--- a/integration_test/tests/bind_cols_tests.rs
+++ b/integration_test/tests/bind_cols_tests.rs
@@ -7,7 +7,7 @@ mod integration {
     };
     use definitions::{
         AttrOdbcVersion, CDataType, FetchOrientation, FreeStmtOption, Handle, HandleType, Integer,
-        Pointer, SQLFetchScroll, SQLFreeStmt, SQLSetStmtAttrW, SmallInt, SqlReturn,
+        Len, Pointer, SQLFetchScroll, SQLFreeStmt, SQLSetStmtAttrW, SmallInt, SqlReturn,
         StatementAttribute, ULen,
     };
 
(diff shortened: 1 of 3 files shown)


integration_test/tests/free_and_cancel_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+            exec_direct_default_query(stmt_handle);
+
+            fetch_and_bind_cols(
+                stmt_handle,
+                vec![CDataType::SQL_C_SLONG, CDataType::SQL_C_SBIGINT],
+            );
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt_handle, FreeStmtOption::SQL_UNBIND)
```
mattChiaravalloti, line 116: The actual SSIS trace does `SQLFreeStmt(SQL_CLOSE)`, but I decided not to do that since that is not as interesting as `SQL_UNBIND`. I can change this test or add a second test if either of you thinks it's worth also testing `SQL_CLOSE` after calls to `SQLBindCol`, though those functions shouldn't have anything to do with one another.

integration_test/tests/free_and_cancel_tests.rs
```diff
@@ -56,6 +56,82 @@ mod integration {
...
+            exec_direct_default_query(stmt_handle);
+
+            fetch_and_bind_cols(
+                stmt_handle,
+                vec![CDataType::SQL_C_SLONG, CDataType::SQL_C_SBIGINT],
+            );
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt_handle, FreeStmtOption::SQL_UNBIND),
```
mattChiaravalloti, line 126: The actual SSIS trace does `SQLFreeStmt(SQL_CLOSE)`, but I decided not to do that since that is not as interesting as `SQL_UNBIND`. I can change this test or add a second test if either of you thinks it's worth also testing `SQL_CLOSE` after calls to `SQLBindCol`, though those functions shouldn't have anything to do with one another.

jchemburkar: I agree unbind is more interesting / covers more in integration tests since we already have close 👍

integration_test/tests/free_and_cancel_tests.rs
```diff
@@ -56,6 +56,82 @@ mod integration {
         }
     }
 
+    unsafe fn exec_direct_default_query(stmt_handle: HStmt) {
```
jchemburkar, line 59: Looks like we can move this up and use this in `test_free_stmt` too!

jchemburkar (commented):
I agree this test will be more valuable if we block it on the row wise binding! Just had a couple small thoughts on improvements but looks good

integration_test/tests/free_and_cancel_tests.rs
```diff
@@ -56,6 +56,82 @@ mod integration {
...
+    ///     - SQLExecDirectW(<query>)
+    ///     - <loop: until SQLFetchScroll returns SQL_NO_DATA>
+    ///         - SQLBindCol for each column
+    ///         - SQLFetchScroll
+    ///     - SQLFreeStmt(SQL_UNBIND)
+    ///     - SQLDisconnect
+    ///     - SQLFreeHandle(SQL_HANDLE_DBC)
+    ///     - SQLFreeHandle(SQL_HANDLE_ENV)
+    #[test]
+    fn test_free_stmt_after_bind_col() {
```
nbagnard, line 86: In the spirit of having test files associated with the functionality we are testing, I think that the logic around retrieving data from bound columns should move to a different file.
The focus of this function, if I understand correctly, is to check that we unbind bound columns correctly.
I would remove the portion of the code setting the SQL_BIND_BY_COLUMN (this is the default value) and SQL_ATTR_ROW_ARRAY_SIZE attributes, it doesn't matter here and will likely not work as expected when binding rowset is supported because it is asking for 1000 rows and the buffer size is 4, so I'd expect SQL_FETCH to return SQL_ERROR or SQL_SUCCESS_WITH_INFO, but likely not SQL_SUCCESS or SQL_NO_DATA.

We should also verify that we bind buffers and put the data in them correctly, but this shouldn't happen here.

For checking free behavior, this is enough :
   ///    - SQLExecDirectW(<query>)
   ///    - SQLBindCol
    ///   - SQLFetchScroll
    ///    - SQLFreeStmt(SQL_UNBIND)
    
    
  For checking rowset retrieval, you'll need :
   ///    - SQLSetStmtAttrW(SQL_ATTR_ROW_ARRAY_SIZE, xxx)
   ///    - SQLExecDirectW(<query>)
   ///    - SQLBindCol
   ///   - <loop: until SQLFetchScroll returns SQL_NO_DATA> <--- Want to verify that SQLFetchScroll returns data for the expected number of times
   ///             - SQLFetchScroll.  <--- Want to check that the buffer content changed (a.k.a new data was copied)

integration_test/tests/free_and_cancel_tests.rs
```diff
@@ -56,6 +56,82 @@ mod integration {
...
+            );
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt_handle, FreeStmtOption::SQL_UNBIND),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, stmt_handle as Handle)
+            );
+
+            disconnect_and_free_dbc_and_env_handles(env_handle, conn_handle);
```
nbagnard, line 131: Is this important for the test or just part of proper clean up?

integration_test/tests/free_and_cancel_tests.rs
nbagnard: True!
You would usually unbind only is you plan to rebind, not if you plan to close your statement/connection.
It might be worth thinking about testing bind/unbind flow specifically if we see a tool doing that.

EthanHardyMongo:
@jchemburkar and @nbagnard Y'all can hold off an reviewing for now. Since I changed `sql_fetch_helper` a lot in #192, that PR needs to be merged first before this integration test can work properly. Also, there needs to be quite a few changes to fit the new behavior of `sql_fetch_helper`.

nbagnard (changes requested):

integration_test/tests/common/mod.rs
```diff
@@ -425,3 +426,35 @@ pub fn get_column_attributes(stmt: Handle, expected_col_count: SmallInt) {
...
     }
 }
+
+#[allow(dead_code)]
+/// Helper function for binding columns in a result set.
+/// - loop:
+///    - SQLBindCol for each column (determined by target_types vector)
+///    - SQLFetchScroll until SQL_NO_DATA is returned
+pub fn fetch_and_bind_cols(stmt_handle: HStmt, target_types: Vec<CDataType>) {
+    let binding_buffer = &mut [0u16; 4] as *mut _;
```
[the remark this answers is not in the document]
mattChiaravalloti: Agreed. Talked to Ethan about this as well yesterday. When I first wrote this test I did it based on what was in master, not realizing the full scope of Ethan's current work. In master, the binding_buffer isn't actually ever used for anything so it was unclear to me that it was actually consequential in this test. Now that he's explained to me how it is used, I'll update this test entirely.

mattChiaravalloti:
I completely rewrote the integration tests, mainly following Natacha's suggestion [here](https://github.com/mongodb/mongo-odbc-driver/pull/196#discussion_r1542041578), and incorporating all the things I learned from Ethan and from Jonathan's comments. I didn't run these yet, so what I just pushed will be the first time they run. I _think_ they look good but it's easy for me to wrong about ODBC testing, so I'll keep an eye on them and fix anything that's incorrect.

This is ready for another look.

integration_test/tests/common/mod.rs
```diff
@@ -425,3 +425,36 @@ pub fn get_column_attributes(stmt: Handle, expected_col_count: SmallInt) {
...
+    for (i, (target_type, binding_buffer, buffer_length)) in target_types.iter().enumerate() {
+        assert_eq!(
+            SqlReturn::SUCCESS,
+            SQLBindCol(
+                stmt_handle,
+                (i + 1) as USmallInt,
+                *target_type as SmallInt,
+                *binding_buffer,
+                *buffer_length,
+                null_mut(),
```
EthanHardyMongo, line 441: The failures are due to a segmentation fault, and here we can see that the `length_or_indicator` is null, so that's probably the cause of the problem.

EthanHardyMongo (changes requested):
Looks good overall! After my 2 comments are addressed, it should work fine.

integration_test/tests/bind_cols_tests.rs
```diff
@@ -0,0 +1,134 @@
...
+            exec_direct_default_query(stmt_handle);
+
+            // Since ROW_ARRAY_SIZE is set to 2, we double the buffer lengths.
+            let id_buffer = &mut [0u8; 8] as *mut _;
+            let a_buffer = &mut [0u8; 16] as *mut _;
+            bind_cols(
+                stmt_handle,
+                vec![
+                    (CDataType::SQL_C_SLONG, id_buffer as Pointer, 8),
+                    (CDataType::SQL_C_SBIGINT, a_buffer as Pointer, 16),
```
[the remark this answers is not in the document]
mattChiaravalloti: You rock 🎸 Good find, thanks for explaining!

integration_test/tests/common/mod.rs
mattChiaravalloti: Updated!

integration_test/tests/bind_cols_tests.rs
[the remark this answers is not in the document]
mattChiaravalloti: Fixed

mattChiaravalloti:
A couple careless mistakes and typos led to the previous failure. The latest commit _should_ finally go green.

integration_test/tests/bind_cols_tests.rs
```diff
@@ -0,0 +1,160 @@
...
+                assert_eq!(SqlReturn::SUCCESS, result);
+
+                // After fetching the data, check that the buffers contain the
+                // expected values. As shown above, the data only has 3 rows.
+                assert!(i < 2, "too much data fetched");
+
+                let (id1, id2) = expected_id_data[i];
+                assert_eq!(id1, *(id_buffer as *mut i32));
+                assert_eq!(id2, *((id_buffer as ULen + 4) as *mut i32));
+                assert_eq!(4, *id_indicator);
```
nbagnard, line 146: You can reuse the transfer octet size here and for the a_buffer offset too if you decide to use it.
Also, this is an array too, so you should use offsets.

nbagnard (changes requested):
I have a couple of suggestions to make it more bullet proof, but otherwise it looks good :) 👍 
The only little thing that I think is off is how you check the indicator/str_len values.
It is an array and reflects the size for each cell in the rowset I think based on this: https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/column-wise-binding?view=sql-server-ver16

integration_test/tests/bind_cols_tests.rs
mattChiaravalloti: > Also, this is an array too, so you should use offsets.

To clarify this, do you mean that since `id_indicator` is an array of length 2, I should ensure that each element, `id_indicator[0]` and `id_indicator[1]` is `ID_TRANSFER_OCTET_LENGTH`? (And do the same for `a`)

EthanHardyMongo: I think what she means is that we should check both values in the indicator array instead of just the first one. Also, the indicator, will always be `isize`, so the offset is 8.

EthanHardyMongo:
```
assert_eq!(4, *id_indicator);
assert_eq!(4, *((id_indicator as ULen + 8) as *mut Len));
```

nbagnard: Yes, because it's the indicator for each I believe or at least this what I see here : https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/column-wise-binding?view=sql-server-ver16

mattChiaravalloti: > ```
> assert_eq!(4, *id_indicator);
> assert_eq!(4, *((id_indicator as ULen + 8) as *mut Len));


Given the change Natacha suggested above, `id_indicator` and `a_indicator` are statically `&mut [isize; 2]` (they are cast to `*mut Len` when argued to `SQLBindCol`). So instead of doing this manual offset, can't I just use array indexing, like this:
```
assert_eq!(ID_TRANSFER_OCTET_LEN, id_indicator[0] as usize);
assert_eq!(ID_TRANSFER_OCTET_LEN, id_indicator[1] as usize);
```
and equivalent for `a_indicator`?

mattChiaravalloti:
I'm still investigating why these tests timed out.

integration_test/tests/bind_cols_tests.rs
EthanHardyMongo: yeah that looks good!

EthanHardyMongo:
> I'm still investigating why these tests timed out.

Yeah I was looking at that too. I'm not sure what's going on

mattChiaravalloti:
The problem seems to be the "cleanup" lines that freed the env and conn handles. Without them, this no longer hangs.

nbagnard (approved):
LGTM!

EthanHardyMongo (approved):
LGTM

jchemburkar (approved):

PR #199, opened 2024-04-02
SQL-1936: Implement do_refresh

pmeredit:
This adds the refresh flow, which trades a refresh token for a new access token (and potentially a new refresh token, but the server we are testing against is setup to return the same refresh token).


core/src/oidc_auth.rs
```diff
@@ -40,9 +40,37 @@ pub enum Error {
     IssuerUriMustBeHttps,
     NoIdpServerInfo,
     CsrfMismatch,
+    MissingRefreshToken,
     Other(String),
 }
```
pmeredit, line 46: I couldn't figure out the necessary types to make this a function, so much easier to have a macro that expands to a closure ;)

bucaojit (approved):
LGTM

core/src/bin/do_auth_flow_and_refresh.rs
```diff
@@ -0,0 +1,25 @@
+use mongo_odbc_core::oidc_auth::*;
+
+#[tokio::main]
+async fn main() {
```
nbagnard, line 4: This is only for testing, right?
What about we rename the files in bin something like "test_auth_flow_and_refresh.rs", "test_auth_flow.rs" and "test_rfc8252_http_server.rs". This makes it clear it is just for testing and nothing else.


Also, I think that these files are built and bundled in the driver when released. Should we use a feature flag to only build it when needed for testing?

core/src/oidc_auth.rs
```diff
@@ -40,9 +40,37 @@ pub enum Error {
...
+                }
+                RequestTokenError::Other(msg) => msg,
+            };
+            Error::Other(format!("OpenID Connect: code exchange failed: {}", msg))
+        }
+    };
+}
+
 pub async fn do_auth_flow(params: CallbackContext) -> Result<IdpServerResponse, Error> {
     let idp_info = params.idp_info.ok_or(Error::NoIdpServerInfo)?;
```
nbagnard, line 75: Creating a client is the same for both do_auth and do_refresh.
Can you move it out in a function which will return a client and the request_scopes.

core/src/oidc_auth.rs
```diff
@@ -175,3 +182,49 @@ pub async fn do_auth_flow(params: CallbackContext) -> Result<IdpServerResponse,
...
+
+    // Create an OpenID Connect client by specifying the client ID,
+    // authorization URL and token URL.
+    let client = CoreClient::from_provider_metadata(
+        provider_metadata,
+        ClientId::new(client_id),
+        None,
+    );
+
+    // This must never be called without a refresh token, but we return this error anyway
```
nbagnard, line 207: Not sure I understand the comment completely. 
I agree it is good to protect our-self here and I believe this is what you imply, correct?

A short comment on top of the function, describing what it is used for (I know, it's mostly self explanatory, but it won't hurt) and stating what fields must be present is also a good idea. 
Maybe for `do_auth_flow` too, since idp_info is an option but it is actually required.

nbagnard (changes requested):

core/src/oidc_auth.rs
pmeredit: Ah, yes, the comment would make a lot more sense if I'd already implemented the next ticket. This function will only ever be called when there is a refresh_token, but I decided to be defensive anyway

nbagnard (approved):
LGTM! 💯

core/src/oidc_auth.rs
pmeredit: the refresh flow does not care about scopes

PR #201, opened 2024-04-09
SQL-2011: switch to async

terakilobyte:
This PR contains the necessary functionality to switch us to async. I've commented the PR down below for what I think are the "interesting" bits.


core/Cargo.toml
```diff
@@ -26,16 +26,17 @@ cstr = { path = "../cstr" }
 fancy-regex = "0.11.0"
 shared_sql_utils = { path = "../shared_sql_utils" }
 log = "0.4.14"
+mongodb = "2.8.2"
+# Do NOT change these features without consulting with other team members. 
+# The features are used to control the behavior of tokio. Tokio is unsafe to use
+# across ABI boundaries in any other runtime but current_thread
+tokio = { version = "1", features = ["rt", "sync", "io-util", "macros", "net"] }
+futures = "0.3"
```
terakilobyte, line 34: A stern warning that I hope captures the attention of future teammates.

core/src/conn.rs
```diff
@@ -2,11 +2,12 @@ use crate::odbc_uri::UserOptions;
 use crate::{err::Result, Error};
 use crate::{MongoQuery, TypeMode};
 use bson::{doc, Bson, UuidRepresentation};
-use mongodb::sync::Client;
+use mongodb::Client;
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 
 #[derive(Debug)]
+#[repr(C)]
```
terakilobyte, line 10: Just for safety, I wanted to ensure the memory layout is preserved across ABI boundary.

core/src/conn.rs
```diff
@@ -36,60 +40,88 @@ impl MongoConnection {
...
         login_timeout: Option<u32>,
         type_mode: TypeMode,
+        mut runtime: Option<tokio::runtime::Runtime>,
     ) -> Result<Self> {
+        let runtime = runtime.take().unwrap_or_else(|| {
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .unwrap()
+        });
```
terakilobyte, line 50: It's arguable we don't want to make this argument optional, but I like this as it makes using it directly int tests easier in the future.

core/src/odbc_uri.rs
```diff
@@ -278,12 +280,22 @@ impl ODBCUri {
...
+        // instead of the system resolver
+        // https://github.com/mongodb/mongo-rust-driver?tab=readme-ov-file#windows-dns-note
+        let parse_func = || async {
+            if cfg!(target_os = "windows") {
+                ClientOptions::parse_with_resolver_config(uri, ResolverConfig::cloudflare()).await
+            } else {
+                ClientOptions::parse(uri).await
+            }
+        };
+        let mut client_options = parse_func().await.map_err(Error::InvalidClientOptions)?;
```
terakilobyte, line 298: The easiest thing to do was convert this to an async function.

odbc/tests/connection_tests.rs
```diff
@@ -1,5 +1,17 @@
...
+static TEST_SYNC: Mutex<Executor> = Mutex::new(Executor);
+#[derive(Clone, Copy)]
+struct Executor;
+
+impl Executor {
+    fn run_test(self, f: impl FnOnce()) {
+        f()
+    }
+}
+
```
terakilobyte, line 14: Olympics! This essentially forces `cargo test -- --test-threads=1`. We do this in the evergreen file, but this is nice when you just want to run `cargo test` from the command line.

odbc/tests/connection_tests.rs
```diff
@@ -30,47 +42,50 @@ macro_rules! test_connection_diagnostics {
...
-            };
+                    verify_sql_diagnostics(
+                        HandleType::SQL_HANDLE_DBC,
+                        conn_handl as *mut _,
+                        1,
+                        expected_sql_state.odbc_3_state,
+                        expected_error_message,
+                        0,
+                    );
+                    let _ = SQLDisconnect(conn_handl as *mut _);
```
terakilobyte, line 77: We weren't calling SQLDisconnect. Since disconnect now shuts down the tokio runtime, it's important we call it for good housekeeping.

core/src/odbc_uri.rs
terakilobyte: Also, on windows, we'll be using the cloudflare resolver...

core/src/conn.rs
```diff
@@ -2,11 +2,12 @@ use crate::odbc_uri::UserOptions;
...
-use mongodb::sync::Client;
+use mongodb::Client;
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 
 #[derive(Debug)]
+#[repr(C)]
 pub struct MongoConnection {
     /// The mongo DB client
     pub client: Client,
```
pmeredit, line 13: @nbagnard this will become `Arc<Client>` for OIDC, and somehow I will map ODBCUri to `Arc<Client>` so that we can retrieve the same client for anything using the same attributes.

core/src/conn.rs
pmeredit: I'm fond of this

rfc8252_http_server/Cargo.toml
```diff
@@ -1,15 +1,14 @@
 [package]
 name = "rfc8252_http_server"
 version = "0.0.0"
-authors = [
```
pmeredit, line 4: Add your name, you touched it :D

pmeredit (approved):
LGTM!

core/src/conn.rs
pmeredit: Actually! `Client` already uses an `Arc`, so all we have to do is clone it, no change to anything. It's actually an intended part of the API design to share `Client`s between async tasks and threads.

nbagnard (changes requested):

core/src/databases.rs
```diff
@@ -200,24 +200,28 @@ impl MongoDatabases {
...
+                        .authorized_databases(true)
+                        .build(),
+                )
+                .await
+                .unwrap()
+                // MHOUSE-7119 - admin database and empty strings are showing in list_database_names
+                .iter()
+                .filter(|&db_name| !db_name.is_empty() && !db_name.eq("admin"))
+                .map(|s| s.to_string())
+                .collect();
```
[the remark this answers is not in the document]
terakilobyte: In `list_tables` we're returning a `Vec<CollectionsForDb>`, which needs an async operation to calculate each one. Here's we're just getting the database names, which is a single operation.

logger/src/lib.rs
```diff
@@ -18,6 +18,12 @@ use std::path::{Path, PathBuf};
 
 const LOG_FILE_SIZE: u64 = 1024 * 500;
 
+#[cfg(any(test, feature = "test"))]
```
nbagnard, line 21:
```suggestion
// Log file name used by our tests. It is different that the regular log files name so that we don't
// inadvertently delete log files we want to keep when running tests.
#[cfg(any(test, feature = "test"))]
```

logger/src/lib.rs
```diff
@@ -18,6 +18,12 @@ use std::path::{Path, PathBuf};
...
+pub const LOG_FILE_NAME: &str = "mongo_odbc_test.log";
+
+#[cfg(not(any(test, feature = "test")))]
```
nbagnard, line 24:
```suggestion
// The regular log file name.
#[cfg(not(any(test, feature = "test")))]
```

nbagnard (approved):
LGTM! 🎉 🎉 🎉

integration_test/tests/connection_tests.rs
```diff
@@ -108,6 +108,7 @@ mod integration {
     // At connection time, when the connection string is processed, if there is a log level specified
     // the logger log level is updated to the connection log level.
     #[test]
+    #[ignore = "TODO: This test is flaky and should be fixed in the future."]
```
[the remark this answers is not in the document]
terakilobyte: Turns out this test needs to be run as an administrator (on windows) to work since it modifies the registry!

PR #202, opened 2024-04-09
SQL-2009: simple  types for null

terakilobyte:
This way a truly NULL column isn't interpreted as an unknown type.

### DIFF
--- a/core/src/bson_type_info.rs
+++ b/core/src/bson_type_info.rs
@@ -303,7 +303,7 @@ impl BsonTypeInfo {
         is_auto_unique_value: None,
         is_unsigned: None,
         num_prec_radix: None,
-        simple_type_info: None,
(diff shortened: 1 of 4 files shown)


core/src/bson_type_info.rs
```diff
@@ -303,7 +303,7 @@ impl BsonTypeInfo {
         is_auto_unique_value: None,
         is_unsigned: None,
         num_prec_radix: None,
-        simple_type_info: None,
+        simple_type_info: SimpleTypeInfo::new(4, 4, 4),
```
mattChiaravalloti, line 306: Do we need to add any testing for this anywhere?

mattChiaravalloti (approved):
Looks good!

EthanHardyMongo (approved):
LGTM! Other than the one failing integration test that needs to be updated, everything looks good.

PR #203, opened 2024-04-10
SQL-2000: decimal128 tests

terakilobyte:
Just a small one adding tests to ensure we're fetching decimal128 values correctly!


EthanHardyMongo (approved):
LGTM!

PR #205, opened 2024-04-15
SQL-2038: Add support for  ODBC 2.0 descriptors SQL_COLUMN_LEN, SQL_SCALE and SQL_PRECISION

nbagnard:


terakilobyte (commented):
Will review again once the tests are good, but just some initial comments. Mostly looks good!

odbc/src/api/col_attr_describe_tests.rs
```diff
@@ -544,13 +544,13 @@ mod unit {
             let mut stmt = Statement::with_state(conn as *mut _, StatementState::Allocated);
             stmt.mongo_statement = RwLock::new(Some(Box::new(MongoFields::empty())));
             let mongo_handle: *mut _ = &mut MongoHandle::Statement(stmt);
-
+            let field_identifier = 1;
```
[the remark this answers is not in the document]
terakilobyte:
```suggestion
            let invalid_field_identifier = 1;
```

terakilobyte (approved):

terakilobyte (approved):
LGTM. I tested with both SSIS and Power BI locally and everything works as expected!

core/src/fields.rs
```diff
@@ -705,46 +705,85 @@ impl MongoStatement for MongoFields {
...
+                // give the maximum number of digits and maximum number of digits to the right of the
+                // decimal point allowed for the column respectively.
+                SqlDataType::SQL_VARCHAR
+                | SqlDataType::SQL_WVARCHAR
+                | SqlDataType::SQL_VARBINARY => match get_meta_data()?.char_octet_length {
+                    None => Bson::Int32(definitions::SQL_NO_TOTAL as i32),
+                    Some(char_octet_length) => Bson::Int32(char_octet_length as i32),
+                },
+                _ => Bson::Null,
+            },
```
pmeredit, line 783:
```suggestion
            16 => {
                let meta_data = get_meta_data()?;
                // For numeric data types. 10 means that the values in COLUMN_SIZE and DECIMAL_DIGITS
                // give the maximum number of digits and maximum number of digits to the right of the
                // decimal point allowed for the column respectively.
                if matches!(
                    meta_data.sql_type,
                    SqlDataType::SQL_VARCHAR
                        | SqlDataType::SQL_WVARCHAR
                        | SqlDataType::SQL_VARBINARY
                ) {
                    Bson::Int32(
                        meta_data
                            .char_octet_length
                            .unwrap_or(definitions::SQL_NO_TOTAL) as i32,
                    )
                } else {
                    Bson::Null
                }
            }
```

pmeredit: It might actually make sense to add an `is_char_type()` to `SqlDataType` that has that `matches!` in it, as well.

nbagnard: Yes, I was thinking of doing that too.
There is only that one place where I needed this kind of rule, so I didn't make that change.
The other places, I baked the rule in the value itself.
But I think it might be valuable to have helpers for is_numeric, is_character, is_binary, is_datetime, etc...
I'll create a quickwin JIRA for this.

core/src/fields.rs
```diff
@@ -705,46 +705,85 @@ impl MongoStatement for MongoFields {
...
-                match l {
-                    None => definitions::SQL_NO_TOTAL as i32,
-                    Some(_) => 0i32,
-                }
-            }),
+            // CHAR_OCTET_LENGTH
+            // The maximum length in bytes of a character or binary data type column.
+            // For all other data types, this column returns a NULL.
+            16 => match get_meta_data()?.sql_type {
+                // For numeric data types. 10 means that the values in COLUMN_SIZE and DECIMAL_DIGITS
```
nbagnard, line 773: This comment is junk and a copy paste of the one if NUM_PREC_RADIX.

PR #206, opened 2024-04-17
SQL-1933: Add oidc_call_back function

pmeredit:
Note that this also handles the timeout. The test function with timeout works on my laptop. I'm still waiting forever and a half for my desktop harddrive to finish encrypting, but I imagine it will work on windows, also.

### DIFF
--- a/core/src/bin/test_oidc_call_back.rs
+++ b/core/src/bin/test_oidc_call_back.rs
@@ -1,0 +1,25 @@
+use mongo_odbc_core::oidc_auth::*;
+
+#[tokio::main]
+async fn main() {
+    let c = CallbackContext {
+        idp_info: Some(IdpServerInfo {
+            issuer: "https://mongodb-dev.okta.com/oauth2/ausqrxbcr53xakaRR357".to_string(),
+            client_id: Some("0oarvap2r7PmNIBsS357".to_string()),
+            // show that we get a refresh_token even when we don't ask for it here
+            request_scopes: Some(vec!["openid".to_string()]),
+        }),
+        refresh_token: None,
+        timeout_seconds: None,
+        version: 1,
+    };
+    let mut refresh_c = c.clone();
+    let IdpServerResponse {
+        access_token,
+        expires: _,
+        refresh_token,
+    } = oidc_call_back(c).await.unwrap();
+    println!("initial access token: {:access_token?}");
+    refresh_c.refresh_token = refresh_token;
+    println!("second response: {:?}", oidc_call_back(refresh_c).await.unwrap());
+}
--- a/core/src/bin/test_oidc_call_back.rs
+++ b/core/src/bin/test_oidc_call_back.rs
@@ -1,7 +1,7 @@
 use mongo_odbc_core::oidc_auth::*;
 
 // This shows that the oidc_call_back function works. The second call will use the refresh token,
-// which will be obvious in that the webbrower only opens once.
+// which will be obvious in that the web browser only opens once.
 #[tokio::main(flavor = "current_thread")]
 async fn main() {
     let c = CallbackContext {
--- a/core/src/bin/test_oidc_call_back.rs
+++ b/core/src/bin/test_oidc_call_back.rs
@@ -1,6 +1,6 @@
 use mongo_odbc_core::oidc_auth::*;
 
-#[tokio::main]
+#[tokio::main(flavor = "current_thread")]
 async fn main() {
     let c = CallbackContext {
         idp_info: Some(IdpServerInfo {
--- a/core/src/bin/test_oidc_call_back.rs
+++ b/core/src/bin/test_oidc_call_back.rs
@@ -15,11 +15,14 @@ async fn main() {
     };
     let mut refresh_c = c.clone();
     let IdpServerResponse {
-        access_token,
+        access_token: _,
         expires: _,
         refresh_token,
     } = oidc_call_back(c).await.unwrap();
-    println!("initial access token: {:access_token?}");
+    println!("initial refresh token: {refresh_token:?}");
     refresh_c.refresh_token = refresh_token;
-    println!("second response: {:?}", oidc_call_back(refresh_c).await.unwrap());
+    println!(
+        "second response: {:?}",
+        oidc_call_back(refresh_c).await.unwrap()
+    );
 }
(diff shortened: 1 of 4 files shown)


core/src/bin/test_oidc_call_back.rs
```diff
@@ -0,0 +1,28 @@
...
+        }),
+        refresh_token: None,
+        timeout_seconds: None,
+        version: 1,
+    };
+    let mut refresh_c = c.clone();
+    let IdpServerResponse {
+        access_token: _,
+        expires: _,
+        refresh_token,
```
pmeredit, line 20: this is similar to the previous refresh test, but it uses oidc_call_back in both places.

core/src/oidc_auth.rs
```diff
@@ -7,8 +6,10 @@ use openidconnect::{
 };
 use rfc8252_http_server::{start, OidcResponseParams};
 use std::{collections::HashSet, hash::RandomState, time::Instant};
+use tokio::time::{self, Duration};
 
 const DEFAULT_REDIRECT_URI: &str = "http://localhost:27097/redirect";
+const DEFAULT_SLEEP_DURATION: Duration = Duration::from_secs(5 * 60); // from_mins is unstable
```
pmeredit, line 12:
```suggestion
const DEFAULT_SLEEP_DURATION: Duration = Duration::from_secs(5 * 60); // from_mins is unstable, so we use from_secs with a multiplication. The multiplication is performed at compile time, anyway
```

pmeredit:
timeout test works on windows, also, though it's a bit wonky if you are not already logged in to Okta (because the local server gets torn down).

core/src/bin/test_oidc_call_back.rs
```diff
@@ -0,0 +1,28 @@
+use mongo_odbc_core::oidc_auth::*;
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
```
nbagnard, line 4: What are you testing here?
My understanding is that you are checking that:
- You can authenticate, both when there is no refresh token and when there is one.
- After the first oidc call, you get back a refresh token and it can be used in subsequent call

If this correct, please, add a short comment on top of your function.

Also, what about using an expired refresh token. Could we add a case for it?

core/src/oidc_auth.rs
```diff
@@ -41,9 +42,38 @@ pub enum Error {
...
     Other(String),
 }
 
+impl From<Error> for mongodb::error::Error {
+    fn from(e: Error) -> Self {
+        mongodb::error::Error::custom(e)
+    }
+}
+
+pub async fn oidc_call_back(params: CallbackContext) -> mongodb::error::Result<IdpServerResponse> {
```
nbagnard, line 55: Please, here too add a comment explaining that depending on the info we need to re-auth "from scratch" or we can use the refresh token when provided.

core/src/oidc_auth.rs
```diff
@@ -41,9 +42,38 @@ pub enum Error {
...
+    let sleep_duration = params
+        .timeout_seconds
+        // turn the supplied timeout Instant into a Duration from now
+        .map(|x| x - Instant::now())
+        .unwrap_or(DEFAULT_SLEEP_DURATION);
+    let timeout = time::sleep(sleep_duration);
+    tokio::pin!(timeout);
+
+    if params.refresh_token.is_some() {
```
nbagnard, line 64: I get the idea behind this now that I looked at the doc, but couldn't we have a simpler and more straightforward way of having a timeout on a job?

It looks like the [`time` feature from Tokio](https://docs.rs/tokio/latest/tokio/time/index.html) would work well here.
For example 
```
let res = timeout(Duration::from_secs(1), do_refresh(params)).await;

if res.is_err() {
    println!("operation timed out");
}
```

nbagnard (changes requested):

core/src/bin/test_oidc_call_back.rs
pmeredit: I added a test with a bad refresh token, which works as expected. The idp is actually not configured for refresh tokens to ever expire, only access tokens expire.

core/src/oidc_auth.rs
pmeredit: I wrote it this way because this is how you'd write it in go :joy:. Had no idea this timeout function worked this way.

core/src/bin/test_oidc_call_back.rs
nbagnard: Ah, yes, I couldn't remember if there was an expiration or not. I know it is possible, but if ours don't have one, then all good :)

core/src/bin/test_oidc_call_back_short_timeout.rs
```diff
@@ -0,0 +1,19 @@
...
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    let c = CallbackContext {
+        idp_info: Some(IdpServerInfo {
+            issuer: "https://mongodb-dev.okta.com/oauth2/ausqrxbcr53xakaRR357".to_string(),
+            client_id: Some("0oarvap2r7PmNIBsS357".to_string()),
+            request_scopes: Some(vec!["openid".to_string()]),
+        }),
+        refresh_token: None,
+        // 2 seconds works well on my laptop. 1 second results in the request dying before the
```
bucaojit, line 12: The 2 seconds works on my laptop as well for these two tests.

core/src/bin/test_oidc_call_back_with_bad_refresh_token.rs
```diff
@@ -0,0 +1,21 @@
...
+            request_scopes: Some(vec!["openid".to_string()]),
+        }),
+        refresh_token: Some("PnWePzgf-4BqC048C5Q7BhtBxfu3nId9e72y5T5-PLj".to_string()),
+        // 2 seconds works well on my laptop. 1 second results in the request dying before the
+        //   server can redirect
+        timeout_seconds: Some(std::time::Instant::now() + std::time::Duration::from_secs(2)),
+        version: 1,
+    };
+    println!(
+        "This will result in an error: {:?}",
```
bucaojit, line 18: Let's add a comment or in this output on what the expected error is since it is possible (but not likely) to get a Timedout error.
```
 Some(Other("OpenID Connect: code exchange failed: Server returned error response: StandardErrorResponse { error: invalid_grant, error_description: Some(\"The refresh token is invalid or expired.\"), error_uri: None }"))
 ```

bucaojit (approved):
LGTM.
Had some minor comments.

PR #207, opened 2024-04-18
SQL-2032: set default type mode to simple_types_only

jchemburkar:


core/src/odbc_uri.rs
```diff
@@ -610,11 +610,11 @@ mod unit {
             use crate::map;
```
nbagnard, line 610: You probably want to rename that test to `standard_types` or `complex_types` or something like that.

nbagnard (approved):
I just have one little comment on a test name.
All looks good otherwise :)

PR #208, opened 2024-04-22
SQL-2051: clean up statements in disconnect if necessary

terakilobyte:
SQLDisconnect now closes cursors on any open statements as part of its internal logic. This is safe to do since the cursor is an option, and closing it sets the option to None, which will allow Rust to clean up its resources automatically. Iterating over the statements is as safe as possible with no unwraps. The nesting gets a bit deep but it's not too unwieldy.

Additionally, as part of this PR I made a few changes to the common test infra to make it easier to get the default connection string.

### DIFF
--- a/integration_test/tests/common/mod.rs
+++ b/integration_test/tests/common/mod.rs
@@ -179,13 +179,16 @@ pub fn connect_and_allocate_statement(
     in_connection_string: Option<String>,
 ) -> (HDbc, HStmt) {
     let conn_str = in_connection_string.unwrap_or_else(generate_default_connection_str);
-    let conn_handle = connect_with_conn_string(env_handle, conn_str).unwrap();
+    let conn_handle = connect_with_conn_string(env_handle, Some(conn_str)).unwrap();
     (conn_handle, allocate_statement(conn_handle).unwrap())
 }
 
 #[allow(dead_code)]
-/// Connects to database with provided connection string
-pub fn connect_with_conn_string(env_handle: HEnv, in_connection_string: String) -> Result<HDbc> {
+/// Connects to database with provided connection string, or default connection string
+pub fn connect_with_conn_string(
+    env_handle: HEnv,
+    in_connection_string: Option<String>,
+) -> Result<HDbc> {
     // Allocate a DBC handle
     let mut dbc: Handle = null_mut();
     unsafe {
@@ -197,6 +200,8 @@ pub fn connect_with_conn_string(env_handle: HEnv, in_connection_string: String) 
             SqlReturn::SUCCESS => (),
             sql_return => return Err(Error::HandleAllocation(sql_return_to_string(sql_return))),
         }
+        let in_connection_string =
+            in_connection_string.unwrap_or_else(generate_default_connection_str);
         let mut in_connection_string_encoded = cstr::to_widechar_vec(&in_connection_string);
         in_connection_string_encoded.push(0);
         let str_len_ptr = &mut 0;
--- a/integration_test/tests/disconnect_tests.rs
+++ b/integration_test/tests/disconnect_tests.rs
@@ -1,0 +1,130 @@
+mod common;
+
+mod integration {
+    use crate::common::{
+        allocate_env, allocate_statement, connect_with_conn_string,
+        default_setup_connect_and_alloc_stmt, disconnect_and_free_dbc_and_env_handles,
+        get_sql_diagnostics,
+    };
+    use definitions::{
+        AttrOdbcVersion, FreeStmtOption, Handle, HandleType, SQLExecDirectW, SQLFreeHandle,
+        SQLFreeStmt, SqlReturn, SQL_NTS,
+    };
+
+    #[test]
+    fn sql_disconnect_closes_statement_implicitly() {
+        let (env, dbc, stmt) = default_setup_connect_and_alloc_stmt(AttrOdbcVersion::SQL_OV_ODBC3);
+        let query = b"SELECT * FROM integration_test.foo\0".map(|b| b as u16);
+        unsafe {
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLExecDirectW(stmt, query.as_ptr(), SQL_NTS as i32,),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt.cast::<Handle>()),
+            );
+
+            // we will not free the statement, instead proceeding directly to disconnect and freeing the dbc and env handles
+            disconnect_and_free_dbc_and_env_handles(env, dbc)
+        }
+    }
+
+    #[test]
+    fn sql_disconnect_implicitly_closing_statment_cursors_is_safe_for_closed_statements() {
+        let (env, dbc, stmt) = default_setup_connect_and_alloc_stmt(AttrOdbcVersion::SQL_OV_ODBC3);
+        let query = b"SELECT * FROM integration_test.foo\0".map(|b| b as u16);
+        unsafe {
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLExecDirectW(stmt, query.as_ptr(), SQL_NTS as i32,),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt.cast::<Handle>()),
+            );
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt, FreeStmtOption::SQL_CLOSE as i16),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt.cast::<Handle>()),
+            );
+
+            // we will not free the statement, instead proceeding directly to disconnect and freeing the dbc and env handles
+            disconnect_and_free_dbc_and_env_handles(env, dbc)
+        }
+    }
+
+    #[test]
+    fn sql_disconnect_implicitly_closing_statment_cursors_is_safe_for_unbound_statemens() {
+        let (env, dbc, stmt) = default_setup_connect_and_alloc_stmt(AttrOdbcVersion::SQL_OV_ODBC3);
+        let query = b"SELECT * FROM integration_test.foo\0".map(|b| b as u16);
+        unsafe {
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLExecDirectW(stmt, query.as_ptr(), SQL_NTS as i32,),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt.cast::<Handle>()),
+            );
+
+            // we didn't actually bind a statement, so this should be a no-op
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt, FreeStmtOption::SQL_UNBIND as i16),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt.cast::<Handle>()),
+            );
+
+            // we will not free the statement, instead proceeding directly to disconnect and freeing the dbc and env handles
+            disconnect_and_free_dbc_and_env_handles(env, dbc)
+        }
+    }
+
+    #[test]
+    fn sql_disconnect_handles_many_statements_properly() {
+        let env = allocate_env(AttrOdbcVersion::SQL_OV_ODBC3);
+        let dbc = connect_with_conn_string(env, None)
+            .expect("Failed to connect with default connection string");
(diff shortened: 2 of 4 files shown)


integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,130 @@
...
+            );
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt, FreeStmtOption::SQL_CLOSE as i16),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt.cast::<Handle>()),
+            );
+
+            // we will not free the statement, instead proceeding directly to disconnect and freeing the dbc and env handles
```
mattChiaravalloti, line 50:
```suggestion
```

integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,130 @@
...
+
+            // we didn't actually bind a statement, so this should be a no-op
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt, FreeStmtOption::SQL_UNBIND as i16),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt.cast::<Handle>()),
+            );
+
+            // we will not free the statement, instead proceeding directly to disconnect and freeing the dbc and env handles
```
mattChiaravalloti, line 75:
```suggestion
            // Directly disconnect and free the dbc and env handles without closing the statement
```
We should either get rid of this comment or change the word "free" to "close" since we do call `SQLFreeStmt` but use it to unbind rather than close.

integration_test/tests/disconnect_tests.rs
mattChiaravalloti: This comment is inaccurate for this test.

integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,130 @@
...
+        let query = b"SELECT * FROM integration_test.foo\0".map(|b| b as u16);
+        unsafe {
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLExecDirectW(stmt, query.as_ptr(), SQL_NTS as i32,),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt.cast::<Handle>()),
+            );
+
+            // we will not free the statement, instead proceeding directly to disconnect and freeing the dbc and env handles
```
mattChiaravalloti, line 26:
```suggestion
            // Directly disconnect and free the dbc and env handles without freeing the statement
```
Small nit about using present tense instead of future tense

integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,130 @@
...
+        statements.iter().for_each(|stmt| unsafe {
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLExecDirectW(*stmt, query.as_ptr(), SQL_NTS as i32,),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt.cast::<Handle>()),
+            );
+        });
+
+        // we will close statements 1 and 3
```
mattChiaravalloti, line 102:
```suggestion
        // Close statements 1 and 3
```

integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,130 @@
...
+            );
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt_3, FreeStmtOption::SQL_CLOSE as i16),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, *stmt_3.cast::<Handle>()),
+            );
+
+            // and we'll free statement 3 handle
```
mattChiaravalloti, line 118:
```suggestion
            // Free statement 3 handle
```

integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,130 @@
...
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                // interestingly, rust doesn't like the cast here, so we'll just use the raw value
```
mattChiaravalloti, line 121:
```suggestion
```
It's unclear what this comment is referring to -- do you mean it doesn't like some cast that isn't included in the code? If so, what would that cast be and why is it important? If it is rejected by the compiler, doesn't that mean it's likely invalid? I think we may want to remove this comment before merging.

integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,130 @@
...
+    fn sql_disconnect_handles_many_statements_properly() {
+        let env = allocate_env(AttrOdbcVersion::SQL_OV_ODBC3);
+        let dbc = connect_with_conn_string(env, None)
+            .expect("Failed to connect with default connection string");
+        let stmt_1 = allocate_statement(dbc).expect("Failed to allocate statement 1");
+        let stmt_2 = allocate_statement(dbc).expect("Failed to allocate statement 2");
+        let stmt_3 = allocate_statement(dbc).expect("Failed to allocate statement 3");
+        let stmt_4 = allocate_statement(dbc).expect("Failed to allocate statement 4");
+        let stmt_5 = allocate_statement(dbc).expect("Failed to allocate statement 5");
+        let statements = [stmt_1, stmt_2, stmt_3, stmt_4, stmt_5];
```
mattChiaravalloti, line 90:
```suggestion
        let stmt_4 = allocate_statement(dbc).expect("Failed to allocate statement 4");
        let statements = [stmt_1, stmt_2, stmt_3, stmt_4];
```
It seems `stmt_5` are extraneous. With just 4 we achieve everything you're trying to test:
- multiple statements are opened and have queries executed on them
- one is closed via SQLFreeStmt
- one is closed via SQLFreeStmt and then free via SQLFreeHandle
- multiple are closed implicitly via SQLDisconnect

Is having a 5th statement absolutely necessary to test some behavior I'm missing?

mattChiaravalloti (changes requested):
The change in `SQLDisconnect` looks good. My comments are only on the test files and mostly about the comments.

integration_test/tests/disconnect_tests.rs
terakilobyte: It's referring to the use of `as *mut _` instead of `.cast::<Handle>`

integration_test/tests/disconnect_tests.rs
terakilobyte: It's just a statement that doesn't have anything done with it.

integration_test/tests/disconnect_tests.rs
terakilobyte: Actually I'm glad you flagged this. In further testing, it's actually unsafe to cast on the pointer. `as *mut _` apparently takes the address of the pointer, which is what is needed, otherwise there's an access violation. `*ptr::addr_of(x)` does the trick nicely and is "safer" than `as`.

mattChiaravalloti (approved):

integration_test/tests/common/mod.rs
```diff
@@ -179,13 +179,16 @@ pub fn connect_and_allocate_statement(
     in_connection_string: Option<String>,
 ) -> (HDbc, HStmt) {
     let conn_str = in_connection_string.unwrap_or_else(generate_default_connection_str);
-    let conn_handle = connect_with_conn_string(env_handle, conn_str).unwrap();
+    let conn_handle = connect_with_conn_string(env_handle, Some(conn_str)).unwrap();
```
nbagnard, line 182: You don't need this logic here anymore if you want to make it an option.
This is a spot where you can use `None`.

odbc/src/api/functions.rs
```diff
@@ -1020,6 +1020,22 @@ pub unsafe extern "C" fn SQLDisconnect(connection_handle: HDbc) -> SqlReturn {
         || {
             let conn_handle = MongoHandleRef::from(connection_handle);
             let conn = must_be_valid!((*conn_handle).as_connection());
+
+            // close open cursors on any statements that haven't been closed
+            if let Ok(stmts) = conn.statements.write() {
```
nbagnard, line 1025: We need to drop the statements too based on the [transition table for SQLDisconnect]( https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/statement-transitions?view=sql-server-ver16#sqldisconnect).
It more or less does a SQLFreeHandle(stmt) for each stmt behind the scene.
And I think we should close the cursor for the stmt in SQLFreeHandle as well, no need to keep it open.


Note
SQLFreeStmt doesn't drop the statements only closes them. They can then be reused with another call to SQLExecute for example.

integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,172 @@
...
+                SqlReturn::SUCCESS,
+                SQLExecDirectW(stmt, query.as_ptr(), SQL_NTS as i32,),
+                "{}",
+                get_sql_diagnostics(
+                    HandleType::SQL_HANDLE_STMT,
+                    *ptr::addr_of!(stmt).cast::<Handle>()
+                ),
+            );
+
+            // stmt isn't freed here, but it should be closed when the DBC is disconnected
```
nbagnard, line 31: SQLDisconnect frees any statements that are allocated on the connection.
You'll want to update the comment here.
This is tied to the comment I added below on SQLDisconnect.

integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,172 @@
...
+        default_setup_connect_and_alloc_stmt, disconnect_and_free_dbc_and_env_handles,
+        get_sql_diagnostics,
+    };
+    use definitions::{
+        AttrOdbcVersion, FreeStmtOption, Handle, HandleType, SQLExecDirectW, SQLFreeHandle,
+        SQLFreeStmt, SqlReturn, SQL_NTS,
+    };
+
+    #[test]
+    fn sql_disconnect_closes_statement_implicitly() {
```
nbagnard, line 17:
```suggestion
    fn sql_disconnect_frees_statement_implicitly() {
```

integration_test/tests/disconnect_tests.rs
```diff
@@ -0,0 +1,172 @@
...
+                get_sql_diagnostics(
+                    HandleType::SQL_HANDLE_STMT,
+                    *ptr::addr_of!(stmt_2).cast::<Handle>()
+                ),
+            );
+
+            // Close statement 3
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLFreeStmt(stmt_3, FreeStmtOption::SQL_CLOSE as i16),
```
nbagnard, line 147: This step doesn't look mandatory if you look at the transition table for statements.
Both SQLDisconnect and SQLFreeHandle transition table look the same for statements.

nbagnard (changes requested):

integration_test/tests/disconnect_tests.rs
terakilobyte: I was just putting statements in different states to exercise the iteration and show that we won't crash when going through them.

nbagnard: yes, that part is great 👍  
I'm just saying that you don't need to do :
1. SQLFreeStmt(stmt_3)
2. SQLFreeHandle(stmt_3)
You could just call SQLFreeHandle(stmt_3) directly since SQLFreeStmt will get you from S4 to S1, but  SQLFreeHandle gets you to S0 from S4 or S1 just the same.

If you are trying to check that we can call SQLFreeHandle(stmt_3) after SQLFreeStmt(stmt_3) maybe this could be a test in free_and_cancel_tests.rs?

integration_test/tests/common/mod.rs
terakilobyte: I think we want to pass the `in_connection_string` argument since it's already an `Option<String>`!

odbc/src/api/functions.rs
```diff
@@ -1017,40 +1017,25 @@ pub unsafe extern "C" fn SQLDescribeParam(
...
     panic_safe_exec_clear_diagnostics!(
         info,
         || {
             let conn_handle = MongoHandleRef::from(connection_handle);
             let conn = must_be_valid!((*conn_handle).as_connection());
 
-            // close open cursors on any statements that haven't been closed
-            if let Ok(stmts) = conn.statements.write() {
+            if let Ok(mut stmts) = conn.statements.write() {
+                // close any open cursors on statements that haven't been dropped
```
nbagnard, line 1027: [nit]
```suggestion
                // Close any open cursors and drop all statements
```

nbagnard (approved):
LGTM!
JUst one nit left

PR #209, opened 2024-04-22
SQL-2037: Feature gate the driver logger tests to only run on Evergreen

nbagnard:


mattChiaravalloti (approved):
lgtm 👍

integration_test/tests/connection_tests.rs
```diff
@@ -109,6 +118,7 @@ mod integration {
     // the logger log level is updated to the connection log level.
     // If you are having problems running this test, ensure you are running as an administrator.
     #[test]
+    #[cfg(feature = "evergreen_tests")]
```
terakilobyte, line 121: This will be great. I was thinking about it over the weekend and remembered this an easy way to do this (from the testing overhaul I did on mongosql).

We can change this declaration here to the following
```suggestion
    #[cfg_attr(not(feature = "evergreen_tests"), ignore)]
```

And then remove all of the other added `[#cfg(..)]` directives you added. This then makes the evergreen file a lot easier.

evergreen.yml
```diff
@@ -1092,8 +1094,8 @@ functions:
           ./resources/run_adf.sh start && cargo run --bin data_loader &&
 
           # Run integration tests with ubuntu
-          cargo test integration -- --skip test_driver_log_level --nocapture &&
-          cargo test --test connection_tests integration::test_driver_log_level -- --nocapture &&
+          cargo test integration -- --nocapture &&
+          cargo test --test connection_tests integration::test_driver_log_level --features evergreen_tests -- --nocapture &&
```
terakilobyte, line 1098: If you accept my suggestion about using `cfg_attr`, I believe all integration tests can be run in one step with the following (it works locally!)
```suggestion
          cargo test integration -- features evergreen_tests -- --nocapture &&
```

terakilobyte (approved):
LGTM. I left some suggestions that I think will make this more universal. With the suggestions, running `cargo test integration` (or `cargo test --package integration_test`) runs all tests except for `test_driver_log_level`. Adding the feature `cargo test integration --features evergreen_tests` runs the log level test. Just a suggestion to cut down on the number of changes. Either works for me!

evergreen.yml
nbagnard: Thank you for the suggestion, I appreciate you taking the time to try to make things more concise.
The reason this is broken in 2 lines is by design.
Tests run in parallel and the logger test can only run by itself.
Otherwise the log files will be polluted and/or held up by the other tests.

I will add a comment so that it is clear for everyone going forward ;)

terakilobyte: We could specify `--test-threads=1`?

nbagnard: I think there is value in running all other tests in parallel and singling this one out.
Why penalize all others tests just because of one specific one is how I see it.

PR #213, opened 2024-05-02
SQL-1937: Add ODBC callback conditionally

pmeredit:
~~This is just a draft so you can see what this looks like. I have a PR up on the driver that should make this work with rust driver `main` once that is merged.~~

The most important thing here is to see how I am reusing the `Client` structs


core/src/bin/test_auth_flow.rs
```diff
@@ -1,17 +1,21 @@
-use mongo_odbc_core::oidc_auth::*;
+use mongo_odbc_core::oidc_auth::do_auth_flow;
+use mongodb::options::oidc::{CallbackContext, IdpServerInfo};
 
 #[tokio::main]
 async fn main() {
```
pmeredit, line 5: The actual public interface for the driver requires using a builder in case the types change in the future. This is the community way to do structs in public interfaces

core/src/collections.rs
```diff
@@ -108,12 +110,8 @@ impl MongoCollections {
             future::join_all(
                 mongo_connection
                     .client
```
pmeredit, line 112: A lot of these are just changes going to rust driver 3.0

core/src/conn.rs
```diff
@@ -1,10 +1,71 @@
...
+    Client,
+};
 use serde::{Deserialize, Serialize};
-use std::time::Duration;
+use std::{
+    sync::{Arc, Weak},
+    time::Duration,
+};
+use tokio::runtime::Runtime;
+
```
pmeredit, line 15: This is the interesting part

core/src/conn.rs
```diff
@@ -1,10 +1,70 @@
...
+                // ClientMap, because the Runtime associated with the Client has been dropped,
+                // which will make the Topology hang forever.
+                } else {
+                    to_remove.push(i);
+                }
+            }
+        }
+        // We have to do this this way due to the borrow checker, it would be nice if we could just
+        // remove in the else above, but we cannot.
+        for i in to_remove.into_iter().rev() {
```
pmeredit, line 44: There shouldn't be any duplicates, but better safe than sorry.

core/src/odbc_uri.rs
```diff
@@ -327,6 +332,20 @@ impl ODBCUri {
         let uuid_representation = ConnectionString::parse(uri)
             .map_err(Error::InvalidClientOptions)?
             .uuid_representation;
```
pmeredit, line 334: Here is the part where we add the callback. Since MONGODB-OIDC can only be added via URI, we know there will be a URI

core/src/odbc_uri.rs
```diff
@@ -327,6 +332,20 @@ impl ODBCUri {
...
+
+        if let Some(AuthMechanism::MongoDbOidc) = client_options
+            .credential
+            .as_ref()
+            .unwrap()
+            .mechanism
+            .as_ref()
+        {
+            use futures::future::FutureExt;
```
pmeredit, line 343: `FutureExt` is necessary for the `boxed` method

core/src/oidc_auth.rs
```diff
@@ -54,7 +30,7 @@ impl From<Error> for mongodb::error::Error {
 
 pub async fn oidc_call_back(params: CallbackContext) -> mongodb::error::Result<IdpServerResponse> {
     let sleep_duration = params
-        .timeout_seconds
```
pmeredit, line 57: `timeout_seconds` was a bad name for an `Instant`.

core/src/conn.rs
```diff
@@ -1,10 +1,92 @@
...
-use std::time::Duration;
+#[cfg(feature = "garbage_collect")]
+use std::sync::Weak;
+use std::{sync::Arc, time::Duration};
+use tokio::runtime::Runtime;
+
+// we make from UserOptions to Client and Weak<Runtime> so that we do not hold around
+// Clients and Runtimes that are no longer in use. In most cases it won't matter, but for drivers that live in
+// memory for a long time, this could be a problem.
+#[cfg(feature = "garbage_collect")]
```
pmeredit, line 19: garbage_collect is expected to only be used by users building their own tools with our driver that won't unload the driver all the time :joy:

core/src/conn.rs
```diff
@@ -68,31 +192,59 @@ impl MongoConnection {
...
+        Ok(())
+    }
+
+    #[cfg(not(feature = "garbage_collect"))]
+    pub fn shutdown(self) -> Result<()> {
+        // we need to lock the CLIENT_MAP to potentially remove the client from the map.
+        // This prevents races on the strong_count or with gc().
+        let client_map = self.runtime.block_on(async { CLIENT_MAP.lock().await });
+        if !client_map.0.is_empty() {
+            // we only want to shutdown if the CLIENT_MAP has not been used.
```
nbagnard, line 221: As soon as a connection is made with OIDC, there will be an entry in the map, no?
I think I'm missing something because the way I read this, if a user makes an OIDC connection, then shutdown will never be called and I thought never calling shutdown was what was causing SSIS to hang.

core/src/bin/test_auth_flow.rs
```diff
@@ -1,17 +1,21 @@
...
-        }),
-        refresh_token: None,
-        timeout_seconds: None,
-        version: 1,
-    };
-    println!("{:?}", do_auth_flow(c).await.unwrap());
+    let c = CallbackContext::builder()
+        .idp_info(
+            IdpServerInfo::builder()
+                .issuer("https://mongodb-dev.okta.com/oauth2/ausqrxbcr53xakaRR357".to_string())
```
bucaojit, line 9: In the JDBC driver I put these repeated URL and client id's in a separate utility file.  You may want to do that as well in case these change in the future it can be changed in one place.

core/src/conn.rs
```diff
@@ -40,22 +167,19 @@ impl MongoConnection {
...
         user_options.client_options.connect_timeout =
             login_timeout.map(|to| Duration::new(to as u64, 0));
-        let guard = runtime.enter();
-        let client = runtime.block_on(async {
-            Client::with_options(user_options.client_options).map_err(Error::InvalidClientOptions)
-        })?;
-        drop(guard);
         let uuid_repr = user_options.uuid_representation;
+        let (client, runtime) = Self::get_client_and_runtime(user_options, runtime)?;
+        // drop the Mutex as soon as possible
```
bucaojit, line 182: Is this comment still valid, I don't see the mutex being dropped here unless it's referring to in `get_client_and_runtime()`

core/src/conn.rs
```diff
@@ -1,10 +1,92 @@
...
+        self.0
+            .retain(|(_, _, weak_client)| Weak::strong_count(weak_client) != 0);
+    }
+
+    // insert inserts a new client into the client map keyed on the user options. Note this will
+    // insert duplicates and it is on the user to check if an entry already exists.
+    #[cfg(feature = "garbage_collect")]
+    fn insert(&mut self, user_options: UserOptions, client: &Client, rt: &Arc<Runtime>) {
+        self.0
+            .push((user_options, client.clone(), Arc::downgrade(rt)));
```
bucaojit, line 77: In my implementation, which is still in progress, I have it keyed on the connection properties and plan to leave out things like log level, extJsonMode, since those are more driver related than connection.

core/src/conn.rs
pmeredit: Yes, this will make it not shutdown if they are using OIDC. Now I'm thinking we may want to pool the clients even if they are not using OIDC given the thing with tableau server we saw and JDBC. And no, not calling shutdown was not what was causing SSIS to hang, the runtime will still be shutdown when the driver is unloaded, and since it's the current thread runtime, there will be no problem with hanging.

bucaojit (approved):
LGTM

core/src/err.rs
```diff
@@ -95,19 +95,11 @@ impl Error {
             | Error::InvalidClientOptions(m)
             | Error::QueryCursorUpdate(m)
             | Error::QueryExecutionFailed(m)
-            | Error::MongoParseConnectionString(m) => {
-                match m.kind.as_ref() {
-                    ErrorKind::Command(command_error) => command_error.code,
-                    // errors other than command errors probably will not concern us, but
-                    // the following is included for completeness.
-                    ErrorKind::BulkWrite(BulkWriteFailure {
```
nbagnard, line 103: Why did you remove BulkWrite and not Write?

core/src/test_config.rs
```diff
@@ -0,0 +1,3 @@
+pub const ISSUER_URL: &str = "https://mongodb-dev.okta.com/oauth2/ausqrxbcr53xakaRR357";
```
nbagnard, line 1: Is this your own test instance?

nbagnard (changes requested):
The OIDC changes all look good, but as you suggested, let's pool all clients regardless of the authentication mechanism used.

core/src/conn.rs
```diff
@@ -20,10 +102,55 @@ pub struct MongoConnection {
...
-    pub runtime: tokio::runtime::Runtime,
+    pub runtime: Arc<Runtime>,
 }
 
 impl MongoConnection {
+    fn get_client_and_runtime(
+        user_options: UserOptions,
+        runtime: Arc<Runtime>,
+    ) -> Result<(Client, Arc<Runtime>)> {
+        // If we are not authenticating via OIDC, we do not pool the Client, simply
```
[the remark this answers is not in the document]
bucaojit: @nbagnard That's right, it pools regardless of the auth mechanism I didn't make it pool for only OIDC.

core/src/err.rs
pmeredit: BulkWrite went away in rust driver 3.0

core/src/conn.rs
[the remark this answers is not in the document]
pmeredit: sgtm

core/src/test_config.rs
pmeredit: It's *our* test instance

pmeredit: If you actually want to check powerbi you'd use mongodb://sqlenginesoidctest-lxonw.a.query.mongodb-dev.net/?directConnection=true&tls=true&authMechanism=MONGODB-OIDC&appName=mongosh+2.2.5

That is only for testing the callbacks

pmeredit:
> The OIDC changes all look good, but as you suggested, let's pool all clients regardless of the authentication mechanism used.

Updated to no longer check the mechanism before pooling, and can confirm powerbi is still working without *too* many popups.

pmeredit:
Power BI still working post merge with master

core/src/test_config.rs
nbagnard: oh, I just wanted to make sure that this is a shared instance we have to maintain.
We should document things in our Wiki. Who is admin, how to access it not from PowerBI to make updates if needed, etc...

integration_test/Cargo.toml
```diff
@@ -18,9 +18,9 @@ log = "0.4"
 regex = "1"
 num-traits = "0.2.14"
 num-derive = "0.3.3"
-mongodb = "2.8.2"
+mongodb = { git = "https://github.com/mongodb/mongo-rust-driver.git", branch = "main", features = ["aws-auth", "dns-resolver"] }
```
nbagnard, line 21: I think you want to move this to 3.0 GA too don't you?

nbagnard (approved):
LGTM!
No need to retag me, but you want to use 3.0 for integration tests as well.
The toml file still points directly to the GitHub branch.

core/src/err.rs
pmeredit: Hm, the compiler told me it didn't exist. I'm going to put it back in

core/src/test_config.rs
pmeredit: That is an okta account set up for what appears to be all of mongodb employees, or at least engineers.

PR #217, opened 2024-05-22
SQL-2064: fix objectId simple type string len

jchemburkar:


mattChiaravalloti (approved):
Interesting. Why does this need to change? Looks like it's been 34 since we introduced `simple_type_info`.

jchemburkar:
Discussed with @mattChiaravalloti offline and created [SQL-2111](https://jira.mongodb.org/browse/SQL-2111) as a followup. We test the metadata of the columns under simple type info during the result set tests but we never verify the values or check the values of the data in a more user-centric way.

nbagnard (approved):

PR #223, opened 2024-06-18
SQL-2066: Update SQLFreeStmt to check that mongo_statement exists before closing cursor

mattChiaravalloti:
This PR updates `SQLFreeStmt` to handle the case when the `FreeStmtOption` is `SQL_CLOSE` and `mongo_statement` is `None`. Previously, we would return an error that wrapped a panic. Now, we no-op. I added a unit test to demonstrate we no longer panic/error.

### DIFF
--- a/odbc/src/api/free_stmt_tests.rs
+++ b/odbc/src/api/free_stmt_tests.rs
@@ -165,4 +165,22 @@ mod unit {
             );
         }
     }
+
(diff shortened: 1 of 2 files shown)


nbagnard (approved):
LGTM!

jchemburkar (approved):
lgtm!

PR #225, opened 2024-06-18
SQL-2099: add silk integration for odbc

jchemburkar:
patch with sbom task which will push sbom to silk and then pull it back: https://spruce.mongodb.com/task/mongosql_odbc_driver_code_quality_security_sbom_patch_dacf6b588115422b6d2cfff8262e4ee9f40efec1_6674c648809c1f0007f8c47b_24_06_21_00_16_20/files?execution=0

augmented sbom: https://translators-connectors-releases.s3.amazonaws.com/mongosql-odbc-driver/mongosql-odbc-snapshot.sbom.json

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -364,6 +364,7 @@ functions:
           artifactory.corp.mongodb.com/release-tools-container-registry-public-local/silkbomb:1.0 \
           upload --silk-asset-group ${SILK_ASSET_GROUP} --sbom-in /pwd/$SBOM_FINAL
           echo "-------------------------------"
+
   "pull augmented SBOM from Silk":
     - command: shell.exec
       type: test
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -336,14 +336,6 @@ functions:
           echo "<<<< Done scanning SBOM"
   
   "push SBOM Lite to Silk":
-    - command: s3.get
-      params:
-        aws_key: ${aws_key}
-        aws_secret: ${aws_secret}
-        local_file: mongosql-odbc-driver/mongo-odbc-driver.full.cdx.json
-        remote_file: mongosql-odbc-driver/artifacts/${version_id}/${build_variant}/ssdlc/mongo-odbc-driver.full.cdx.json
-        content_type: application/json
-        bucket: mciuploads
     - command: shell.exec
       type: test
       params:
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -332,6 +332,65 @@ functions:
           ./$SBOM_DIR/grype sbom:$SBOM_LICENSES --fail-on low
           echo "---------------------------------------------"
           echo "<<<< Done scanning SBOM"
+  
+  "push SBOM Lite to Silk":
+    - command: s3.get
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_file: mongosql-odbc-driver/mongo-odbc-driver.full.cdx.json
+        remote_file: mongosql-odbc-driver/artifacts/${version_id}/${build_variant}/ssdlc/mongo-odbc-driver.full.cdx.json
+        content_type: application/json
+        bucket: mciuploads
+    - command: shell.exec
+      type: test
+      params:
+        shell: bash
+        working_dir: mongosql-odbc-driver
+        script: |
+          ${prepare_shell}
+          cat << EOF > silkbomb.env
+          SILK_CLIENT_ID=${SILK_CLIENT_ID}
+          SILK_CLIENT_SECRET=${SILK_CLIENT_SECRET}
+          EOF
+          
+          echo "SBOM_FINAL = $SBOM_FINAL"
+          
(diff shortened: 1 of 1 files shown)


evergreen.yml
```diff
@@ -1685,6 +1744,8 @@ tasks:
       - func: "install rust toolchain"
       - func: set and check packages version
       - func: "generate SBOM and scan"
```
nbagnard, line 1746: Let's split up the generate and scan into 2 different parts.
This way we can push and pull an augmented SBOM even if scanning fail

nbagnard (changes requested):

evergreen.yml
```diff
@@ -332,6 +334,86 @@ functions:
...
+          
+          echo "SBOM_FINAL = $SBOM_FINAL"
+          
+          echo "-- Uploading initial SBOM Lite to Silk --"
+          docker run -i --platform="linux/amd64" --rm -v "$PWD":/pwd \
+          --env-file silkbomb.env \
+          artifactory.corp.mongodb.com/release-tools-container-registry-public-local/silkbomb:1.0 \
+          upload --silk-asset-group ${SILK_ASSET_GROUP} --sbom-in /pwd/$SBOM_FINAL
+          echo "-------------------------------"
+  "pull augmented SBOM from Silk":
```
nbagnard, line 367:
```suggestion

  "pull augmented SBOM from Silk":
```

nbagnard (approved):
LGTM!

evergreen.yml
```diff
@@ -1988,3 +2075,4 @@ buildvariants:
       - name: release
       - name: snapshot
       - name: download-center-update
+      - name: publish-augmented-sbom
```
EthanHardyMongo, line 2078:
```suggestion
      - name: publish-augmented-sbom
      - name: download-center-update
```
Natacha wanted me to flip the order of these for the silk integration on JDBC.

evergreen.yml
```diff
@@ -332,6 +334,87 @@ functions:
...
+  
+  "push SBOM Lite to Silk":
+    - command: s3.get
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_file: mongosql-odbc-driver/mongo-odbc-driver.full.cdx.json
+        remote_file: mongosql-odbc-driver/artifacts/${version_id}/${build_variant}/ssdlc/mongo-odbc-driver.full.cdx.json
+        content_type: application/json
+        bucket: mciuploads
```
EthanHardyMongo, line 346: Do you need this part? Since you're generating the sbom in the same task, I don't think you need to use the `s3.get`.

evergreen.yml
EthanHardyMongo, line 2078: Doesn't the release depend on the `sbom` task finishing since it has `publish-augmented-sbom` included?

EthanHardyMongo (changes requested):
I had a few, minor suggestions, but overall looks good!

evergreen.yml
jchemburkar: Ah good catch, I think both of your comments are related. I originally had them split up, where the pushing/pulling from silk would happen on every commit, but the publishing to the `translators-connectors-releases` bucket only happened on release. 

Once we decided to publish on every commit, since non-tagged releases will just overwrite the "snapshot" sbom, this become unnecessary -- this task actually doesn't even exist anymore.

EthanHardyMongo: Makes sense! 

One last thing: my other comment was highlighting a different `s3.get` than the one you removed. However, it looks like you don't need either. Also, I don't think you need the s3.put in `pull augmented SBOM from Silk` now that you are no longer using the `s3.get` in `publish augmented SBOM`.

evergreen.yml
```diff
@@ -332,6 +334,79 @@ functions:
...
```
EthanHardyMongo, line 346:
```suggestion
  
  "push SBOM Lite to Silk":
```

evergreen.yml
```diff
@@ -332,6 +334,79 @@ functions:
...
+          echo "-------------------------------"
+    - command: s3.put
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_file: mongosql-odbc-driver/mongo-odbc-driver.augmented.sbom.json
+        remote_file: mongosql-odbc-driver/artifacts/${version_id}/${build_variant}/ssdlc/mongo-odbc-driver.augmented.sbom.json
+        content_type: application/json
+        bucket: mciuploads
+        permissions: public-read
```
EthanHardyMongo, line 396:
```suggestion
          artifactory.corp.mongodb.com/release-tools-container-registry-public-local/silkbomb:1.0 \
          download --silk-asset-group ${SILK_ASSET_GROUP} --sbom-out /pwd/mongo-odbc-driver.augmented.sbom.json
          echo "-------------------------------"
```

EthanHardyMongo (approved):
LGTM! Like I said in my reply to a different comment, I think you can remove the `s3.get` and `s3.put` that I've highlighted in this review.

PR #230, opened 2024-06-26
SQL-2109: fix evergreen script

jchemburkar:
This adds the appropriate s3.get to the publish SBOM task, which was missed because the evergreen settings caused ssdlc tasks not to be scheduled. Those have now been updated as well, as can be seen via the evergreen patch.


nbagnard (approved):
LGTM!

terakilobyte (approved):

PR #231, opened 2024-06-26
Fix papertrail and clean up ssdlc artifacts paths

jchemburkar:


evergreen.yml
```diff
@@ -316,7 +316,7 @@ functions:
         aws_secret: ${aws_secret}
         local_files_include_filter:
           - mongosql-odbc-driver/*.cdx.json
-        remote_file: mongosql-odbc-driver/artifacts/${version_id}/ssdlc/
```
nbagnard, line 319: These should stay `${version_id}` because it is the path on mciuploads

evergreen.yml
```diff
@@ -1744,16 +1744,39 @@ functions:
...
+      params:
+        shell: bash
+        working_dir: mongosql-odbc-driver
+        script: |
+          mkdir papertrail
+    - command: s3.get
+      params:
+        aws_key: ${release_aws_key}
+        aws_secret: ${release_aws_secret}
+        local_file: mongosql-odbc-driver/papertrail/mongoodbc-signed.msi
```
nbagnard, line 1757:
```suggestion
        local_file: mongosql-odbc-driver/papertrail/${MSI_FILENAME}
```

nbagnard (approved):
LGTM!
It might be better to trace the file with the final name on S3

PR #232, opened 2024-07-05
SQL-2193: Add support for wlongvarchar and wchar types with conversions to those types

pmeredit:
For most string operations powerbi uses varchar, for whatever reason, partitioning data for scrolling direct query uses wlongvarchar. Also added wchar just in case, since it does not hurt anything.


constants/src/lib.rs
```diff
@@ -236,6 +236,8 @@ pub const MONGO_CAST_SUPPORT: u32 = SQL_CVT_CHAR
     | SQL_CVT_REAL
     | SQL_CVT_DOUBLE
     | SQL_CVT_VARCHAR
+    | SQL_CVT_WLONGVARCHAR
```
nbagnard, line 239: We should add LONGVARCHAR and WCHAR to this PR, for completeness, or shall I say to cover everything we might get from PowerBI.

core/src/bson_type_info.rs
```diff
@@ -218,6 +218,52 @@ impl BsonTypeInfo {
         column_size: make_default_attr_func!(Some(MAX_STRING_SIZE)),
         simple_type_info: None,
     };
+    pub const WLONGVARCHAR: BsonTypeInfo = BsonTypeInfo {
```
nbagnard, line 221: Can you add a similar comment than the one on top of "VARCHAR" explaining that this was added for PowerBI and is necessary because of the way the apply a cast using the type name.

core/src/bson_type_info.rs
```diff
@@ -218,6 +218,52 @@ impl BsonTypeInfo {
...
+        literal_suffix: Some("'"),
+        sql_code: None,
+        is_auto_unique_value: None,
+        is_unsigned: None,
+        num_prec_radix: None,
+        decimal_digit: None,
+        column_size: make_default_attr_func!(Some(MAX_STRING_SIZE)),
+        simple_type_info: None,
+    };
+    pub const WCHAR: BsonTypeInfo = BsonTypeInfo {
```
nbagnard, line 244: Can you add a similar comment than the one on top of "VARCHAR" explaining that this was added for PowerBI and is necessary because of the way the apply a cast using the type name.

nbagnard (changes requested):

core/src/err.rs
```diff
@@ -40,7 +40,7 @@ pub enum Error {
     QueryCursorUpdate(mongodb::error::Error),
     #[error("Getting metadata for query failed with error: {0}")]
     QueryDeserialization(mongodb::bson::de::Error),
-    #[error("Trying to execute query failed with error: {0}")]
+    #[error("Trying to execute query failed with error: {0:?}")]
```
pmeredit, line 43: I noticed a lot of the displays for query errors were just "user defined error", so I changed it to use debug to hopefully get us more info when it comes up.

constants/src/lib.rs
nbagnard: uhm, I think it's not just here that the change is needed, but everywhere else in this PR.
It needs to be added in bson_type_info.rs, type_info.rs and in the tests.
Otherwise we will end up with the same problem where PowerBI want to cast a value to LONGVARCHAR or WCHAR and it fails because there is no LONGVARCHAR and WCHAR types in SQLGetTypeInfo, no?
I suppose that we should map them to the "string" name too 🤔
Maybe we should try to do more tests, because we list SQL_CVT_CHAR, but there is no SQL_CHAR reported by SQLGetTypeInfo. I have no way to tell if PowerBI will map something to SQL_CHAR or not. It is technically the data type for fixed length string, so if data is consistent in length, they might be using it, like 'Y', 'N' for boolean representation or something like that.

nbagnard (changes requested):

constants/src/lib.rs
pmeredit: WCHAR is there, it's missing LONGVARCHAR however

nbagnard (approved):
LGTM!
Just a very small request, before you merge, can you add a comment on top of `pub const STRING: BsonTypeInfo = BsonTypeInfo` to call out that this is representing the real MongoDB string type.

PR #234, opened 2024-07-17
SQL-2196: Add type mappings for all easily mappable types

pmeredit:
These were all easy enough to add. I think we should consider supporting numeric, especially for decimal128 types, but that has to be done in another ticket because it will be quite complex especially in terms of tests to ensure that we are correctly creating the numeric c struct.

I actually thought I would need to update the data retrieving for SQL_REAL, at least, but I had forgotten that we actually already supported retrieving data as f32.

I'd also like to say that SQL_REAL is a terrible name for 32 bit float. You'd think it were referring to, at least, a very large floating point type.


core/src/bson_type_info.rs
```diff
@@ -169,6 +169,32 @@ impl BsonTypeInfo {
         column_size: make_default_attr_func!(Some(15)),
         simple_type_info: None,
     };
+    // We support REAL for any types that use it. We map it to "double" as the mongo name for the
+    // purposes of CAST in the syntax (e.g., it will generate CAST(x AS DOUBLE) in Direct Query).
+    // For retrieving Data, we will need to cast the f64 to an f32.
```
pmeredit, line 174:
```suggestion
```

core/src/bson_type_info.rs
```diff
@@ -169,6 +169,31 @@ impl BsonTypeInfo {
...
+    pub const REAL: BsonTypeInfo = BsonTypeInfo {
+        type_name: "double",
+        sql_type: SqlDataType::SQL_REAL,
+        non_concise_type: None,
+        searchable: SQL_PRED_BASIC,
+        is_case_sensitive: false,
+        fixed_prec_scale: false,
+        scale: Some(0),
+        length: make_default_attr_func!(None),
+        precision: Some(15),
```
nbagnard, line 183: Precision for real is 7.
Even if we map to Bson double, we should keep the SQL type info accurate.

core/src/bson_type_info.rs
```diff
@@ -169,6 +169,31 @@ impl BsonTypeInfo {
...
+        char_octet_length: make_default_attr_func!(None),
+        transfer_octet_length: Some(8),
```
nbagnard, line 185: Same here, octet length should be 4

core/src/bson_type_info.rs
```diff
@@ -169,6 +169,31 @@ impl BsonTypeInfo {
...
+        display_size: make_default_attr_func!(Some(24)),
+        literal_prefix: None,
+        literal_suffix: None,
+        sql_code: None,
+        is_auto_unique_value: Some(false),
+        is_unsigned: Some(false),
+        num_prec_radix: Some(2),
+        decimal_digit: Some(0),
+        column_size: make_default_attr_func!(Some(15)),
```
nbagnard, line 194: Here is 7

core/src/bson_type_info.rs
```diff
@@ -625,6 +698,56 @@ impl BsonTypeInfo {
...
+    pub const TINYINT: BsonTypeInfo = BsonTypeInfo {
+        type_name: "int",
+        sql_type: SqlDataType::SQL_TINYINT,
+        non_concise_type: None,
+        searchable: SQL_PRED_BASIC,
+        is_case_sensitive: false,
+        fixed_prec_scale: true,
+        scale: Some(0),
+        length: make_default_attr_func!(None),
+        precision: Some(10),
```
nbagnard, line 712:
```suggestion
        precision: Some(3),
```

core/src/bson_type_info.rs
```diff
@@ -625,6 +698,56 @@ impl BsonTypeInfo {
...
+        char_octet_length: make_default_attr_func!(None),
+        transfer_octet_length: Some(4),
```
nbagnard, line 714:
```suggestion
        transfer_octet_length: Some(1),
```

core/src/bson_type_info.rs
```diff
@@ -625,6 +698,56 @@ impl BsonTypeInfo {
...
+        display_size: make_default_attr_func!(Some(11)),
+        literal_prefix: None,
+        literal_suffix: None,
+        sql_code: None,
+        is_auto_unique_value: Some(false),
+        is_unsigned: Some(false),
+        num_prec_radix: Some(10),
+        decimal_digit: Some(0),
+        column_size: make_default_attr_func!(Some(10)),
```
nbagnard, line 723:
```suggestion
        column_size: make_default_attr_func!(Some(3)),
```

core/src/bson_type_info.rs
```diff
@@ -625,6 +698,56 @@ impl BsonTypeInfo {
...
+    pub const SMALLINT: BsonTypeInfo = BsonTypeInfo {
+        type_name: "int",
+        sql_type: SqlDataType::SQL_SMALLINT,
+        non_concise_type: None,
+        searchable: SQL_PRED_BASIC,
+        is_case_sensitive: false,
+        fixed_prec_scale: true,
+        scale: Some(0),
+        length: make_default_attr_func!(None),
+        precision: Some(10),
```
nbagnard, line 737:
```suggestion
        precision: Some(5),
```

core/src/bson_type_info.rs
```diff
@@ -625,6 +698,56 @@ impl BsonTypeInfo {
...
+        char_octet_length: make_default_attr_func!(None),
+        transfer_octet_length: Some(4),
+        display_size: make_default_attr_func!(Some(11)),
```
nbagnard, line 740:
```suggestion
        display_size: make_default_attr_func!(Some(6)),
```

core/src/bson_type_info.rs
```diff
@@ -625,6 +698,56 @@ impl BsonTypeInfo {
...
+        literal_prefix: None,
+        literal_suffix: None,
+        sql_code: None,
+        is_auto_unique_value: Some(false),
+        is_unsigned: Some(false),
+        num_prec_radix: Some(10),
+        decimal_digit: Some(0),
+        column_size: make_default_attr_func!(Some(10)),
```
nbagnard, line 748:
```suggestion
        column_size: make_default_attr_func!(Some(5)),
```

nbagnard:
> I think we should consider supporting numeric, especially for decimal128 types, but that has to be done in another ticket because it will be quite complex especially in terms of tests to ensure that we are correctly creating the numeric c struct.

You mean reporting decimal128 as SQL_NUMERIC in this case. I thought the issue is that the precision is lower and we would/could loose precision that's why we stayed with a string representation for now. If we had info on the max precision for a decimal128 field or an option to accept truncation, then I think we could get something to map. But it would have to be an opt in feature probably.

Adding the types in SQLGetTypeInfo is really only make sure that if DirectQuery picks a type for a column (even if it is not the type reported), then it will generate the a viable SQL query, isn't it? If not, can you please explain a bit more because then it means that I missed something Power BI is doing.

nbagnard (changes requested):
Let's report the SQL type information accurately.
And, if you want to know like me why we need to report the info since it's already defined in the spec, I don't know...

pmeredit:
> > I think we should consider supporting numeric, especially for decimal128 types, but that has to be done in another ticket because it will be quite complex especially in terms of tests to ensure that we are correctly creating the numeric c struct.
> 
> You mean reporting decimal128 as SQL_NUMERIC in this case. I thought the issue is that the precision is lower and we would/could loose precision that's why we stayed with a string representation for now. If we had info on the max precision for a decimal128 field or an option to accept truncation, then I think we could get something to map. But it would have to be an opt in feature probably.
> 
> Adding the types in SQLGetTypeInfo is really only make sure that if DirectQuery picks a type for a column (even if it is not the type reported), then it will generate the a viable SQL query, isn't it? If not, can you please explain a bit more because then it means that I missed something Power BI is doing.

To my knowledge they never try to cast to numeric. The problem with stating that we support numeric comes in when someone tries to then retrieve data as numeric, which uses a special C struct.

pmeredit:
> Let's report the SQL type information accurately. And, if you want to know like me why we need to report the info since it's already defined in the spec, I don't know...

:joy:

nbagnard:
> To my knowledge they never try to cast to numeric. The problem with stating that we support numeric comes in when someone tries to then retrieve data as numeric, which uses a special C struct.
oh, you were thinking that part. Yes, we would have to read the C struct and convert it to a Decimal128 correctly if a tool was passing in data.
But, unless we support parameters, I don't see it happening.  PowerBI generates a SQL query, so I don't think it applies.
Anyway, if they don't use that type internally, we should be good to go 👍

core/src/bson_type_info.rs
pmeredit: colum_size too then

resources/integration_test/tests/functions.yml
```diff
@@ -92,35 +92,35 @@ tests:
...
-      - [ "undefined", -9, 20, null, null, null, 1, 0, 0, null, 0, null, "undefined", null, null, -9, null, null, null ]
-      - [ "string", 1, 65535, "'", "'", null, 1, 1, 3, null, 0, null, "string", null, null, 1, null, null, null ]
-      - [ "int", 4, 10, null, null, null, 1, 0, 2, 0, 1, 0, "int", 0, 0, 4, null, 10, null ]
-      - [ "int", 5, 10, null, null, null, 1, 0, 2, 0, 1, 0, "int", 0, 0, 5, null, 10, null ]
-      - [ "double", 7, 15, null, null, null, 1, 0, 2, 0, 0, 0, "double", 0, 0, 7, null, 2, null ]
-      - [ "double", 8, 15, null, null, null, 1, 0, 2, 0, 0, 0, "double", 0, 0, 8, null, 2, null ]
-      - [ "date", 11, 23, "'", "'", null, 1, 0, 2, null, 1, null, "date", 3, 3, 11, 3, null, null ]
-      - [ "varchar", 12, 65535, "'", "'", null, 1, 1, 3, null, 0, null, "varchar", null, null, 12, null, null, null ]
-      - [ "date", 93, 23, "'", "'", null, 1, 0, 2, null, 1, null, "date", 3, 3, 93, 3, null, null ]
+    - [ "string", -10, 65535, "'", "'", null, 1, 1, 3, null, 0, null, "string", null, null, -10, null, null, null ]
```
pmeredit, line 95: Sorry about the space deleting here, but it was driving me crazy that this test didn't line up with the other tests

core/src/bson_type_info.rs
```diff
@@ -560,6 +585,29 @@ impl BsonTypeInfo {
         column_size: make_default_attr_func!(Some(23)),
         simple_type_info: None,
     };
+    pub const DATETIME: BsonTypeInfo = BsonTypeInfo {
```
nbagnard, line 588: SQL_DATETIME is not special SQL type.
This is only used for the verbose type name, which in conjunction with the sql code gives you the final type.
See here : https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/data-type-identifiers-and-descriptors?view=sql-server-ver16

core/src/bson_type_info.rs
```diff
@@ -840,6 +891,58 @@ impl BsonTypeInfo {
         column_size: make_default_attr_func!(None),
         simple_type_info: SimpleTypeInfo::default(),
     };
+    // This is just here to support any tools that attempt to cast to NUMERIC.
+    pub const NUMERIC: BsonTypeInfo = BsonTypeInfo {
```
nbagnard, line 895: [Note]
I have a feeling that things can still fail, because Numeric is for exact numeric value which means a fixed precision and scale.
We don't define any boundaries for it. So, similarly to what happened for VARCHAR, PowerBI might not like the unknown.

constants/src/lib.rs
```diff
@@ -220,18 +217,18 @@ pub const SQL_CVT_LONGVARBINARY: u32 = 0x00040000;
...
 pub const SQL_CVT_WVARCHAR: u32 = 0x00800000;
 #[allow(unused)]
 pub const SQL_CVT_GUID: u32 = 0x01000000;
 pub const MONGO_CAST_SUPPORT: u32 = SQL_CVT_CHAR
     | SQL_CVT_NUMERIC
     | SQL_CVT_INTEGER
     | SQL_CVT_SMALLINT
+    | SQL_CVT_TINYINT
+    | SQL_CVT_BIGINT
+    | SQL_CVT_NUMERIC
```
nbagnard, line 231: Looks like SQL_CVT_DECIMAL is missing.

nbagnard (changes requested):
A few very small nits left

core/src/bson_type_info.rs
pmeredit: This standard is just incredibly messy

constants/src/lib.rs
pmeredit: Absolutely, thanks!

core/src/bson_type_info.rs
pmeredit: That was only a problem because it was trying to find a type big enough to contain 4 for something like "TRUE". I'm not sure what to do here, we might want to just discuss it quickly when you are back

jchemburkar (approved):
spent some more time with the spec today, i think this all lgtm 🙏

nbagnard (changes requested):
DATETIME still exists in core/src/bson_type_info.rs. It won't show up because it is not listed in `const DATA_TYPES`, but we might has well have a clean code.

nbagnard (approved):
LGTM

PR #235, opened 2024-07-29
SQL-2187: update release steps

jchemburkar:
this just copies over the changes from JDBC that were done as part of SQL-2186


nbagnard (approved):
LGTM! 
It looks like we had something in there, just not something super clear

PR #237, opened 2024-08-09
SQL-2272: sanitize deps

terakilobyte:
Sanitizing task and work for odbc. Here the `--all-features` trick doesn't work because some of the features enable things that aren't compatible with all systems. [Here is a patch](https://spruce.mongodb.com/version/66b6501b76d45c000746d996/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) showing failure because of an unused dependency.

### DIFF
--- a/cstr/Cargo.toml
+++ b/cstr/Cargo.toml
@@ -19,3 +19,6 @@ utf32 = []
 
 [lints]
 workspace = true
+
+[package.metadata.cargo-machete]
+ignored = ["num-traits"]
--- a/definitions/Cargo.toml
+++ b/definitions/Cargo.toml
@@ -17,3 +17,6 @@ iodbc = []
 odbc_version_3_50 = []
 odbc_version_3_80 = ["odbc_version_3_50"]
 odbc_version_4 = ["odbc_version_3_80"]
+
+[package.metadata.cargo-machete]
+ignored = ["num-traits"]
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -177,6 +177,22 @@ functions:
           ${prepare_shell}
           cargo fmt --all --  --check
 
+  "check unused dependencies":
+    - command: shell.exec
+      type: test
+      params:
+        shell: bash
+        working_dir: mongosql-odbc-driver
+        script: |
+          ${prepare_shell}
+          cargo install cargo-machete
+          cargo machete
+          RETURN=$?
+          if [ $RETURN -ne 0 ]; then
+            echo "Unused dependencies found"
+            cargo machete
+          fi
+
   "generate SBOM":
     - command: shell.exec
       type: test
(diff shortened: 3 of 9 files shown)


cstr/Cargo.toml
```diff
@@ -19,3 +19,6 @@ utf32 = []
 
 [lints]
 workspace = true
+
+[package.metadata.cargo-machete]
+ignored = ["num-traits"]
```
terakilobyte, line 24: false positive that needed to be ignored

definitions/Cargo.toml
```diff
@@ -17,3 +17,6 @@ iodbc = []
 odbc_version_3_50 = []
 odbc_version_3_80 = ["odbc_version_3_50"]
 odbc_version_4 = ["odbc_version_3_80"]
+
+[package.metadata.cargo-machete]
+ignored = ["num-traits"]
```
terakilobyte, line 22: false positive that needed to be ignored

evergreen.yml
```diff
@@ -270,18 +286,18 @@ functions:
...
           fi
           touch $temp_output
-          
+
           while IFS= read -r line
           do
             if [[ "$line" == *"purl"* ]]; then
               bash_purl=$(echo $line | cut -d '"' -f4)
-              command=$(echo "./$SBOM_DIR/jq '.components[] | select(.purl == \"$bash_purl\").licenses' $SBOM_LICENSES")
+              command=$(echo "./$SBOM_DIR/jq '.components[] | select(.purl == \"$bash_purl\").licenses' $SBOM_LICENSES | ./$SBOM_DIR/jq -s 'flatten(1)'")
```
terakilobyte, line 300: I had to update this because ultimately, jq finding something like
```
[ {"license": "MIT"} ] [ {"license": "MIT"} ] 
```
and falling over trying to add the team responsible info because it wasn't comma separated. Adding this additional command combines them into a single array, and allows the write to succeed.

nbagnard: Weird. It looks like it was the new version of cargo-cyclonedx which broke the flow.

evergreen.yml
```diff
@@ -177,6 +177,22 @@ functions:
...
+      params:
+        shell: bash
+        working_dir: mongosql-odbc-driver
+        script: |
+          ${prepare_shell}
+          cargo install cargo-machete
+          cargo machete
+          RETURN=$?
+          if [ $RETURN -ne 0 ]; then
+            echo "Unused dependencies found"
```
nbagnard, line 192: [nit] Can you use >&2 echo "xxx" to send it to stderr instead of stdout so that it shows in red on Evergreen.

nbagnard (approved):
LGTM!

jchemburkar (approved):
lgtm

PR #238, opened 2024-08-26
SQL-2287: Create integration test to test mongodb+srv style connection string with ODBC Driver

EthanHardyMongo:
This PR creates the "mongodb+srv" uri connection string test. Like what was done in the JDBC Driver, as of now, I have the test expecting the error that occurs. I added a TODO because this will need to be updated as part of SQL-2291.


integration_test/tests/connection_tests.rs
```diff
@@ -43,6 +43,18 @@ mod integration {
         let _ = unsafe { Box::from_raw(env_handle) };
     }
 
+    #[test]
+    fn test_srv_style_uri_connection() {
+        let env_handle = allocate_env(AttrOdbcVersion::SQL_OV_ODBC3);
+        let conn_str = crate::common::generate_srv_style_connection_string();
+        let result = connect_with_conn_string(env_handle, Some(conn_str));
+
+        // TODO: SQL-2291: Support direct cluster mode (This should no longer expect an Error after that).
```
bucaojit, line 52: Let's add a commented line that can be uncommented once SQL-2291 is implemented.   Something like:
```
// assert!(connection_result.is_ok(), "Expected successful connection, got error: {:?}", connection_result);
```

mattChiaravalloti (approved):
looks good pending Oliver's suggestion!

PR #239, opened 2024-09-04
SQL-2296: Load mongosqltranslate library

bucaojit:
Adds functionality to load the mongosqltranslate library if present in the same directory as where the ODBC driver is installed. 
Used the paths in the installer to determine the installation path. 
Added a mock library with `runCommand()` to test that the loading and running of the function is successful.


core/src/load_library.rs
```diff
@@ -0,0 +1,132 @@
...
+#[cfg(target_os = "windows")]
+const LIBRARY_INSTALL_PATH: &str = if cfg!(target_arch = "x86_64") {
+    "C:\\Program Files\\MongoDB\\Atlas SQL ODBC Driver\\bin"
+} else {
+    "C:\\Program Files\\MongoDB\\ODBC\\bin"
+};
+#[cfg(target_os = "macos")]
+const LIBRARY_INSTALL_PATH: &str = "/Library/MongoDB/MongoDB Atlas SQL ODBC Driver/";
+#[cfg(target_os = "linux")]
+const LIBRARY_INSTALL_PATH: &str = "/opt/mongodb/atlas-sql-odbc-driver/";
```
bucaojit, line 27: We don't have this specified for `linux`, I added it to pass the evergreen tests.  We can modify it or remove support.

odbc/src/api/functions.rs
```diff
@@ -254,6 +255,8 @@ fn sql_alloc_handle(
     input_handle: *mut MongoHandle,
     output_handle: *mut Handle,
 ) -> Result<()> {
+    load_mongosqltranslate_library();
```
bucaojit, line 258: This seemed like a good spot to load it, since the allocating of the handle was usually the first operation.

mock_mongosqltranslate/rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "nightly"
```
terakilobyte, line 2: I might have missed it while I was out, but what do we need nightly for?

terakilobyte (commented):
Nothing major I noticed, nice job. Just a question.

mock_mongosqltranslate/rust-toolchain.toml
bucaojit: I added that for the mock library when I was more closely following what was in the design with this line:
```
// Convert from bson::Document to an array of BSON bytes
let bytes = bson::to_vec(&result_docs).expect(<message>);
let (ptr, len, cap) = bytes.into_raw_parts();  <-----
(ptr.cast_const(), len, cap)
```
The [Vec::into_raw_parts()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_raw_parts) needed the nightly toolchain.  I modified the library to not need that function and forgot to remove this file, I'll do that now it should work without it.  Thanks for bringing it up.

terakilobyte (approved):

EthanHardyMongo (approved):
LGTM!

PR #240, opened 2024-09-13
SQL-2290: Differentiate between cluster types on connection

bucaojit:
Adds change to check whether the cluster connected to is community edition, enterprise, or ADF.  
Removes the `SELECT 1` query and instead uses the check for buildinfo. 
Added integration test that has a script to start `mongod` community and enterprise versions taken from `mongo-jdbc-driver` repo.

### DIFF
--- a/core/src/cluster_type.rs
+++ b/core/src/cluster_type.rs
@@ -27,6 +27,7 @@ pub async fn determine_cluster_type(client: &Client) -> MongoClusterType {
     // if "ok" is not 1, then the target type could not be determined.
     match cmd_res.get("ok") {
         Some(Bson::Double(f)) if *f == 1.0 => {}
+        Some(Bson::Int64(i)) if *i == 1 => {}
         Some(Bson::Int32(i)) if *i == 1 => {}
         _ => {
             log::error!(
--- a/core/src/cluster_type.rs
+++ b/core/src/cluster_type.rs
@@ -1,4 +1,4 @@
-use mongodb::bson::{doc, Document};
+use mongodb::bson::{doc, Bson, Document};
 use mongodb::Client;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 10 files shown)


core/src/conn.rs
```diff
@@ -160,22 +162,42 @@ impl MongoConnection {
...
+            }
+        }
+
         let connection = MongoConnection {
             client,
             operation_timeout: operation_timeout.map(|to| Duration::new(u64::from(to), 0)),
             uuid_repr,
             runtime,
         };
-        // Verify that the connection is working and the user has access to the default DB
```
bucaojit, line 169: Not sure if we want to remove this check since it verifies access to the default db which the cluster detection doesn't do.

core/src/cluster_type.rs
```diff
@@ -0,0 +1,60 @@
...
+        Err(e) => {
+            log::error!("Failed to run buildInfo command: {:?}", e);
+            return MongoClusterType::UnknownTarget;
+        }
+    };
+
+    // if "ok" is not 1, then the target type could not be determined.
+    match cmd_res.get("ok") {
+        Some(Bson::Double(f)) if *f == 1.0 => {}
+        Some(Bson::Int32(i)) if *i == 1 => {}
```
mattChiaravalloti, line 30: Should we also include a `Bson::Int64` case?

core/src/conn.rs
mattChiaravalloti: Oh that is a good point. Is that permission check required at connection time according to the ODBC spec, or is it just a nice-to-have to fail earlier? I don't think there's any harm in keeping this, assuming we eventually decide to only support versions of on-prem mongo that support `$documents`.

mattChiaravalloti (approved):
Looks good. One small idea with respect to possible numeric types for `ok`, and I shared my 2 cents on the `select 1` question. Probably no need to retag me.

core/src/conn.rs
EthanHardyMongo: hmm maybe we should keep it then. Is it possible to have access to the admin DB but not the default DB?

EthanHardyMongo (commented):
LGTM! I think we should probably keep the `Select 1`. It doesn't seem like there's any harm in keeping it.

EthanHardyMongo (approved):

core/src/conn.rs
bucaojit: I looked through the spec and it didn't have guidance on a permission check on connection.  I do think it would be helpful to fail earlier so I'll keep it.  Thanks for the comments on this.

core/src/cluster_type.rs
bucaojit: Sure I'll add that.

PR #241, opened 2024-09-16
 SQL-2299: Use github token for cloning instead of ssh

bucaojit:
Copying changes from the [mongo-jdbc-repo](https://github.com/mongodb/mongo-jdbc-driver/pull/283) over. 

Running test on host without ssh keys [here](https://spruce.mongodb.com/version/66e8c5754198dd000734ee38/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&statuses=success).

### DIFF
--- a/resources/run_adf.sh
+++ b/resources/run_adf.sh
@@ -49,9 +49,18 @@ GO="$GOBINDIR/go"
 
 PATH=$GOBINDIR:$PATH
 
+# GITHUB_TOKEN must be set for cloning 10gen/mongohouse repository from Evergreen hosts, and the
+# dependencies needed to build it.
+# If unset, it will default to using the SSH private key on the local system.
+if [[ ${GITHUB_TOKEN} ]]; then
+  MONGOHOUSE_URI=https://x-access-token:${GITHUB_TOKEN}@github.com/10gen/mongohouse.git
+  git config --global url."https://x-access-token:${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
+else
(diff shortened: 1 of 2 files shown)


resources/run_adf.sh
```diff
@@ -49,9 +49,24 @@ GO="$GOBINDIR/go"
 
 PATH=$GOBINDIR:$PATH
 
+# GITHUB_TOKEN must be set for cloning 10gen/mongohouse repository from Evergreen hosts, and the
+# dependencies needed to build it.
+# If unset, it will default to using the SSH private key on the local system.
+if [[ ${GITHUB_TOKEN} ]]; then
+  # Clear git url configurations if they exist
+  git config --global --get-regexp '^url\.' | while read -r key _; do
```
bucaojit, line 57: The global git config doesn't get reset by Evergreen so this was causing intermittent failures.  I see they have a ticket to consider fixing that [here](https://jira.mongodb.org/browse/DEVPROD-10373).  This change will make it so we don't try to use an existing token configuration.  I will make this change in the jdbc repo also.

pmeredit (approved):
LGTM!

resources/run_adf.sh
nbagnard: QQ: Why is not necessary for the PowerBI connector run_adf script?

nbagnard (approved):
LGTM :)

resources/run_adf.sh
bucaojit: Oh whoops, I forgot to push the commit.  It's there now in the powerbi connector. 👍

PR #242, opened 2024-09-23
SQL-2291: Implement direct cluster flow in ODBC Driver

EthanHardyMongo:
This PR implements the direct cluster flow. The testing for the direct cluster flow will be implemented during [SQL-2288](https://jira.mongodb.org/browse/SQL-2288).


core/src/util/mod.rs
```diff
@@ -24,6 +28,52 @@ lazy_static! {
...
+    let run_command_function: Symbol<
+        'static,
+        unsafe extern "C" fn(LibmongosqltranslateDataIO) -> LibmongosqltranslateDataIO,
+    > = unsafe { library.get(b"runCommand") }.expect("Failed to load runCommand symbol.");
+
+    let command_type = command.get_str("command").unwrap();
+
+    let command_bytes_vec =
+        bson::to_vec(&command).expect("Failed to serialize Document into BSON bytes.");
+
```
EthanHardyMongo, line 43: Since `command` is a document, we can't use `serde_stacker`. The only other way I could find to do this would be to use [to_writer](https://docs.rs/bson/2.11.0/bson/document/struct.Document.html#method.to_writer).

core/src/util/mod.rs
```diff
@@ -24,6 +28,52 @@ lazy_static! {
...
+    let decomposed_returned_doc = unsafe { run_command_function(libmongosqltranslate_command) };
+
+    let returned_doc: Document = unsafe {
+        bson::from_slice(
+            Vec::from_raw_parts(
+                decomposed_returned_doc.data.cast_mut(),
+                decomposed_returned_doc.length,
+                decomposed_returned_doc.capacity,
+            )
+            .as_slice(),
```
EthanHardyMongo, line 63: Since we're deserializing into a `Document`, `serde_stacker` can't be used. The only other way I could find to do this would be to use [from_reader](https://docs.rs/bson/2.11.0/bson/document/struct.Document.html#method.from_reader).

core/src/conn.rs
```diff
@@ -165,6 +211,20 @@ impl MongoConnection {
 
         load_mongosqltranslate_library();
 
+        let is_libmongosqltranslate_compatible_with_driver_version =
+            if get_mongosqltranslate_library().is_some() {
+                let libmongosqltranslate_version = Self::get_libmongosqltranslate_version()?;
+
+                // TODO where do I put the library version for the logs?
+                dbg!(libmongosqltranslate_version);
```
EthanHardyMongo, line 219: I know I'm supposed to add the library version to the client metadata, but I wasn't sure where I was supposed to add it. Is it supposed to be added to the `ClientOptions.app_name` or `ClientOptions.driver_info`? Or somewhere else?

core/src/load_library.rs
```diff
@@ -29,13 +30,6 @@ const LIBRARY_INSTALL_PATH: &str = "/opt/mongodb/atlas-sql-odbc-driver/";
 static INIT: Once = Once::new();
 static mut MONGOSQLTRANSLATE_LIBRARY: Option<Library> = None;
 
-#[repr(C)]
-pub struct RunCommandResult {
-    result_ptr: *const u8,
-    result_len: usize,
-    result_cap: usize,
-}
-
```
EthanHardyMongo, line 38: I refactored this file to use `LibmongosqltranslateDataIO` instead of `RunCommandResult` because they are ultimately the same thing.

core/src/util/mod.rs
terakilobyte: I'm not sure it's a concern on the serialization side, it's more the deserialization where we see recursive blowup.

core/src/conn.rs
terakilobyte: This should be added into the appname.

core/src/load_library.rs
terakilobyte: More generally, we could  call it `BsonBuffer`. I'm not sure we should be adding "Result" to the name, since the type doesn't reflect that. Thoughts?

EthanHardyMongo: This is what's currently being used:
```
#[repr(C)]
pub struct LibmongosqltranslateDataIO {
    pub data: *const u8,
    pub length: usize,
    pub capacity: usize,
}
```
I got rid of `RunCommandResult`

core/src/conn.rs
EthanHardyMongo: Thanks! Done!

core/src/conn.rs
```diff
@@ -133,6 +138,47 @@ impl MongoConnection {
...
+            "command": "getMongosqlTranslateVersion",
+            "options": {},
+        };
+
+        let returned_doc =
+            handle_libmongosqltranslate_command(get_mongosqltranslate_version_command)?;
+
+        let libmongosqltranslate_version = returned_doc
+            .get("version")
+            .expect(
```
bucaojit, line 152: One general comment, I see quite a few `expect()`s added in the patch.  Take another pass at some of these and see where an adding an error variant would make more sense if it can be returned instead of panicking.

core/src/query.rs
```diff
@@ -30,7 +36,116 @@ pub struct MongoQuery {
...
+            current_db: db_doc
+        };
+
+        let translate_command = doc! {
+            "command": "translate",
+            "options": {
+                "sql": sql_query,
+                "db": current_db,
+                "excludeNamespaces": false,
+                "relaxSchemaChecking": true,
```
bucaojit, line 137: Is true the default for relax schema checking?  I was looking for where that is set, I'll have to update it in my PR to true.

bucaojit (commented):
Had a few initial comments I'll finish reviewing tomorrow.

terakilobyte (changes requested):
Also an initial review. I'm going over this on my Windows computer but wanted to get the review started. Wholly agree with @bucaojit about not panicking!

core/Cargo.toml
```diff
@@ -17,6 +17,7 @@ thiserror = "1"
 constants = { path = "../constants" }
 regex = "1.6.0"
 serde = { version = "1", features = ["derive"] }
+serde_stacker = "0.1"
```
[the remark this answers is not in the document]
EthanHardyMongo: Oh I just meant for the places I pointed out, we can't use it. However, it was used when deserializing into the `Translation` and `Namespace` types.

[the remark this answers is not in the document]
bucaojit: I was wondering about that because now we have that test that checks for unused dependencies that would have failed if you weren't using `serde_stacker`.

core/src/query.rs
EthanHardyMongo: I couldn't find it used anywhere, and when I looked into `mongohouse`, it seemed like they ignore it. 

This is their TranslateSQL function, however, they don't even have a parameter for `relaxSchemaChecking` mode, so I just always chose the default value (relaxed).  

```
func TranslateSQL(dbName, sqlStatement string, schemas NamespaceToSchemaMap, excludeNamespaces bool) (Translation, error) {
	translation, err := mongosql_rs.Translate(mongosql_rs.TranslationArgs{
		DB:                dbName,
		SQL:               sqlStatement,
		CatalogSchema:     schemas,
		ExcludeNamespaces: excludeNamespaces,
	})
	if err != nil {
		return Translation{}, convertSafeSQLError(err)

	}

	ret := Translation{
		Pipeline:         translation.Pipeline,
		TargetDB:         translation.TargetDB,
		TargetCollection: translation.TargetCollection,
		ResultSetSchema:  translation.ResultSetSchema,
		SelectOrder:      translation.SelectOrder,
	}
	return ret, nil
}
```

core/src/query.rs
```diff
@@ -169,10 +338,46 @@ impl MongoStatement for MongoQuery {
             _ => Error::QueryExecutionFailed(e),
         };
 
-        let _guard = connection.runtime.enter();
-        let cursor: Cursor<Document> = connection
-            .runtime
-            .block_on(async { aggregate.await.map_err(map_query_error) })?;
+        let cursor: Cursor<Document> = if let Some(c) = collection {
```
nbagnard, line 345: No need for so much code duplication.
You can simply do
```
    let mut aggregate : Aggregate = if let Some(c) = collection {
                        let collection = db.collection::<Document>(c.as_str());
            collection.aggregate(pipeline).comment(stmt_id)
        }
        else {
            db.aggregate(pipeline).comment(stmt_id)
        };
```
And then keep the common code as before.

You can even take out the duplication of .comment and add it after the initial creation of the aggregate variable.

core/src/query.rs
```diff
@@ -30,7 +36,122 @@ pub struct MongoQuery {
...
+    }
+
+    fn translate_sql(
+        sql_query: &str,
+        current_db: &String,
+        namespaces: BTreeSet<Namespace>,
+        client: &MongoConnection,
+        db: &Database,
+    ) -> Result<Translation> {
+        let schema_collection = db.collection::<Document>("__sql_schemas_");
```
bucaojit, line 103:
```suggestion
        let schema_collection = db.collection::<Document>("__sql_schemas");
```
The collection that the schema builder tool creates doesn't have the `_` suffix

core/src/err.rs
```diff
@@ -54,6 +54,24 @@ pub enum Error {
     UnsupportedOperation(&'static str),
     #[error("Statement not executed")]
     StatementNotExecuted,
+    #[error("The driver version `{0}` is incompatible with libmongosqltranslate")]
+    LibmongosqltranslateLibraryIsIncompatible(&'static str),
+    #[error("The schema document for collection `{0}` could not be found in the `__sql_schemas_` collection")]
```
bucaojit, line 59:
```suggestion
    #[error("The schema document for collection `{0}` could not be found in the `__sql_schemas` collection")]
```

definitions/src/lib.rs
```diff
@@ -412,3 +412,10 @@ pub enum RowStatus {
     SQL_ROW_ERROR = 5,
     SQL_ROW_SUCCESS_WITH_INFO = 6,
 }
+
+#[repr(C)]
+pub struct LibmongosqltranslateDataIO {
```
nbagnard, line 417: This name is a little bit too specific if you consider what it represents.
Could we use something more generic? Like ByteBuffer or ByteArray or RawBson?

bucaojit (changes requested):
I ran one of our integration tests and hit these issues.

core/src/util/test_connection.rs
```diff
@@ -119,34 +119,35 @@ mod test {
...
-            input_text_to_string_w(
-                buffer.as_mut_ptr(),
-                isize::try_from(buffer_len)
-                    .expect("buffer length is too large for {isize::MAX} on this platform"),
-            )
-            .to_lowercase()
-            .contains("unsupported cluster configuration: unknown cluster/target type detected.")
-        });
-    }
+    // TODO: Fix in SQL-2288
```
nbagnard, line 122: I am not sure I understand why this test needs to be commented out?
Same for the other ones below. Except for `test_enterprise_with_library_fails_due_to_missing_sql_get_result_schema_command` which I understand needs the library to be present (although this could be a mock library).

core/src/util/mod.rs
```diff
@@ -24,6 +28,63 @@ lazy_static! {
     static ref NON_ESCAPED_PERCENT: FancyRegex = FancyRegex::new(r"(?<!\\\\)%").unwrap();
 }
 
+pub(crate) fn handle_libmongosqltranslate_command(command: Document) -> Result<Document> {
```
nbagnard, line 31: Please add a detailed comment describing what this does.

nbagnard, line 31: [nit] I think that the function name is not bad, except that because it is the ODBC driver I would rather stay clear of using the word `handle` if possible 😅
```suggestion
pub(crate) fn libmongosqltranslate_runCommand(command: Document) -> Result<Document> {
```

nbagnard (changes requested):
The logic is looking good! 😊
Most comments are for ideas to simplify the code.

core/src/err.rs
```diff
@@ -54,6 +54,24 @@ pub enum Error {
     UnsupportedOperation(&'static str),
     #[error("Statement not executed")]
     StatementNotExecuted,
+    #[error("The driver version `{0}` is incompatible with libmongosqltranslate")]
```
nbagnard, line 57: Could we also list the version of libmongosqltranslate in the error message?

terakilobyte (changes requested):
I think this is a great first pass. I left some specific suggestions.

In general, I think we should consider restructuring this a bit.

* It might be helpful to isolate specific structs and functionality to interfacing with `libmongosqltranslate` into a dedicated `mongosqltranslate` mod within `core`. 
* There is a lot of grabbing specific keys, which results in the current usage of a lot of `expect`s. I think we should be converting these `Document`s to structs and using them that way (serialize, deserialize) rather than working with the raw documents (though it may make sense to use Document as a type for the schema of a collection).

core/src/conn.rs
```diff
@@ -133,6 +138,47 @@ impl MongoConnection {
...
+                "The `version` key is missing from the document returned by libmongosqltranslate.",
+            )
+            .as_str()
+            .expect("The `version` key should map to a `String`")
+            .to_string();
```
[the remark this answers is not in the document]
EthanHardyMongo: I added an error here instead. I think it's better if we have an error instead of attaching an unknown version because the version should always be present.

core/src/query.rs
EthanHardyMongo: I originally tried this multiple ways, but rust doesn't allow it. Here's a patch a ran with the code you're suggesting: https://spruce.mongodb.com/version/66f2fd62a11f4700074e2f7f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

core/src/util/test_connection.rs
EthanHardyMongo: This code is the problem:
```
        load_mongosqltranslate_library();

        let is_libmongosqltranslate_compatible_with_driver_version =
            if get_mongosqltranslate_library().is_some() {
                let libmongosqltranslate_version = Self::get_libmongosqltranslate_version()?;

                // This appends "|libmongosqltranslate+<version>" to the app_name
                user_options
                    .client_options
                    .app_name
                    .as_mut()
                    .expect("app_name unexpectedly has the value `None`")
                    .push_str(
                        &("|libmongosqltranslate+".to_owned() + &libmongosqltranslate_version),
                    );

                let compatibility = Self::is_libmongosqltranslate_compatible_with_driver_version()?;

                Some(compatibility)
            } else {
                None
            };
```
Basically, the mock library gets loaded, so the if-statement is true, then the driver tries to issue real commands to the mock library. This is causing all the failures.

core/src/conn.rs
```diff
@@ -158,13 +198,37 @@ impl MongoConnection {
...
+        let is_libmongosqltranslate_compatible_with_driver_version =
+            if get_mongosqltranslate_library().is_some() {
+                let libmongosqltranslate_version = Self::get_libmongosqltranslate_version()?;
+
+                // This appends "|libmongosqltranslate+<version>" to the app_name
+                user_options
+                    .client_options
+                    .app_name
+                    .as_mut()
+                    .expect("app_name unexpectedly has the value `None`")
```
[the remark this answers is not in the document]
EthanHardyMongo: I used `expect` here because its not possible for `app_name` to be `None` at this point in the code. You can think of this `expect` as a `unreachable!()`. Also, I feel this way about the few `expects` remaining in this PR. I'm not sure we want to error where they are because they indicate that something is very wrong internally if they are ever hit.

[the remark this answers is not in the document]
terakilobyte: Consider what the user will see and what will be logged. It's much better to handle all of these in one place and put an error on it that we could bubble up to inform the user something is horribly wrong and they should report it.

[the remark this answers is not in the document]
EthanHardyMongo: Makes sense; I'll change it to an error. Something I just thought of though: should I change my two instances of `unreachable!()` into an error too? Because I feel like the same logic applies there. I'm guessing in general we would rather not panic?

```
            MongoClusterType::Community | MongoClusterType::UnknownTarget => {
                // On connection, these types should get caught and throw an error.
                unreachable!()
            }
```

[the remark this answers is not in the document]
terakilobyte: Eventually it should probably panic, but in a centralized, controlled way. We can chat on Slack if you'd like about how to do this. If we go with structs for our commands and implement custom serializer and deserializer we can handle it in one spot.

core/src/query.rs
```diff
@@ -40,36 +155,69 @@ impl MongoQuery {
...
+                unreachable!()
+            }
+        };
 
         Ok(Self {
             resultset_cursor: None,
             resultset_metadata: metadata,
             current: None,
-            current_db: Some(current_db),
+            current_db: Some(current_db.clone()),
```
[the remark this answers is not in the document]
EthanHardyMongo: It's fixed now! It was due to shadowing.

core/src/query.rs
```diff
@@ -40,36 +155,69 @@ impl MongoQuery {
...
+            MongoClusterType::Enterprise => {
+                // Get relevant namespaces
+                let namespaces: BTreeSet<Namespace> =
+                    Self::get_sql_query_namespaces(query, current_db)?;
+
+                // Translate sql
+                let mongosql_translation =
+                    Self::translate_sql(query, current_db, namespaces, client, &db)?;
+
+                let translation_metadata = SqlGetSchemaResponse {
```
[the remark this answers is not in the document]
EthanHardyMongo: Is the `Users` struct supposed to be there?

core/src/query.rs
nbagnard: ah yes, because Collection is defined in the block.
Just define it outside. I renamed collection -> collection_name to make it clearer that's what it is.
```
let collection;
let mut aggregate : Aggregate = if let Some(c_name) = collection_name {
    collection = db.collection::<Document>(c_name.as_str());
    collection.aggregate(pipeline).comment(stmt_id)
}
else {
    db.aggregate(pipeline).comment(stmt_id)
};
....
```

That compiles fine for me

core/src/mongosqltranslate_data_types.rs
```diff
@@ -0,0 +1,196 @@
...
+        map.end()
+    }
+}
+
+impl<T> Command<T> {
+    pub fn new(options: T) -> Self {
+        Self { options }
+    }
+}
+
```
terakilobyte, line 35: If we add a `From` impl, the usage becomes a lot nicer. Just a suggestion, it's not a blocker. My other suggestions will be built on this one.

```suggestion
impl<T: CommandName> From<T> for Command<T> {
    fn from(options: T) -> Self {
        Self { options }
    }
}
```

core/src/util/mod.rs
```diff
@@ -24,6 +31,68 @@ lazy_static! {
...
+/// handles serializing it into a BSON byte vector, calls the libmongosqltranslate runCommand,
+/// deserializes the response, and returns either an error or a valid response
+/// for the given `runCommand`.
+pub(crate) fn libmongosqltranslate_run_command<T: CommandName + Serialize>(
+    command: Command<T>,
+) -> Result<CommandResponse> {
+    let library = get_mongosqltranslate_library().ok_or(Error::UnsupportedClusterConfiguration(
+        "Enterprise edition was detected, but libmongosqltranslate was not found.".to_string(),
+    ))?;
+
```
terakilobyte, line 51: If there is a generic From impl...

```suggestion
pub(crate) fn libmongosqltranslate_run_command<T: CommandName + Serialize>(
    command: impl<Into<Command<T>>,
) -> Result<CommandResponse> {
    let command = command.into();
    let library = get_mongosqltranslate_library().ok_or(Error::UnsupportedClusterConfiguration(
        "Enterprise edition was detected, but libmongosqltranslate was not found.".to_string(),
    ))?;

```

core/src/query.rs
```diff
@@ -31,6 +38,87 @@ pub struct MongoQuery {
 }
 
 impl MongoQuery {
+    fn get_sql_query_namespaces(sql_query: &str, db: &String) -> Result<BTreeSet<Namespace>> {
+        let command = Command::new(GetNamespaces::new(sql_query.to_string(), db.to_string()));
+
```
terakilobyte, line 43: And building on my suggestion, we can get rid of these `Command::new` wrappers entirely!

terakilobyte (approved):
🔥  This looks great from a code perspective.

Could we add a new error type for malformed responses and convert to it? Not a blocker.

core/src/err.rs
```diff
@@ -54,6 +54,34 @@ pub enum Error {
...
+    #[error(
+        "The libmongosqltranslate command `{0}` failed. Error message: `{1}`. Error is internal: {2}"
+    )]
+    LibmongosqltranslateCommandFailed(&'static str, String, bool),
+    #[error("Loading the runCommand symbol from libmongosqltranslate failed with error: {0}")]
+    RunCommandSymbolNotFound(String),
+    #[error("Deserializing libmongosqltranslate response to BSON Document failed with error: {0}")]
+    LibmongosqltranslateDeserialization(mongodb::bson::de::Error),
+    #[error("Deserializing libmongosqltranslate response Document to CommandResponse failed with error: {0}")]
+    BsonDocumentToCommandResponseDeserialization(mongodb::bson::de::Error),
```
EthanHardyMongo, line 72: > Could we add a new error type for malformed responses and convert to it?

@terakilobyte Is this the new error you wanted added? It's for the conversion from `Document` to `CommandResponse`, so it will catch any incorrectly formatted `Documents`.

nbagnard (changes requested):
A couple of things can be streamlined further, but this is looking better and better!

core/src/load_library.rs
```diff
@@ -88,8 +82,8 @@ pub fn load_mongosqltranslate_library() {
     });
 }
 
-pub fn get_run_command() -> Result<
-    Symbol<'static, unsafe extern "C" fn(*const u8, usize) -> RunCommandResult>,
+pub fn get_mock_run_command() -> Result<
```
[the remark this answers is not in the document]
EthanHardyMongo: I changed it because it can only be used for the mock runCommand due to this part: `unsafe extern "C" fn(*const u8, usize)`. The other one requires `unsafe extern "C" fn(BsonBuffer)`.

core/src/mongosqltranslate.rs
```diff
@@ -0,0 +1,378 @@
...
+/// deserializes the response, and returns either an error or a valid response
+/// for the given `runCommand`.
+pub(crate) fn libmongosqltranslate_run_command<T: CommandName + Serialize>(
+    command: impl Into<Command<T>>,
+) -> Result<CommandResponse> {
+    let library = get_mongosqltranslate_library().ok_or(Error::UnsupportedClusterConfiguration(
+        "Enterprise edition was detected, but libmongosqltranslate was not found.".to_string(),
+    ))?;
+
+    let run_command_function: Symbol<'static, unsafe extern "C" fn(BsonBuffer) -> BsonBuffer> =
```
nbagnard, line 296: And use `get_run_command` here once the mock is updated.

core/src/mongosqltranslate.rs
```diff
@@ -0,0 +1,378 @@
...
+mod unit {
+    use super::*;
+    use bson::{doc, Document};
+
+    #[test]
+    fn library_load_and_run_command_test() {
+        load_mongosqltranslate_library();
+        assert!(get_mongosqltranslate_library().is_some());
+
+        let run_command = get_mock_run_command().expect("Failed to load runCommand symbol");
```
nbagnard, line 352: And use `get_run_command` here once the mock is updated.

core/src/query.rs
```diff
@@ -40,37 +168,84 @@ impl MongoQuery {
...
+                    db.run_command(get_result_schema_cmd)
+                        .await
+                        .map_err(Error::QueryExecutionFailed)
+                })?;
+                drop(guard);
+                let get_result_schema_response: SqlGetSchemaResponse =
+                    mongodb::bson::from_document(schema_response)
+                        .map_err(Error::QueryDeserialization)?;
+
+                // 2. Run the $sql aggregation to get the result set cursor.
```
nbagnard, line 193:
```suggestion
                // 2. Generate the $sql aggregation pipeline to use at execution time.
```

core/src/util/test_connection.rs
```diff
@@ -120,6 +120,7 @@ mod test {
     }
 
     #[test]
+    #[ignore = "SQL-2288: need real libmongosqltranslate"]
```
nbagnard, line 123: [Note] If it is grabbing the mock library now, it will still later on because it is using cfg!(test).
We'll like have to add mock implementations for the different commands too.

nbagnard (approved):
The remaining comments are not blocking. But if the changes are not done here, they will have to handled in SQL-2288 anyway.

bucaojit (approved):
Great work with all the improvements 🔥

PR #249, opened 2024-10-11
SQL-2396: update evergreen file

jchemburkar:
Note: parsing of libmongotranslate version was done in a preceding pr so both Patrick and I can access it.

Updated all references to mongoodbc.msi and mongoodbc.tar.gz to just use the versioned filename to avoid any confusion.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -854,7 +854,7 @@ functions:
           - ubuntu2204
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: mongosql-odbc-driver/release/mongoodbc.tar.gz
+        local_file: mongosql-odbc-driver/release/${UBUNTU_FILENAME}
         remote_file: mongosql-odbc-driver/artifacts/${version_id}/${build_variant}/release/${UBUNTU_FILENAME}
         bucket: mciuploads
         permissions: public-read
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -2087,6 +2087,8 @@ tasks:
       - func: "trace artifacts"
 
   - name: download-center-update
+    # TODO: remove this before merging eap branch back into main branch, to allow download center to update.
+    disable: true
     git_tag_only: true
     depends_on:
       - name: trace-artifacts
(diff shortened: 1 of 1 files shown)


evergreen.yml
```diff
@@ -68,21 +68,24 @@ functions:
             export RELEASE_VERSION=snapshot
             export LIBMONGOTRANSLATE_VERSION=snapshot
           fi
-          export MSI_FILENAME="mongoodbc.msi"
+          export MSI_FILENAME="mongoodbc-eap-$RELEASE_VERSION-lib-$LIBMONGOTRANSLATE_VERSION.msi"
```
nbagnard, line 71: I'm wondering if need to go this granular on the file name.
It make sense because we can technically have vx.y.z-libva.b.c and vx.y.z-libvd.e.f tags, but at the same time, it is not our intent. An upgrade of mongotranslate means a new version of ODBC is needed to bundle it so we should always have a different ODBC version.
 and for users, it would be much easier to only have to deal with the OBDC driver version. They shouldn't have to worry about the mongotranslate library version since we take care of everything for them.
Our release directory path has the mongotranslate version info so we shouldn't have any mix up problem if we mess up a tag (even if it should not happen).

evergreen.yml
```diff
@@ -2043,6 +2047,7 @@ tasks:
       #   variants: [ ubuntu2204 ]
 
   - name: sign
+    allowed_requesters: ["patch", "github_tag"]
```
nbagnard, line 2050:
```suggestion
    allowed_requesters: ["adhoc", "github_tag"]
```

Patch is for when you run a patch to test something and I think for PR checks.
adhoc is for periodic builds.

nbagnard (changes requested):

evergreen.yml
jchemburkar: ah yes, thanks. I got turned around by the wording in the ticket

evergreen.yml
jchemburkar: That's cool with me too; I'm happy to remove the lib version from the name.

nbagnard (approved):
LGTM! 👍

EthanHardyMongo (approved):
LGTM!

PR #245, opened 2024-10-07
SQL-2374: Implement columns metadata logic for Direct cluster

EthanHardyMongo:
This PR updates `get_next_metadata` in `fields.rs`, so that it works in Direct Cluster mode.

PR #251, opened 2024-10-14
SQL-2288: Create mongosqltranslate integration tests

EthanHardyMongo:
This PR corrects the changes made in [SQL-2291](https://github.com/mongodb/mongo-odbc-driver/pull/242) and [SQL-2374](https://github.com/mongodb/mongo-odbc-driver/pull/245) and adds integration tests. 

The path for libmongosqltranslate will instead be modified in this follow up ticket: https://jira.mongodb.org/browse/SQL-2403


integration_test/tests/odbc_2_tests.rs
```diff
@@ -105,6 +105,8 @@ mod integration {
     /// call SQLGetTypeInfo to verify the correct types are returned. For all types,
     /// we should get both date types back. For date, we should get the specific date type
     /// we expect back.
+    ///
+    /// NOTE: FLAKY TEST
```
EthanHardyMongo, line 109: As I was testing, I noticed sometimes this test would fail, so I'm guessing it's flaky. Has anyone else noticed this?

integration_test/tests/mongosqltranslate_tests.rs
```diff
@@ -0,0 +1,157 @@
...
+        get_sql_diagnostics,
+    };
+    use cstr::WideChar;
+    use definitions::{
+        AttrOdbcVersion, CDataType, HStmt, Handle, HandleType, SQLExecDirectW, SQLExecute,
+        SQLPrepareW, SqlReturn, SQL_NTS,
+    };
+
+    #[test]
+    fn test_srv_style_uri_connection() {
```
EthanHardyMongo, line 16: This test requires libmongosqltranslate, so I moved it here to make testing easier.

constants/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "constants"
-version = "0.0.0"
+version = "2.0.0"
 authors = [
```
EthanHardyMongo, line 4: I changed the version to `2.0.0` to test with because otherwise `CheckDriverVersion` would always return incompatible. Obviously, this can't stay like this, so I was wondering how we test with the library without doing this?

core/src/databases.rs
```diff
@@ -221,7 +221,13 @@ impl MongoDatabases {
             })
             .unwrap()
             .iter()
-            .filter(|&db_name| !db_name.is_empty() && !db_name.eq("admin"))
+            .filter(|&db_name| {
```
bucaojit, line 224: Since this list of databases and filter is in a couple places, can we add a helper function to filter out the databases we don't want?

core/src/fields.rs
```diff
@@ -542,23 +550,66 @@ impl MongoFields {
...
+                                    },
+                                );
+
+                            if let Err(error) = result_set_schema {
+                                // If there is an Error while deserializing the schema, we won't show any columns for it
+                                warnings.push(error);
+                                continue;
+                            }
+
+                            result_set_schema.unwrap()
```
bucaojit, line 608: How about something like this to avoid the unwrap, it has similar behavior:
```
match result_set_schema {
    Ok(schema) => schema,
    Err(error) => {
        warnings.push(error);
        continue;
    }
}
```

integration_test/tests/odbc_2_tests.rs
bucaojit: No I haven't noticed it.  What was the error it was reporting?

integration_test/tests/mongosqltranslate_tests.rs
```diff
@@ -0,0 +1,157 @@
...
+                "test".to_string(),
+            ))),
+        );
+
+        unsafe {
+            let mut query: Vec<WideChar> = cstr::to_widechar_vec("select * from foo");
+            query.push(0);
+
+            assert_eq!(
+                SqlReturn::ERROR,
```
bucaojit, line 147: It would be good to add a check for what error we're getting here.  If this one is `SchemaDocumentNotFoundInSchemaCollection` then it would be something like:
```
        let error_message = get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, stmt as Handle);
        assert!(
            error_message.contains("The following collection(s) were not found in the `__sql_schemas` collection"),
            "Expected error message about missing schema collection, got: {}",
            error_message
        );
        ```

File: integration_test/tests/mongosqltranslate_tests.rs [hunk 7]
```diff
@@ -0,0 +1,157 @@
...
+            ))),
+        );
+
+        unsafe {
+            let mut query: Vec<WideChar> =
+                cstr::to_widechar_vec("select * from non_existent_collection");
+            query.push(0);
+
+            assert_eq!(
+                SqlReturn::ERROR,
```
bucaojit, line 122: Same comment here to confirm what error we get.

core/src/fields.rs
```diff
@@ -542,23 +550,66 @@ impl MongoFields {
...
+                            == MongoClusterType::AtlasDataFederation
+                        {
+                            let get_schema_cmd = doc! {"sqlGetSchema": &collection_name};
+
+                            let sql_get_schema_response: Result<SqlGetSchemaResponse> =
+                                mongodb::bson::from_document(
+                                    db.run_command(get_schema_cmd).await.unwrap(),
+                                )
+                                .map_err(|e| {
+                                    Error::CollectionDeserialization(collection_name.clone(), e)
```
bucaojit, line 567: How about an integration test that confirms we get this on a deserialization error?

integration_test/tests/mongosqltranslate_tests.rs
```diff
@@ -0,0 +1,157 @@
...
+                SQLPrepareW(stmt as HStmt, query.as_ptr(), SQL_NTS),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, stmt as Handle)
+            );
+
+            disconnect_and_close_handles(dbc, stmt);
+        }
+        let _ = unsafe { Box::from_raw(env_handle) };
+    }
+}
```
bucaojit, line 157: One test I see in the design document that I don't see implemented yet is `Invalid Command`.

bucaojit (changes requested):
Lots of good work.  I had some suggestions.

integration_test/tests/mongosqltranslate_tests.rs
EthanHardyMongo: The `Invalid Command` tests ended up being unit tests in `mongosqltranslate`. Since we control the API and the driver's calls to the API, I don't think it's possible to have an integration test that has an invalid command.

core/src/col_metadata.rs
```diff
@@ -171,9 +171,12 @@ pub struct VersionedJsonSchema {
...
 
 // Struct representing the ResultSetSchema.
+// The `schema` field needs the alias `result_set_schema` because this struct is used to get the schema
+// from the __sql_schemas collection, which stores the schema in it's `schema` field, and the libmongosqltranslate
+// `translate` command, which stores the schema in it's `result_set_schema` field.
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone, Default)]
 pub struct ResultSetSchema {
-    #[serde(rename = "result_set_schema")]
+    #[serde(alias = "result_set_schema")]
     pub schema: crate::json_schema::Schema,
```
EthanHardyMongo, line 180: I'm not sure why this wasn't failing earlier, but I fixed it and added a comment explaining what's going on.

EthanHardyMongo: Oh actually I think it's because the error it would cause gets pushed into a `warnings` vector instead of causing a crash. This is probably why my tests weren't failing.

```
let result_set_schema: Result<ResultSetSchema> =
                                ResultSetSchema::from_sql_schemas_document(&schema_doc).map_err(
                                    |e| {
                                        Error::CollectionDeserialization(collection_name.clone(), e)
                                    },
                                );

                            match result_set_schema {
                                Ok(result_set_schema) => result_set_schema,
                                Err(error) => {
                                    // If there is an Error while deserializing the schema, we won't show any columns for it
                                    warnings.push(error);
                                    continue;
                                }
                            }
```

EthanHardyMongo: Oh wait never mind. The part of the code that would fail was never called by my tests. That's why this wasn't causing any failures.

integration_test/tests/odbc_2_tests.rs
bucaojit: I see, I looked through the logs and didn't find any other useful error messages to indicate the problem.  I don't see this failing in the waterfall runs so it could be new or specific to your runs.  You can create an investigation ticket to look into it further, although it doesn't have to be high priority unless we see it happen again.

EthanHardyMongo: Yeah makes sense. Considering it only happened in these instances, I'm guessing it was something to do with my patches (maybe I modified the code in a weird way). Anyways, I'll remove the comment. Thanks for looking into it! I'll keep an eye on it, and if it happens again, I'll make an investigation ticket.

integration_test/tests/mongosqltranslate_tests.rs
```diff
@@ -0,0 +1,171 @@
...
+            query.push(0);
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
+                SQLPrepareW(stmt as HStmt, query.as_ptr(), SQL_NTS),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, stmt as Handle)
+            );
+
+            get_column_attributes(stmt as Handle, 5);
```
bucaojit, line 52: Since we're not running the result_set_tests on mongosqltranslate, we don't check the actual outputs only that the counts match.  Can you see if here we can add a check that we are getting the actual column values we expect?  

That would help confirm we're getting the correct values from the change made in getting the schema information.  Let me know if you think that would be helpful.  In the result set test we do it [here](https://github.com/mongodb/mongo-odbc-driver/blob/52be409a03589324e1401501925b1e3c86ced764/integration_test/src/test_runner/mod.rs#L580).

core/src/query.rs
```diff
@@ -59,90 +60,113 @@ impl MongoQuery {
...
+        let collection_names: Vec<String> = client.runtime.block_on(async {
+            db.list_collection_names()
+                .await
+                .map_err(Error::QueryExecutionFailed)
+        })?;
+
+        let sql_schemas_collection_exists =
+            collection_names.contains(&SQL_SCHEMAS_COLLECTION.to_string());
+
+        if !sql_schemas_collection_exists {
```
nbagnard, line 72: That's not quite true. If we pass in an empty schema, it will work "sometimes". For example, I think that `select a,b,c from mytable` works even without schema.

integration_test/tests/mongosqltranslate_tests.rs
```diff
@@ -0,0 +1,216 @@
...
+
+            assert_eq!(
+                SqlReturn::ERROR,
+                SQLPrepareW(stmt as HStmt, query.as_ptr(), SQL_NTS),
+                "{}",
+                get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, stmt as Handle)
+            );
+
+            let error_message = get_sql_diagnostics(HandleType::SQL_HANDLE_STMT, stmt as Handle);
+            assert!(
```
nbagnard, line 130: It looks like this is more for testing the behaviour for collections with no schema rather than an invalid SQL query which would failed parsing or algebrizing.
Also, it should to fail later if we send empty schemas for collections with no schemas found, no?.

nbagnard (changes requested):

core/src/query.rs
EthanHardyMongo: Yeah that's why I wrote "so **most** queries will fail.", which I think is correct? What would you like me to change it to?

core/src/fields.rs
```diff
@@ -542,23 +544,71 @@ impl MongoFields {
...
+                                    }
+                                };
+
+                            error_checked_sql_get_schema_response.into()
+                        } else if mongo_connection.cluster_type == MongoClusterType::Enterprise {
+                            let schema_collection =
+                                db.collection::<Document>(SQL_SCHEMAS_COLLECTION);
+
+                            let schema_doc: Document = mongo_connection
+                                .runtime
```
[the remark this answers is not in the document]
EthanHardyMongo: I guess this part in `query.rs` needs to change too.
```
            if namespaces.len() != collections_schema_doc.len() {
                let missing_collections: Vec<String> = namespaces
                    .iter()
                    .map(|namespace| namespace.collection.clone())
                    .filter(|collection| !collections_schema_doc.contains_key(collection.as_str()))
                    .collect();

                return Err(Error::SchemaDocumentNotFoundInSchemaCollection(
                    missing_collections,
                ));
            }
```
I guess in general we shouldn't be returning errors for missing schema docs, but instead, should create an empty schema.

core/src/query.rs
nbagnard: ah, I see what you meant now. Maybe we can be a bit more direct and not confuse users with the schema collection name.
What about "There are no schema information in database {0}. SQL capabilities will be very limited. Please make sure to add schema information before querying" or maybe "There are no schema information in database {0}. SQL capabilities will be very limited. Please make sure to add schema information before using the driver"?

core/src/fields.rs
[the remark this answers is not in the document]
nbagnard: Yes exactly, empty schema should be used when no schema are available :)

core/src/mongosqltranslate.rs
```diff
@@ -63,12 +63,12 @@ fn get_mock_library_path() -> PathBuf {
...
 // It is stored in a static variable to ensure that it is only loaded once.
-pub fn load_mongosqltranslate_library() {
+pub fn load_mongosqltranslate_library(mock_library: bool) {
     INIT.call_once(|| {
-        let library_path = if cfg!(test) {
+        let library_path = if mock_library {
             get_mock_library_path()
         } else {
-            get_library_path(LIBRARY_NAME)
+            get_library_path()
```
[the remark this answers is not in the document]
EthanHardyMongo: The problem is that we use `cargo test` for all of our tests, and there's other tests that need the real library such as the `bad_host()` test. On the other hand, only one test needs the mock library and this doesn't seem likely to change. I guess we could also make a feature flag for the one test that needs the mock library: `library_load_and_run_command_test()`.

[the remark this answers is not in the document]
EthanHardyMongo: Actually, I wrote that wrong. Its more like we have to make sure the `bad_host()` test doesn't have a library available, but this will never work if it just defaults to the mock library.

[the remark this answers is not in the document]
EthanHardyMongo: What I'm trying to get at is that the mock library isn't really a mock library that can be used in the absence of the real one. It's more like a dummy library that we can use to test for loading and calling `run_command`, but nothing else. So it's really only useful for that one test.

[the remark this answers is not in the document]
nbagnard: If you need a different library then it would mean that this is an integration test. Their are already cluster type tests under integration tests and they seem well suited for testing cluster types.
I looked into the history of that test.
It used to be for testing that if the uri was not pointing to a mongodb host, then we would fail with "server selection timeout".
I think the issue is coming from the change in cluster_type.rs.
In `pub async fn determine_cluster_type(client: &Client) -> MongoClusterType {`, if we can't get build info back we assume it's because we don't know the type of cluster and we swallow the error.
```
    let cmd_res: Document = match db.run_command(build_info_cmd).await {
        Ok(res) => res,
        Err(e) => {
            log::error!("Failed to run buildInfo command: {:?}", e);
            return MongoClusterType::UnknownTarget;
        }
    };
```
We should return `Result<MongoClusterType>` and bubble up the error.
This would fix the test (you will need to revert the change made in PR #240 for the error message) and let you only use the mock library when running unit tests I believe.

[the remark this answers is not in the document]
EthanHardyMongo: I think the issue is here:
```
        load_mongosqltranslate_library();

        let (is_libmongosqltranslate_compatible_with_driver_version, libmongosqltranslate_version) =
            if get_mongosqltranslate_library().is_some() {
                let libmongosqltranslate_version = Self::get_libmongosqltranslate_version()?;

                // This appends "|libmongosqltranslate+<version>" to the app_name
                user_options
                    .client_options
                    .app_name
                    .as_mut()
                    .ok_or(Error::EmptyAppName)?
                    .push_str(
                        &("|libmongosqltranslate+".to_owned() + &libmongosqltranslate_version),
                    );

                let compatibility = Self::is_libmongosqltranslate_compatible_with_driver_version()?;

                log::info!("libmongosqltranslate version: `{0}`. ODBC driver version: `{1}`. The ODBC driver and libmongosqltranslate library is compatible: `{2}`.", libmongosqltranslate_version, *DRIVER_METRICS_VERSION, compatibility);

                (Some(compatibility), Some(libmongosqltranslate_version))
            } else {
                (None, None)
            };
```
The mock library can't handle the runCommands that occur before we know the cluster type.

core/src/fields.rs
EthanHardyMongo: To get this error, there would have to be a document in the `__sql_schemas` collection that has the `schema` field set to either the wrong type (an int) or an incorrectly made schema that can't be correctly deserialized. Are you sure we need an integration test for this? It's a very specific case. I think maybe a unit test would work better?

integration_test/tests/mongosqltranslate_tests.rs
EthanHardyMongo: I changed the SQL query to `SELECT bed_type.type FROM listingsAndReviews`. 

`bed_type` is a String, so it fails with this error: 
`[MongoDB][Core] The libmongosqltranslate command 'translate' failed. Error message: algebrize error: Error 1002: Incorrect argument type for 'FieldAccess'. Required: object type. Found: string.`.

I think it's fine now?

nbagnard: Yes, if it is what you get that looks good. I don't see the change on the PR yet though ;)

core/src/fields.rs
bucaojit: I see. A unit test would be fine then.

core/src/query.rs
```diff
@@ -59,90 +60,118 @@ impl MongoQuery {
...
-                .try_collect::<Vec<Document>>()
-                .await
-                .map_err(Error::QueryExecutionFailed)
-        })?;
+            if schema_catalog_doc_vec.len() > 1 {
+                return Err(Error::MultipleSchemaDocumentsReturned(
+                    schema_catalog_doc_vec.len(),
+                ));
+            } else if schema_catalog_doc_vec.is_empty() {
+                return Err(Error::SchemaCollExistsButNoSchemaInformationWasReturned(
```
nbagnard, line 143: I am not sure this case is ever hit. And if it is, shouldn't we assign empty schemas to all the collections in the namespace?

evergreen.yml
```diff
@@ -1409,68 +1408,92 @@ functions:
...
+        local_file: mongosql-odbc-driver/libmongosqltranslate.so
+        bucket: translators-connectors-releases
+    - command: shell.exec
+      type: test
+      params:
+        shell: bash
+        working_dir: mongosql-odbc-driver
+        script: |
+          ${prepare_shell}
+          sudo mkdir -p /opt/mongodb/atlas-sql-odbc-driver/
```
nbagnard, line 1434: Now that Nathan's PR is in, you should be able to change the logic to get the latest snapshot /release version.

nbagnard (commented):
There is just that last thing with `SchemaCollExistsButNoSchemaInformationWasReturned` which seems odd, but otherwise it all looks good I think! Can you walked me through the logic. I might have just missed something.

bucaojit (approved):
LGTM!  Thanks for addressing my comments.

core/src/query.rs
EthanHardyMongo: This case is hit if you doing something like `SELECT * FROM <nonexistent_collection>`. Basically, the aggregation pipeline returns nothing, so then the `schema_catalog_doc_vec` is empty. This happens if you have a `__sql_schemas` collection in the DB and namespaces in the query, but none of the namespaces have an entry in the `__sql_schemas` collection. At first, I thought this was such a bizarre case that we should return an error, but I saw what you did the JDBC driver if this case is hit, and I think it's a better approach (i.e, put a warning in the logs and generate empty schemas). I'll modify the code to do that instead 👍.

evergreen.yml
EthanHardyMongo: I copied what was done in the JDBC Driver.

EthanHardyMongo: never mind. I didn't see that patrick already did that lol

core/src/query.rs
```diff
@@ -59,90 +60,134 @@ impl MongoQuery {
...
+            })?;
+
+            if schema_catalog_doc_vec.len() > 1 {
+                return Err(Error::MultipleSchemaDocumentsReturned(
+                    schema_catalog_doc_vec.len(),
+                ));
+            } else if schema_catalog_doc_vec.is_empty() {
+                log::warn!("There is a `__sql_schemas` collection in database `{0}`; however, no schema information was returned for the requested collections.\
+                This means that all the requested collections either don't exists in `{0}` or don't have an entry in the `__sql_schemas` collection. For now, they will\
+                be assigned empty schemas. Hint: You either need to generate schemas for your collections or correct your query.", current_db);
```
nbagnard, line 146: Use a similar warning message to the one below. I would like to avoid putting too much technical details in the message.

```suggestion
                log::warn!("No schema information found for the collections {0} in database `{1}`.\
                Either the collections don't exists in `{1}` or they don't have a schema. Hint: You either need to generate schema for your collections or correct your query.", collection_names, current_db);
```

evergreen.yml
nbagnard: ah, yes, I see the S3 get is above 👍 
It might be that Oliver's changes will break your integration tests because the driver will not look into that directory anymore.

nbagnard (approved):
LGTM! 
Can you just quickly simplify the error message when there are no schemas returned at all and then you can merge :)
Great work on that PR 👍

PR #255, opened 2024-10-25
SQL-2288: Fix integration tests

EthanHardyMongo:
My tests were failing in the periodic builds because they are flaky, so I added an `ORDER BY` clause to the queries to ensure that the right result set is always used for testing.


nbagnard (approved):
LGTM!

PR #257, opened 2024-10-28
SQL-2422: replace get_adf_version

terakilobyte:
Switching out `get_adf_version` for `get_server_version`. Datalake is now optional so it's universal. If for some reason the command fails it'll be automatically handled.


jchemburkar (approved):
lgtm

EthanHardyMongo (approved):
LGTM!

PR #264, opened 2024-11-12
SQL-2466: handle changing databases for power bi on prem

jchemburkar:
I struggled to find a clean way to get a list of databases to the schema pulling, so rather, I just used the client to pull all databases at the time of querying. The reason it was an issue is that we create new HDbc handles, and MongoConnections, frequently enough that the only way to persist the databases from the initial listing was to push all the way down to the env handle (which didn't seem right). I'm very open to alternative implementations if there's an easier way!

I also needed to update the db set on the MongoQuery returned from mongoprepare in order to get the execute function working properly.

This seems to work in power bi, though would love a second set of eyes in case I missed anything!

### DIFF
--- a/core/src/query.rs
+++ b/core/src/query.rs
@@ -8,7 +8,6 @@ use crate::{
         TranslateCommandResponse,
     },
     stmt::MongoStatement,
-    util::DISALLOWED_DB_NAMES,
     Error, TypeMode,
 };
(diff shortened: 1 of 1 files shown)


terakilobyte (approved):
LGTM. Nice work

EthanHardyMongo (approved):
LGTM!

PR #265, opened 2024-11-12
Add rust toolchain

terakilobyte:
The rust toolchain needed to be added. The integration test issue was due to someone writing to the source database, and happening to write to the exact collection we expect not to exists.

https://spruce.mongodb.com/version/6734d908150c5700079b6c79/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -2026,6 +2026,7 @@ tasks:
     #   - name: sbom
     #     variant: code-quality-security
     commands:
+      - func: "install rust toolchain"
       - func: "install iODBC"
         variants: [macos, macos-arm]
       - func: "install rust toolchain"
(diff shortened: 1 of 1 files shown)


evergreen.yml
```diff
@@ -2026,6 +2026,7 @@ tasks:
     #   - name: sbom
     #     variant: code-quality-security
     commands:
+      - func: "install rust toolchain"
```
nbagnard, line 2029: Now I see `install rust toolchain` being there twice. I'm thinking that the issue might have been that `install rust toolchain` was restricted (so to speak) to the variants defined right under it and the integration tests are running on another.

terakilobyte: well, it was used in the sbom stuff. This all emerges from how the ssdlc tasks have really intertwined themselves almost everywhere. When this merged back and we do ssdlc stuff on it, it won't need to be there.

nbagnard: The sbom task installs Rust too, so it's not related.
The issue is that the release variants is now trying to run the compile task and this task isn't successful because compile is only supposed to run on variants we build the library for. `eap-release` and `release` are not such variants.
`
 [2024/11/12 10:31:45.264] Command 'shell.exec' in function 'set and check packages version' (step 4 of 14) failed: shell script encountered problem: exit code 127.
 `
 
 I think the issue was introduce with https://github.com/mongodb/mongo-odbc-driver/commit/4d537d8e0da6d47c0d1048ccd158c996116ff4a7.
 Can you please double check why the `compile` task was added?
 Removing it and removing this change should be all you need.

nbagnard (changes requested):

PR #269, opened 2024-11-18
SQL-2474: support x509 connections in ODBC

terakilobyte:
I opted to get rid of the username in x509 as their username isn't important to operations afaict. I also slightly modified the code so that if we have to do anything special with any of the other auth mechanisms it's pretty apparent where that work needs to happen.

I tested this with PowerBI on my windows desktop and was able to make a connection to the M0 test cluster.

### DIFF
--- a/core/src/odbc_uri.rs
+++ b/core/src/odbc_uri.rs
@@ -335,26 +335,37 @@ impl ODBCUri {
             .map_err(Error::InvalidClientOptions)?
             .uuid_representation;
 
-        if let Some(AuthMechanism::MongoDbOidc) = client_options
+        match client_options
             .credential
             .as_ref()
             .unwrap()
             .mechanism
             .as_ref()
         {
-            use futures::future::FutureExt;
-            let cred = client_options.credential.as_mut().unwrap();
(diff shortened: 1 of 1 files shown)


bucaojit (approved):
LGTM.  Nice work getting it working and testing the change 👍

core/src/odbc_uri.rs
```diff
@@ -335,26 +335,38 @@ impl ODBCUri {
...
+                client_options.credential.as_mut().unwrap().password = None;
+            }
+            Some(AuthMechanism::MongoDbOidc) => {
+                use futures::future::FutureExt;
+                let cred = client_options.credential.as_mut().unwrap();
+                cred.oidc_callback = mongodb::options::oidc::Callback::human(move |c| {
+                    async move { crate::oidc_auth::oidc_call_back(c).await }.boxed()
+                });
+                // Unset the password and username if they are empty strings.
+                // This is to accommodate tools like Power BI that require adding empty username and password fields.
```
terakilobyte, line 356: @pmeredit I modified your comment slightly.

pmeredit (approved):

PR #272, opened 2024-11-21
Host switch

terakilobyte:
It's been 3 days since we've had a successful periodic build, though we're assured they'll run at some point. Proposing switching to more available hosts.


nbagnard (approved):
LGTM

bucaojit (approved):

PR #275, opened 2024-11-26
SQL-2486: inject username and password into uri for parsing if required

terakilobyte:
Overall the changes are relatively small. A couple new helper functions to make it easy. If there is no authMechanism specified, or if the authMechanism is one of the Scrams or PLAIN, then username and password are injected if not there already. Otherwise, it's fine to send along for parsing by the driver.

### DIFF
--- a/core/src/odbc_uri.rs
+++ b/core/src/odbc_uri.rs
@@ -11,6 +11,7 @@ use mongodb::{
 use regex::{Regex, RegexBuilder, RegexSet, RegexSetBuilder};
 use shared_sql_utils::Dsn;
 use std::collections::HashMap;
+use std::str::FromStr;
 
 const EMPTY_URI_ERROR: &str = "URI must not be empty";
 const INVALID_ATTR_FORMAT_ERROR: &str = "all URI attributes must be of the form keyword=value";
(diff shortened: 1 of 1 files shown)


core/src/odbc_uri.rs
```diff
@@ -947,20 +1110,41 @@ mod unit {
...
+        }
+
         #[tokio::test(flavor = "current_thread")]
         async fn auth_source_correctness() {
             use crate::odbc_uri::ODBCUri;
-            for (source, uri) in [
+            for (expected, uri) in [
                 (Some("authDB".to_string()), "URI=mongodb://localhost/?authSource=authDB;UID=foo;PWD=bar"),
-                (None, "URI=mongodb://localhost/;UID=foo;PWD=bar"),
+                (Some("admin".to_string()), "URI=mongodb://localhost/;UID=foo;PWD=bar"),
```
terakilobyte, line 1138: Ended up being more correct, since the default authSource is admin if it's not specified elsewhere!

mattChiaravalloti (approved):
Looks good 👍 

It's really a shame the driver does connection string semantic validation at parse time. We duplicated so much of their parse work here just to do this. In an ideal world, we'd be able to parse into a `ConnectionString` struct and do these manipulations in a structured way using an api, rather than all this regex-matching and string-formatting magic 😞 Well done though, this is a great workaround to a restrictive api!

nbagnard (approved):
LGTM!

PR #278, opened 2024-11-27
do not modify uri if no mechanism explicitly set

terakilobyte:
After looking at Natacha's PR for Java and speaking with her, I realized I should not be modifying URIs that do not have an authMechanism explicitly set. This will work for implicit SCRAM as well as authless connections.


nbagnard (approved):
LGTM!

bucaojit (approved):
LGTM

PR #279, opened 2024-12-03
SQL-2501: Implement additional column attributes

jchemburkar:
I tested this with Microstrategy using the instructions in the doc. It seems to work as expected, or at least, I do not see any of the previously mentioned errors in the logs.

I did not test this wiht Qlik View because the doc does not contain instructions to set it up, and there seems to be other blockers for that tool as well.

### DIFF
--- a/odbc/src/api/connect_attr_tests.rs
+++ b/odbc/src/api/connect_attr_tests.rs
@@ -238,7 +238,7 @@ mod unit {
         }
     }
 
-    const UNSUPPORTED_ATTRS: [ConnectionAttribute; 17] = [
+    const UNSUPPORTED_ATTRS: [ConnectionAttribute; 16] = [
         ConnectionAttribute::SQL_ATTR_ASYNC_ENABLE,
         ConnectionAttribute::SQL_ATTR_ACCESS_MODE,
         ConnectionAttribute::SQL_ATTR_TRACE,
@@ -247,7 +247,6 @@ mod unit {
         ConnectionAttribute::SQL_ATTR_TRANSLATE_OPTION,
         ConnectionAttribute::SQL_ATTR_TXN_ISOLATION,
         ConnectionAttribute::SQL_ATTR_ODBC_CURSORS,
-        ConnectionAttribute::SQL_ATTR_QUIET_MODE,
         ConnectionAttribute::SQL_ATTR_PACKET_SIZE,
         ConnectionAttribute::SQL_ATTR_DISCONNECT_BEHAVIOR,
         ConnectionAttribute::SQL_ATTR_ASYNC_DBC_FUNCTIONS_ENABLE,
--- a/odbc/src/api/connect_attr_tests.rs
+++ b/odbc/src/api/connect_attr_tests.rs
@@ -12,7 +12,7 @@ mod unit {
         SQLGetConnectAttrW, SQLSetConnectAttrW,
     };
     use cstr::input_text_to_string_w;
-    use definitions::{ConnectionAttribute, Integer, Pointer, SqlReturn, UInteger};
+    use definitions::{Autocommit, ConnectionAttribute, Integer, Pointer, SqlReturn, UInteger};
     use std::sync::RwLock;
 
     mod get {
@@ -154,6 +154,15 @@ mod unit {
             expected_value = 1u32,
             actual_value_modifier = modify_numeric_attr,
         );
+
+        test_get_attr!(
+            autocommit,
+            attribute = ConnectionAttribute::SQL_ATTR_AUTOCOMMIT as i32,
+            expected_sql_return = SqlReturn::SUCCESS,
+            expected_length = std::mem::size_of::<u32>() as i32,
+            expected_value = 1u32,
+            actual_value_modifier = modify_numeric_attr,
+        );
     }
 
     // Test setting LoginTimeout attribute.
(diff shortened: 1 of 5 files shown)


odbc/src/api/connect_attr_tests.rs
```diff
@@ -205,17 +214,38 @@ mod unit {
...
+                    0
+                )
+            );
+            let conn_handle = (*mongo_handle).as_connection().unwrap();
+            let attributes = &conn_handle.attributes.read().unwrap();
+            assert_eq!(attributes.autocommit, Autocommit::Off);
+        }
+    }
+
+    const UNSUPPORTED_ATTRS: [ConnectionAttribute; 15] = [
```
nbagnard, line 241: Matt's JIRA ticket implies that when setting SQL_QUIET_MODE to a value, the driver would panic. This is odd. 
I didn't see a `set_unsupported_attr` test when I looked at the code previously, but it looks like this is what `set_invalid_attr` tests and that test is passing fine.

Could you please:
- Verify that the driver doesn't panic but report SQL_ERROR if you try to set for example `SQL_ATTR_TRACE` to `SQL_OPT_TRACE_ON`
- Rename the test `set_invalid_attr` -> `set_unsupported_attr`
- Add a test with an incorrect attribute identifier to also make sure that we report SQL_ERROR and don't crash (I'm pretty sure we don't, but I could see a validation)

odbc/src/api/functions.rs
```diff
@@ -1272,14 +1259,13 @@ pub unsafe extern "C" fn SQLDriversW(
...
-#[named]
 #[no_mangle]
 pub unsafe extern "C" fn SQLEndTran(
     _handle_type: HandleType,
-    handle: Handle,
+    _handle: Handle,
     _completion_type: SmallInt,
 ) -> SqlReturn {
-    unimpl!(handle);
+    SqlReturn::SUCCESS
```
nbagnard, line 1268: Can you add a short comment explaining that even though we report that we don't support transactions, some tools still try to use them, and because we only offer read-only access, having SQLEndTran be a no-op is safe.

odbc/src/api/functions.rs
```diff
@@ -3557,7 +3547,8 @@ unsafe fn set_connect_attrw_helper(
                 conn.attributes.write().unwrap().login_timeout = Some(value_ptr as u32);
                 SqlReturn::SUCCESS
             }
-            ConnectionAttribute::SQL_ATTR_APP_WCHAR_TYPE => SqlReturn::SUCCESS,
+            ConnectionAttribute::SQL_ATTR_APP_WCHAR_TYPE
+            | ConnectionAttribute::SQL_ATTR_QUIET_MODE => SqlReturn::SUCCESS,
```
nbagnard, line 3551: uhm, I don't really know why someone would do that, but won't we return SQL_ERROR if an application calls SQLGetConnAttr(SQL_ATTR_QUIET_MODE) while we still let them set it to whatever?
To be safe, can we default to Null (quiet) and store the value passed-in here so that we return whatever was set/the default when SQLGetConnAttr is called.

odbc/src/api/connect_attr_tests.rs
nbagnard, line 1: Can you add a test to set and get `SQL_ATTR_ACCESS_MODE` and `SQL_ATTR_QUIET_MODE`.
This way we are covered.
For `SQL_ATTR_ACCESS_MODE`, checking that the logic to fall back to SQL_MODE_READ works would be nice too.
Something like what we do for `SQL_ATTR_CONNECTION_POOLING` in `get_set_env_attr`

nbagnard (changes requested):
I only a few small requests, but overall it's looking good already!

odbc/src/api/functions.rs
```diff
@@ -1182,22 +1182,9 @@ pub unsafe extern "C" fn SQLDriverConnectW(
...
-                                "{driver_completion:?}"
-                            ))
-                        );
-                        return SqlReturn::ERROR;
-                    }
-                    DriverConnectOption::SQL_DRIVER_NO_PROMPT => {}
-                },
+            // We will treat any valid option passed for driver complete as a no-op
+            match <DriverConnectOption as FromPrimitive>::from_u16(driver_completion) {
+                Some(_) => {}
```
EthanHardyMongo, line 1187:
```suggestion
            // We will treat any valid option passed for DriverCompletion as a no-op
            match <DriverConnectOption as FromPrimitive>::from_u16(driver_completion) {
                Some(_) => {}
```
[nit] I think this makes it a little clearer.

odbc/src/api/connect_attr_tests.rs
EthanHardyMongo: > Matt's JIRA ticket implies that when setting SQL_QUIET_MODE to a value, the driver would panic. This is odd.
I didn't see a set_unsupported_attr test when I looked at the code previously, but it looks like this is what set_invalid_attr tests and that test is passing fine.

Isn't this because Jon added this:
```
ConnectionAttribute::SQL_ATTR_APP_WCHAR_TYPE
| ConnectionAttribute::SQL_ATTR_QUIET_MODE => SqlReturn::SUCCESS,
```

nbagnard: That was done after the ticket was created, so I don't think so, but I can't tell for sure 🤷‍♀️

odbc/src/api/functions.rs
```diff
@@ -2206,6 +2197,7 @@ macro_rules! sql_get_diag_rec_impl {
                     return SqlReturn::ERROR;
                 }
                 let mongo_handle = $handle.cast::<MongoHandle>();
+                println!("here");
                 let odbc_version = (*mongo_handle).get_odbc_version();
```
EthanHardyMongo, line 2201:
```suggestion
                let mongo_handle = $handle.cast::<MongoHandle>();
                let odbc_version = (*mongo_handle).get_odbc_version();
```
[nit]

EthanHardyMongo (approved):
LGTM!

odbc/tests/connection_tests.rs
```diff
@@ -121,31 +121,4 @@ mod integration {
         expected_error_message =
             "[MongoDB][API] Missing property \"Driver\" or \"DSN\" in connection string"
     );
-    test_connection_diagnostics!(
```
nbagnard, line 124: Can we keep these tests and update the expected SQL return to be SQL_SUCESS, for completeness.

nbagnard (approved):
LGTM!
I would keep the tests where we set the driver values for Driver completion, just so that it is clear that's we expect. But, considering that we just do nothing with the value, I'm not too concerned.

PR #280, opened 2024-12-04
SQL-2502: port from other PR

terakilobyte:
This addresses the following JIRAs in the main branch, and is a carbon copy of the fixes introduced in PR #276 

- SQL-2496: handle null pointer parameters in SQLTablesW
- SQL-2497 handle null pointer parameters in SQLColumnsW
- SQL-2500: treat emptry string as no filter in table_type_filter_to_vec
- SQL-2505: address clippy warnings/errors that are causing the waterfall to not be green

### DIFF
--- a/core/src/util/mod.rs
+++ b/core/src/util/mod.rs
@@ -54,8 +54,8 @@ pub(crate) fn is_match(name: &str, filter: &str, accept_search_patterns: bool) -
 
 // Create the list of Collection types to filter on
 pub(crate) fn table_type_filter_to_vec(table_type: &str) -> Option<Vec<CollectionType>> {
-    return match table_type {
-        SQL_ALL_TABLE_TYPES => None,
+    match table_type {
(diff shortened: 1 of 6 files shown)


EthanHardyMongo (approved):
LGTM!

core/src/util/mod.rs
```diff
@@ -102,63 +102,92 @@ macro_rules! set {
...
-        assert!(!is_match("customerssales", "customer_sales", false));
-        assert!(!is_match("conversions2022", "conversions%", false));
-        assert!(!is_match("integration_test", "%test", false));
-        assert!(!is_match("integration_test", "integrationstest", false));
-    }
+    mod is_match {
+        use super::is_match;
+
+        #[test]
+        fn test_is_positive_match_literal() {
```
pmeredit, line 143: huh, usually functions like this put pattern before value to search, but I guess it's too late to change this :joy:

pmeredit (approved):
LGTM!

PR #287, opened 2024-12-09
SQL-2498: Fix schema fetching in SQLColumnsW

EthanHardyMongo:
In order to get schemas, you have to use a `$match` because `find()` and `find_one()` do not work for collections that start with an underscore. This is why it previously wasn't working. Additionally, the empty schemas that we default to must contain `bsonType: "object"` to get converted to a simplified::Schema properly, so I added that too in this PR. 

I tested this PR out on QlikView and it worked; the schemas are successfully retrieved by SQLColumnsW!


terakilobyte (approved):
LGTM. Before merging, let's capture this in an integration test to ensure we don't have future regressions.

core/src/fields.rs
```diff
@@ -576,20 +577,40 @@ impl MongoFields {
...
                                 .await
                                 .map_err(Error::QueryExecutionFailed)?
-                                .unwrap_or({
-                                    log::warn!("No schema was found for collection `{}`. It will be assigned\
-                                    an empty schema. Hint: Generate schemas for your collections.", collection_name);
+                                .try_collect::<Vec<Document>>()
+                                .await
+                                .map_err(Error::QueryExecutionFailed)?;
+
+                            if schema_coll_agg_response.len() > 1 {
```
nbagnard, line 594:
```suggestion
                            if schema_coll_agg_response.len() > =1 {
```
There could be only one, no?

nbagnard (approved):
LGTM

PR #291, opened 2025-01-06
SQL-2527: Handle possible null string_length_2 parameter

bucaojit:
Add null checking for `string_length_2` in the `SQLDriverConnectW()` function. 

This allowed the `iusql -v MongoDB_Atlas_SQL` command on Ubuntu to succeed. 

```
$ iusql -v MongoDB_Atlas_SQL
+---------------------------------------+
| Connected!                            |
|                                       |
| sql-statement                         |
| help [tablename]                      |
| quit                                  |
|                                       |
+---------------------------------------+
SQL> 
```

Added a test to confirm that the connection works with the `string_length_2` parameter is null and confirmed the test fails without the change.


terakilobyte (approved):
LGTM. Nice job on the test updates too.

mattChiaravalloti (approved):
lgtm 💯

PR #295, opened 2025-01-17
SQL-2533: add third party notices

terakilobyte:
This adds the 3rd party notices to both the linux distributable tar and the windows msi.

It will in fact regenerate them every time, but I've also included a (static) one that we can update from time to time for compliance.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -789,11 +789,10 @@ functions:
           rm target/release/mock_mongosqltranslate.dll
 
           cp target/release/*.dll installer/msi
-          cp THIRD_PARTY_LICENSES.txt installer/msi
           # TODO: Uncomment prior to merging back into master
-          # cp ./README.md ./mongo-odbc-driver.augmented.sbom.json installer/msi
+          # cp ./THIRD_PARTY_LICENSES.txt ./README.md ./mongo-odbc-driver.augmented.sbom.json installer/msi
           # TODO: Delete the following line prior to merging back into master
-          cp ./README.md installer/msi
+          cp ./THIRD_PARTY_LICENSES.txt ./README.md installer/msi
           cd installer/msi
           if [ "$RELEASE_VERSION" == "snapshot" ]; then
               MINOR_VERSION="0.1"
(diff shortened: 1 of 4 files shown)


pmeredit (changes requested):
Let's not reinvent the wheel

Cargo Third Party License
To generate a third-party license file in a Cargo project, you can use tools like cargo-bundle-licenses or cargo-license-generator. Here’s a brief overview of each:

cargo-bundle-licenses: This tool generates a THIRDPARTY file that includes all the licenses of your project’s dependencies. It relies on the information provided in the package metadata and can be configured to output in formats like JSON, YAML, or TOML. You can install it using cargo install cargo-bundle-licenses and then run it with cargo bundle-licenses --format yaml --output THIRDPARTY.yml.
cargo-license-generator: This tool generates a LICENSE file for your project. It’s a cargo subcommand that allows you to specify the license type and author information. You can install it using cargo install cargo-license-generator and then run it with cargo generate-license --author "Your Name" MIT to generate a MIT license file.
These tools help ensure that your project complies with the licensing requirements of its dependencies and provides clear documentation of the licenses used.

terakilobyte:
There's no reinvention of the wheel. I was told we don't want to distribute the file with a `.yml` extension, and we need the header information there.

I could prepend the header to the generated file and rename it from `.yml` to `.txt` but that's equivalent to what I'm doing now imo

pmeredit (approved):
Looks fine, I don't think we need to check in both the txt and yml files though

mattChiaravalloti:
> This adds the 3rd party notices to both the linux distributable tar and the windows msi.
> 
> It will in fact regenerate them every time, but I've also included a (static) one that we can update from time to time for compliance.

If it regenerates them every time we build, why do we need to include them statically? When would we do this time-to-time update for compliance? What would those updates even look like if ultimately this is going to be generated?

terakilobyte:
They'd look identical to what is happening in the script. However, I don't think we should keep the `licenses` file around. I'm testing a patch now to make sure it's safe to delete with what we have.

mattChiaravalloti (approved):

PR #297, opened 2025-02-04
SQL-2615: Add null check for handle in panic_safe_exec_keep_diagnostics

bucaojit:
Fix for the ASAN test failure. 

I also added an update to the `openssl` dependency since it was failing the `sbom` check.

@EthanHardyMongo The fix I mentioned to you for the sbom failure is here, I ran `cargo update -p openssl` to update it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2394,9 +2394,9 @@ dependencies = [
 
 [[package]]
 name = "openssl"
-version = "0.10.68"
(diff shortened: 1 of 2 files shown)


terakilobyte (approved):
LGTM. Nice find and fix.

EthanHardyMongo (approved):
LGTM! 💯

PR #300, opened 2025-02-07
SQL-2615: Switch handle conversion to TryFrom to handle null values

bucaojit:
Switched `From<Handle>` to `TryFrom<Handle>`. 
Updated the uses to use a macro `try_mongo_handle!` that checks for error, if so returns `SqlReturn::INVALID_HANDLE`.
Added unit test.

### DIFF
--- a/odbc/src/handles/unit.rs
+++ b/odbc/src/handles/unit.rs
@@ -4,8 +4,9 @@
     clippy::cast_possible_wrap
 )]
 
+use crate::api::errors::ODBCError;
 use crate::{handles::definitions::*, has_odbc_3_behavior, SQLAllocHandle, SQLFreeHandle};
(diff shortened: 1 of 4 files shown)


odbc/src/handles/unit.rs
```diff
@@ -151,6 +152,64 @@ fn test_statement_alloc_free() {
...
+#[test]
+fn test_try_from_null_handles() {
+    let null_handle: Handle = std::ptr::null_mut();
+    let result = MongoHandleRef::try_from(null_handle);
+    assert!(
+        result.is_err(),
+        "Expected conversion of a null Handle to fail"
+    );
+    assert!(
+        matches!(result, Err(ODBCError::InvalidHandleType(msg)) if msg == "handle cannot be null"),
```
pmeredit, line 164: this is so cool (the matches)

pmeredit (approved):
Awesome

PR #307, opened 2025-02-26
SQL-2288: Move EAP cluster handling into master branch.

EthanHardyMongo:
The PR that fixed the error handling had a bunch of other changes, so it was easier to manually copy over the relevant parts.


core/src/conn.rs
```diff
@@ -234,8 +234,9 @@ impl MongoConnection {
                 // Ensure the library is loaded if Enterprise edition is detected
                 if get_mongosqltranslate_library().is_none() {
                     return Err(Error::UnsupportedClusterConfiguration(
-                        "Enterprise edition detected, but mongosqltranslate library not found."
-                            .to_string(),
+                        "Enterprise edition detected, but mongosqltranslate library not found. \
+                        You are either (1) trying to connect to an on-premises cluster and have the wrong driver \
+                        or are (2) trying to connect to an Atlas cluster directly instead of through an ADF instance.".to_string(),
```
EthanHardyMongo, line 239: This message is the only difference between what's in this PR and what's in the on-prem driver. I think this will help prevent future confusion.

mattChiaravalloti (approved):
lgtm 👍

core/src/conn.rs
nbagnard, line 239:
```suggestion
                        or are (2) trying to connect directly to an Atlas cluster directly instead of through a Data Federation instance.".to_string(),
```

nbagnard (approved):
Just a small nit because users don't know what is ADF I think

PR #309, opened 2025-02-28
SQL-2660: additional logging for ODBC

terakilobyte:
I ended up reworking this to put the diagnostic info on the connection, where it was more easily accessible in the odbc layer.

If the error occurs on a connection or statement, the diagnostics are read from the corresponding MongoConnection and logged at error level.

There is also a  debug level logging event in prepare. I ended up going with debug because trace is very noise, and we'd probably ask for debug logs first.

### DIFF
--- a/core/src/err.rs
+++ b/core/src/err.rs
@@ -6,6 +6,23 @@ use constants::{
 use mongodb::error::{ErrorKind, WriteFailure};
 use thiserror::Error;
 
+#[derive(Debug, Clone)]
+pub struct Diagnostics {
+    pub query: String,
+    pub schema: String,
+    pub translation: String,
+}
+
+impl Diagnostics {
+    pub fn new(query: String, schema: String, translation: String) -> Self {
+        Self {
+            query,
+            schema,
+            translation,
+        }
+    }
+}
+
 pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Error, Debug, Clone)]
--- a/core/src/err.rs
+++ b/core/src/err.rs
@@ -13,7 +13,7 @@ pub struct QueryDiagnostics {
     pub translation: String,
 }
 
-impl Diagnostics {
+impl QueryDiagnostics {
     pub fn new(query: String, schema: String, translation: String) -> Self {
         Self {
             query,
--- a/core/src/err.rs
+++ b/core/src/err.rs
@@ -7,7 +7,7 @@ use mongodb::error::{ErrorKind, WriteFailure};
 use thiserror::Error;
 
 #[derive(Debug, Clone)]
-pub struct Diagnostics {
+pub struct QueryDiagnostics {
     pub query: String,
     pub schema: String,
     pub translation: String,
--- a/core/src/query.rs
+++ b/core/src/query.rs
@@ -2,7 +2,7 @@ use crate::{
     cluster_type::MongoClusterType,
     col_metadata::{MongoColMetadata, ResultSetSchema, SqlGetSchemaResponse},
     conn::MongoConnection,
-    err::{Diagnostics, Result},
+    err::{QueryDiagnostics, Result},
     mongosqltranslate::{
         libmongosqltranslate_run_command, CommandResponse, GetNamespaces, Namespace, Translate,
         TranslateCommandResponse,
@@ -212,7 +212,7 @@ impl MongoQuery {
         query: &str,
         type_mode: TypeMode,
         max_string_length: Option<u16>,
-    ) -> Result<(Self, Diagnostics)> {
+    ) -> Result<(Self, QueryDiagnostics)> {
         log::debug!("Preparing query with metadata - query: {query}");
         let working_db = current_db.as_ref().ok_or(Error::NoDatabase)?;
         let db = client.client.database(working_db);
@@ -298,7 +298,8 @@ impl MongoQuery {
             "Prepared query with metadata - pipeline: {pipeline:?}, json_schema: {json_schema:?}, result_set_schema: {result_set_schema:?}",
         );
 
-        let diagnostics = Diagnostics::new(query.to_string(), json_schema, format!("{pipeline:?}"));
+        let diagnostics =
+            QueryDiagnostics::new(query.to_string(), json_schema, format!("{pipeline:?}"));
 
         Ok((
             Self {
--- a/core/src/query.rs
+++ b/core/src/query.rs
@@ -295,7 +295,7 @@ impl MongoQuery {
             result_set_schema.process_result_metadata(working_db, type_mode, max_string_length)?;
 
         log::debug!(
-            "Prepared query with metadata - pipeline: {pipeline:?}, json_schema: {json_schema:?}",
+            "Prepared query with metadata - pipeline: {pipeline:?}, json_schema: {json_schema:?}, result_set_schema: {result_set_schema:?}",
         );
 
         let diagnostics = Diagnostics::new(query.to_string(), json_schema, format!("{pipeline:?}"));
--- a/core/src/query.rs
+++ b/core/src/query.rs
@@ -2,7 +2,7 @@ use crate::{
     cluster_type::MongoClusterType,
     col_metadata::{MongoColMetadata, ResultSetSchema, SqlGetSchemaResponse},
     conn::MongoConnection,
-    err::Result,
+    err::{Diagnostics, Result},
     mongosqltranslate::{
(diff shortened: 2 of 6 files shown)


core/src/query.rs
```diff
@@ -160,9 +166,17 @@ impl MongoQuery {
...
+            let collections_schema_doc =
+                schema_catalog_doc
+                    .get_document_mut(current_db)
+                    .map_err(|e| {
+                        log::warn!(
+                            "Failed to get schema document for database `{0}`: {1}",
+                            current_db,
+                            e
+                        );
+                        Error::SchemaValueAccess(current_db.to_string(), e)
```
terakilobyte, line 178: In looking at the errors, I determined this wasn't the same as a ValueAccess error elsewhere, since it his highly localized to schema reads.

core/src/err.rs
```diff
@@ -188,3 +252,28 @@ impl Error {
...
+
+    #[test]
+    fn test_error_details_exist() {
+        let diagnostics = QueryDiagnostics {
+            query: Arc::from("test_query"),
+            schema: Arc::from("test_schema"),
+            translation: Arc::from("test_translation"),
+        };
+    }
+}
```
terakilobyte, line 279: Determining the best way to have a test that ensures errors that should have diagnostics do have diagnostics so we can catch it during testing rather than an `assert` at runtime.

odbc/src/api/errors.rs
```diff
@@ -235,6 +236,9 @@ impl ODBCError {
 
 impl From<mongo_odbc_core::Error> for ODBCError {
     fn from(err: mongo_odbc_core::Error) -> Self {
+        if let Some(details) = err.details() {
+            log::error!("{details:?}");
+        }
```
terakilobyte, line 241: This and the following is why I went the trait route. When we convert a core error to an odbc error, it's easy to get the information we want to log this way.

core/src/query.rs
```diff
@@ -339,13 +370,19 @@ impl MongoStatement for MongoQuery {
     // Get the BSON value for the cell at the given colIndex on the current row.
     // Fails if the first row as not been retrieved (next must be called at least once before getValue).
     fn get_value(&self, col_index: u16, max_string_length: Option<u16>) -> Result<Option<Bson>> {
-        let current = self.current.as_ref().ok_or(Error::InvalidCursorState)?;
+        // let current = self.current.as_ref().ok_or(Error::InvalidCursorState)?;
+        let current = self
+            .current
+            .as_ref()
+            .ok_or(Error::InvalidCursorStateWithDiagnostics(self.into()))?;
```
pmeredit, line 377: Use ok_or_else, arc cloning is still not free

pmeredit (commented):
Looks fine to me. I don't think the Arc is really necessary, but it also doesn't hurt, so it's cool. Just using _or_else really gets rid of most of the benefit (the performance of the error case doesn't really matter)

core/src/err.rs
```diff
@@ -38,14 +52,24 @@ pub enum Error {
...
     UnknownColumn(String),
     #[error("Error retrieving data for field {0}: {1}")]
+    ValueAccessWithDiagnostics(
+        String,
+        mongodb::bson::document::ValueAccessError,
+        QueryDiagnostics,
+    ),
+    #[error("Error retrieving data for field {0}: {1}")]
     ValueAccess(String, mongodb::bson::document::ValueAccessError),
+    #[error("Error retrieving schema information, data for field {0}: {1}")]
```
nbagnard, line 71:
```suggestion
    #[error("Error retrieving schema information data for field {0}: {1}")]
```

core/src/query.rs
```diff
@@ -212,10 +212,11 @@ impl MongoQuery {
         query: &str,
         type_mode: TypeMode,
         max_string_length: Option<u16>,
-    ) -> Result<Self> {
+    ) -> Result<(Self, Diagnostics)> {
+        log::debug!("Preparing query with metadata - query: {query}");
```
terakilobyte, line 216: Adding this here as it's possible for the all to fail in translate and the query would help us.

pmeredit (approved):
LGTM!

odbc/src/api/functions.rs
```diff
@@ -3389,7 +3389,13 @@ fn sql_prepare(
             Err(ODBCError::InvalidCursorState)
         }
     };
-    mongo_statement
+    match mongo_statement_result {
+        Ok((mongo_statement, diagnostics)) => {
```
nbagnard, line 3393: My understanding from reading the code is that :
- We stash the internal diagnostics (query info) on the Connection.
- The last call to `sql_prepare` overrides whatever is in *connection.diagnostics already with the current query diagnostic.
- When an error happen and we want to add a new ODBC diagnostics (I would like to change the name of our internal diagnostics info to something clearly flagging it as our own) to the handle we use `add_diag_info`. In this function,  you retrieve the diagnostics from the connection associated with the statement and log it.

A query is always associated to a statement, not just a connection I think. 
Why is the internal diagnostics stored on the Connection?

I am afraid that it will work the way you expect.

If I understand the code correctly, this flow will report the query diagnostics for the query associated with the statement S2, while it happens on the query associated with S1.
[c1][s1] Prepare
[c1][s2] Prepare
[c1][s2] Execute
[c1][s1] Execute <-- Errors

This flow will also be problematic I think.
[c1][s1] ExecDirect
[c1][s2] ExecDirect
[c1][s2] GetData+
[c1][s1] GetData+ <-- Errors

odbc/src/handles/definitions.rs
```diff
@@ -71,9 +71,31 @@ impl MongoHandle {
...
             MongoHandle::Statement(s) => {
+                // SAFETY: we must ensure the connection is not null,
+                // otherwise this entire chain will cause unhappiness and chaos
+                if !s.connection.is_null() {
+                    unsafe {
+                        if let Some(ref diagnostics) = s
+                            .connection
+                            .as_ref()
+                            .unwrap()
+                            .as_connection()
```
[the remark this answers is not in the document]
terakilobyte: The funciton is implemented for `MongoHandle`, but we're in the variant here so it won't work without some refactoring.

core/src/err.rs
```diff
@@ -6,6 +6,23 @@ use constants::{
 use mongodb::error::{ErrorKind, WriteFailure};
 use thiserror::Error;
 
+#[derive(Debug, Clone)]
+pub struct Diagnostics {
+    pub query: String,
+    pub schema: String,
```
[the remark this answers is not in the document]
terakilobyte: We could definitely add it, but it's also super easy to recreate from the schema and query. Considering this is stored for every statement, it's up to you.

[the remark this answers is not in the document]
nbagnard: yeah, I see what you mean. I am not sure it is a huge issue if we clean it up when the statement is closed.
Can we at least log it from `translate_sql` in debug so that we can always ask the customer to repro and send the log and we then will have all the info in the logs instead of having to recreate some of it.
I think this way we also empower TSEs to investigate without our help.

odbc/src/handles/definitions.rs
```diff
@@ -74,6 +74,13 @@ impl MongoHandle {
                 c.errors.write().unwrap().push(error);
             }
             MongoHandle::Statement(s) => {
+                // SAFETY: we must ensure the connection is not null,
+                // otherwise this entire chain will cause unhappiness and chaos
+                if !s.connection.is_null() {
+                    if let Some(diagnostics) = s.diagnostics.read().unwrap().as_ref() {
+                        log::error!("Diagnostics: {:?}", diagnostics);
```
nbagnard, line 81:
```suggestion
                        log::error!("Query diagnostics: {:?}", diagnostics);
```

core/src/err.rs
```diff
@@ -6,6 +6,23 @@ use constants::{
 use mongodb::error::{ErrorKind, WriteFailure};
 use thiserror::Error;
 
+#[derive(Debug, Clone)]
+pub struct Diagnostics {
```
nbagnard, line 10: [nit] Would you be Ok renaming it QueryDiagnostics for newbies to avoid confusion with the ODBC diagnostics?
```suggestion
pub struct QueryDiagnostics {
```

nbagnard (changes requested):
I think I'd really like to have the ability to log the result set schema.
If we can troubleshoot only from a log file, then it is very valuable

nbagnard (approved):
LGTM!
You will have to fix compile issue I think because QueryDiagnostics renaming was not propagated

PR #314, opened 2025-03-13
SQL-2695: safely handle date, time, and datetime construction

terakilobyte:
It was folly to blindly use expect here, my bad. With the addition of these helper functions we'll generate an ODBCError instead of panicking.


nbagnard (commented):
LGMT!

nbagnard (approved):

PR #316, opened 2025-03-19
SQL-2683: switch to mongosql rlib

terakilobyte:
This moves to using mongosql as an rlib instead of a dll.


pmeredit (approved):
LGTM!

core/src/json_schema.rs
```diff
@@ -157,13 +193,52 @@ pub enum BsonTypeName {
...
+            mongosql::json_schema::BsonTypeName::Regex => Self::Regex,
+            mongosql::json_schema::BsonTypeName::DbPointer => Self::DbPointer,
+            mongosql::json_schema::BsonTypeName::Javascript => Self::Javascript,
+            mongosql::json_schema::BsonTypeName::Symbol => Self::Symbol,
+            mongosql::json_schema::BsonTypeName::JavascriptWithScope => Self::JavascriptWithScope,
+            mongosql::json_schema::BsonTypeName::Timestamp => Self::Timestamp,
+            mongosql::json_schema::BsonTypeName::MinKey => Self::MinKey,
+            mongosql::json_schema::BsonTypeName::MaxKey => Self::MaxKey,
+            mongosql::json_schema::BsonTypeName::Undefined => Self::Undefined,
+        }
```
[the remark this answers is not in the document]
terakilobyte: I think the proc macro idea is a better long term solution, but outside of this pr as we might find it useful elsewhere. I like the macro idea, but also don't like it because it doesn't really offer any improvement other than a little bit of horizontal text gone, but adds the macro and additional processing time that incurs in rust analyzer and other tools.

odbc/build.rs
```diff
@@ -1,16 +1,13 @@
...
 
-#[cfg(windows)]
-fn main() {
-    // the default for winres is to parse the version number from the neighboring Cargo.toml
-    let mut res = winres::WindowsResource::new();
-    res.set_version_info(winres::VersionInfo::FILETYPE, 2); // Dll file
-    res.set("CompanyName", "MongoDB Inc.");
-
-    // compile will write the rc file, and enable the cargo linker to link the compiled resource file
-    res.compile().unwrap();
```
EthanHardyMongo, line 12: Why are we removing the comments here? Are they no longer correct?

evergreen.yml
```diff
@@ -105,37 +95,31 @@ functions:
           export COMPLIANCE_REPORT_NAME="mongo-odbc-driver_compliance_report.md"
           export STATIC_CODE_ANALYSIS_NAME="mongo-odbc-driver.sast.sarif"
           if [[ "${triggered_by_git_tag}" != "" ]]; then
-            # tag should be formatted as `v<ODBC major>.<minor>.<patch>-libv<libmongosqltranslate major>.<minor>.<patch>`
             # we split on `-libv` to get the two versions
             export RELEASE_VERSION=$(echo ${triggered_by_git_tag} | awk -F'-libv' '{print $1}' | sed s/v// )
-            export LIBMONGOSQLTRANSLATE_VERSION=$(echo ${triggered_by_git_tag}  | awk -F'-libv' '{print $2}')
```
EthanHardyMongo, line 111: Lines 109 and 110 can be updated to remove the `libv` stuff.

core/src/err.rs
```diff
@@ -75,10 +75,8 @@ pub enum Error {
         "The ODBC driver version `{0}` is incompatible with libmongosqltranslate version `{1}`"
     )]
     LibmongosqltranslateLibraryIsIncompatible(&'static str, String),
-    #[error(
-        "The libmongosqltranslate command `{0}` failed. Error message: `{1}`. Error is internal: {2}"
-    )]
-    LibmongosqltranslateCommandFailed(&'static str, String, bool),
+    #[error("The libmongosqltranslate command `{0}` failed. Error message: `{1}`.")]
+    LibmongosqltranslateCommandFailed(&'static str, String),
```
EthanHardyMongo, line 79:
```suggestion
    #[error("The mongosql command `{0}` failed. Error message: `{1}`.")]
    MongosqlCommandFailed(&'static str, String),
```
I think we should rename all the `libmongosqltranslate` stuff to `mongosql`. I noticed there's other code that can also be renamed to `mongosql`. I think it would be good to convert all instances of `mongosqltranslate` and `libmongosqltranslate` to `mongosql`. Also, I believe the error above this one and several below it can be removed since they were only used by mongosqltranslate.

EthanHardyMongo (changes requested):
Overall, looks great! I just had some clean up suggestions.

odbc/build.rs
terakilobyte: Nope, I just deleted them when moving code around. I can add them back.

evergreen.yml
```diff
@@ -105,37 +95,31 @@ functions:
           export COMPLIANCE_REPORT_NAME="mongo-odbc-driver_compliance_report.md"
           export STATIC_CODE_ANALYSIS_NAME="mongo-odbc-driver.sast.sarif"
           if [[ "${triggered_by_git_tag}" != "" ]]; then
-            # tag should be formatted as `v<ODBC major>.<minor>.<patch>-libv<libmongosqltranslate major>.<minor>.<patch>`
             # we split on `-libv` to get the two versions
-            export RELEASE_VERSION=$(echo ${triggered_by_git_tag} | awk -F'-libv' '{print $1}' | sed s/v// )
-            export LIBMONGOSQLTRANSLATE_VERSION=$(echo ${triggered_by_git_tag}  | awk -F'-libv' '{print $2}')
+            export RELEASE_VERSION=$(echo ${triggered_by_git_tag} | sed s/v// )
```
EthanHardyMongo, line 99: [nit] Let's remove `# we split on -libv to get the two versions`. (Unfortunately, github won't let me suggest changes here)

EthanHardyMongo (approved):
Looks great! I just had one, really tiny nit.

PR #263, opened 2024-11-07
fixing some yaml warnings

terakilobyte:
Validating our evergreen file there were warnings that this cleans up. I'm still trying to figure out why the [release process is failing](https://spruce.mongodb.com/version/mongosql_odbc_driver_eap_v2.0.0_alpha_libv1.0.0_alpha_1_672ccc6682ee320007eee133/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), even though I can do a patch release and it works fine.

PR #320, opened 2025-03-25
SQL-2714: get latest mongosql tag in evergreen

terakilobyte:
I also:
- ran cargo update, so the lockfile changed
- fixed a new error about type information being needed

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -102,6 +102,9 @@ functions:
           export MSI_FILENAME="mongoodbc-eap-$RELEASE_VERSION.msi"
           export UBUNTU_FILENAME="mongoodbc-eap-$RELEASE_VERSION.tar.gz"
 
+          # get latest mongosql tag
+          export MONGOSQL_VERSION=$(curl -s "https://api.github.com/repos/mongodb/mongosql/tags" | jq -r '.[0].name')
+
           cat <<EOT > expansions.yml
           RELEASE_VERSION: "$RELEASE_VERSION"
           MSI_FILENAME: "$MSI_FILENAME"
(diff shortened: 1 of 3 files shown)


evergreen.yml
```diff
@@ -672,6 +681,9 @@ functions:
         script: |
           ${prepare_shell}
 
+          # change mongosql version in core/Cargo.toml from branch to tag
+          sed -i "s/branch = \"main\"/tag = \"${MONGOSQL_VERSION}\"/" core/Cargo.toml
```
nbagnard, line 685: I think we should do this as part of "set and check package version".
First, it will be only in one place instead of 2 which is better for maintainability. 
Second, it might impact the SBOM content. We want to have to correct version reported if it appears as a dependency (it should). 
Unfortunately, I cannot tell because the SBOM generation was disabled in https://github.com/mongodb/mongo-odbc-driver/pull/263. Can you check if the `sbom` task can be re-enabled. I agree that we don't need to produce the SSDLC release artifacts, but it would be good to have the sbom task still running. This will make sure that if we introduce new/different libraries, we are also clear.

evergreen.yml
```diff
@@ -657,6 +662,10 @@ functions:
         working_dir: mongosql-odbc-driver
         script: |
           ${prepare_shell}
+
+          # change mongosql version in core/Cargo.toml from branch to tag
+          sed -i "s/branch = \"main\"/tag = \"${MONGOSQL_VERSION}\"/" core/Cargo.toml
```
bucaojit, line 667:
```suggestion
          sed -i "/^mongosql = {.*branch = \"main\".*}/ s/branch = \"main\"/tag = \"${MONGOSQL_VERSION}\"/" core/Cargo.toml
```
Added a check for `mongosql` in the line.

terakilobyte: Updated. The syntax differs slightly but I verified in the logs it compiles the right version

bucaojit (approved):

nbagnard (approved):
LGTM

PR #323, opened 2025-04-16
SQL-2543:  prepare to build both in one shot

terakilobyte:


evergreen/create-expansions.sh
```diff
@@ -0,0 +1,118 @@
...
+  export ODBCSYSINI="$(pwd)"/setup
+EOT
+
+cat expansions.yml
+
+curl https://translators-connectors-releases.s3.amazonaws.com/mongosql-odbc-driver/ubuntu2204/1.4.3/release/mongoodbc.tar.gz --output mongoodbc.tar.gz
+curl -LO https://translators-connectors-releases.s3.amazonaws.com/mongosql-odbc-driver/ubuntu2204/1.4.3/release/mongoodbc.tar.gz.sig
+curl -LO https://pgp.mongodb.com/atlas-sql-odbc.asc
+gpg --import atlas-sql-odbc.asc
+gpg --verify mongoodbc.tar.gz.sig mongoodbc.tar.gz
```
bucaojit, line 115: This looks like it was leftover from debugging perhaps.  We can remove it.

evergreen/create-expansions.sh
```diff
@@ -0,0 +1,118 @@
...
+fi
+export DUMP_FOLDER=dumps
+export LOCAL_DUMP_ORIGINAL_REG_VAL=local_dump_original_value.reg
+export MONGOODBC_DEBUGGING_INFO_ARCHIVE=crashDebuggingInfo
+export SCRIPT_FOLDER=resources
+export COMPLIANCE_REPORT_NAME="mongo-odbc-driver_compliance_report.md"
+export STATIC_CODE_ANALYSIS_NAME="mongo-odbc-driver.sast.sarif"
+export FEATURE_FLAGS=""
+export PRODUCT_NAME="mongoodbc"
+export PATH_PREFIX=""
```
bucaojit, line 17: This could have been added for the S3 path.  I don't see it being used, it can be removed if not.

evergreen/eap.yml
```diff
@@ -0,0 +1,360 @@
...
+      - name: result-set-test-eap
+        variant: ".release-variant"
+      - name: sign-eap
+        variant: ".release-variant"
+    commands:
+      - func: "upload eap release"
+
+  - name: compile-eap
+    commands:
+      - func: "generate expansions"
```
bucaojit, line 179: Cool that's a good workaround for setting the values in the generate expansion.  So it'll run it once for the `pre` task but we want to set snapshot-eap to true so this will update the expansions with that variable set the way we want.  With it set, do we still need to set `vars: snapshot-eap: "true"` in the commands after this?  It's fine that it's there, but I would have expected it to already be set here.

evergreen/eap.yml
```diff
@@ -0,0 +1,360 @@
...
+        aws_key: ${release_aws_key}
+        aws_secret: ${release_aws_secret}
+        local_file: mongosql-odbc-driver/target/release/libatsql.so
+        remote_file: eap/mongosql-odbc-driver/ubuntu2204/${RELEASE_VERSION}/release/libatsql.so
+        bucket: translators-connectors-releases
+        permissions: public-read
+        content_type: application/octet-stream
+
+tasks:
+  - name: snapshot-eap
```
bucaojit, line 158: I think I'll need to do the same with mine and create a separate snapshot-eap task.  I was trying to figure out how to not have to duplicate the tasks for this but ended up only uploading a regular snapshot for the mainline build.  I'll think about this a bit more for the jdbc merge.

bucaojit (changes requested):
Nice changes, and nice way of setting the variable when running `generate expansions`.

Is there an evergreen run of the release task so I can double-check the snapshot-eap tasks?

evergreen/eap.yml
terakilobyte: Yeah I removed the vars invocation from the other function calls.

evergreen/create-expansions.sh
terakilobyte: Yeah I removed it since the tasks have to be duplicated :(

bucaojit (approved):
LGTM!

evergreen/eap.yml
```diff
@@ -0,0 +1,344 @@
...
+        aws_secret: ${aws_secret}
+        remote_file: mongosql-odbc-driver/artifacts/docs/MongoDB_ODBC_Guide.pdf
+        local_file: mongosql-odbc-driver/release/docs/MongoDB_ODBC_Guide.pdf
+        bucket: mciuploads
+    - command: s3.put
+      params:
+        aws_key: ${release_aws_key}
+        aws_secret: ${release_aws_secret}
+        local_file: mongosql-odbc-driver/release/docs/MongoDB_ODBC_Guide.pdf
+        remote_file: eap/mongodb-odbc-driver/docs/MongoDB_ODBC_Guide.pdf
```
nbagnard, line 15: Related to the change of directory where the release artifacts are copied in now, the guide will still be under `eap`, while the other artifacts will be under `mongodb-eap`. It would be better to unify and use the same directory for all release artifacts.

nbagnard (changes requested):
Just a couple of minor questions/comments, but otherwise, it's looking good :)

evergreen.yml
```diff
@@ -1525,13 +1519,18 @@ functions:
           # Start MongoDB instances
           ./resources/start_local_mdb.sh $mdb_version_com $mdb_version_ent $arch
 
-
-          # Enterprise test
+          # Enterprise test no eap
           cargo test --features cluster_type_tests -- cluster_type::test_determine_cluster_type_enterprise_succeeds --nocapture
+          # Enterprise test eap
+          cargo test --features cluster_type_tests, eap -- cluster_type::test_determine_cluster_type_enterprise_succeeds --nocapture
           ENTERPRISE_EXITCODE=$?
```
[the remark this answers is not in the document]
terakilobyte: I don't believe so. We should see the failures. However, I should be capturing the exit code!

evergreen/create-expansions.sh
```diff
@@ -0,0 +1,104 @@
...
+export MSI_FILENAME="$PRODUCT_NAME-$RELEASE_VERSION.msi"
+export UBUNTU_FILENAME="$PRODUCT_NAME-$RELEASE_VERSION.tar.gz"
+
+cat <<EOT >expansions.yml
+RELEASE_VERSION: "$RELEASE_VERSION"
+FEATURE_FLAGS: "$FEATURE_FLAGS"
+PRODUCT_NAME: "$PRODUCT_NAME"
+MSI_FILENAME: "$MSI_FILENAME"
+UBUNTU_FILENAME: "$UBUNTU_FILENAME"
+WINDOWS_INSTALLER_PATH: "${PRODUCT_NAME}/mongosql-odbc-driver/windows/$RELEASE_VERSION/release/$MSI_FILENAME"
```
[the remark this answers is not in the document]
terakilobyte: Good catch!

evergreen/eap.yml
terakilobyte: Yes I believe it would!

evergreen.yml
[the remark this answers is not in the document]
nbagnard: Yes, that's what I meant. The task would still show a success even if the non eap test failed.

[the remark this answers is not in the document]
terakilobyte: In testing it I found it didn't. However, each particular test run is now captured and all are checked.

evergreen/eap.yml
nbagnard: @terakilobyte Sorry for the misunderstanding. I was thinking about unifying between `mongoodbc-eap` and `eap` as the root directory for eap builds.
It looks like right now, with the new changes, the msi and tar.gz will be in the same root directory than the regular non eap builds, but some other like `atsql.dll` will be in the eap directory.
Here is what I feel could work:
- Call generate expansion with eap true before calling upload release.
- Remove the hard coded `mongosql-odbc-driver` from `WINDOWS_INSTALLER_PATH` and `UBUNTU2204_INSTALLER_PATH`.
- Add a variable in `create-expansions.sh`, for example `RELEASE_ROOT_DIR` which should be `mongosql-odbc-driver` for the regular non-eap releases and `eap/mongosql-odbc-driver` for eap releases
- Use the variable in `upload release` to control where things go
- Remove `upload eap release` from eap.yml and use `upload release` instead after calling `generate expansion` with the var `snapshot-eap: "true"`

nbagnard (changes requested):
I think where the release artifacts go in the release bucket need one last pass.
Otherwise, it looks all good.

nbagnard (approved):
LGTM :)

PR #324, opened 2025-04-17
SQL-2781: Handle empty select list the same than missing select list

nbagnard:


core/src/col_metadata.rs
```diff
@@ -256,24 +256,26 @@ impl ResultSetSchema {
...
+                        v => v,
+                    })
+                    .collect()
+            } else {
+                // given a select order, convert the values of the map into an ordered vector
+                self.select_order
+                    .as_ref()
+                    .unwrap()
+                    .iter()
+                    .map(|key| processed_result_set_metadata.remove(key).unwrap())
```
terakilobyte, line 275: It's probably "safer" here to use `filter_map` and obviate the unwrap.


```suggestion
                    .filter_map(|key| processed_result_set_metadata.remove(key))
```

terakilobyte (approved):
I think it would be good to capture our select order behavior in a unit test, perhaps something like the following:

```rs

        #[test]
        fn test_result_set_metadata_ordering() {
            use crate::{
                col_metadata::ResultSetSchema,
                json_schema::{BsonType, BsonTypeName, Schema},
                map, TypeMode,
            };

            let input = ResultSetSchema {
                schema: Schema {
                    bson_type: Some(BsonType::Single(BsonTypeName::Object)),
                    properties: Some(map! {
                        "datasource2".to_string() => Schema {
                            bson_type: Some(BsonType::Single(BsonTypeName::Object)),
                            properties: Some(map! {
                                "field_b".to_string() => Schema {
                                    bson_type: Some(BsonType::Single(BsonTypeName::Int)),
                                    ..Default::default()
                                },
                                "field_a".to_string() => Schema {
                                    bson_type: Some(BsonType::Single(BsonTypeName::String)),
                                    ..Default::default()
                                }
                            }),
                            ..Default::default()
                        },
                        "datasource1".to_string() => Schema {
                            bson_type: Some(BsonType::Single(BsonTypeName::Object)),
                            properties: Some(map! {
                                "field_c".to_string() => Schema {
                                    bson_type: Some(BsonType::Single(BsonTypeName::Double)),
                                    ..Default::default()
                                },
                                "field_d".to_string() => Schema {
                                    bson_type: Some(BsonType::Single(BsonTypeName::Bool)),
                                    ..Default::default()
                                }
                            }),
                            ..Default::default()
                        }
                    }),
                    ..Default::default()
                },
                select_order: Some(vec![
                    vec!["datasource1".to_string(), "field_d".to_string()],
                    vec!["datasource2".to_string(), "field_a".to_string()],
                    vec!["datasource1".to_string(), "field_c".to_string()],
                    vec!["datasource2".to_string(), "field_b".to_string()],
                ]),
            };

            // Select order is specified, so the order should be preserved
            let result_with_select_order = input
                .process_result_metadata("test_db", TypeMode::Standard, None)
                .unwrap();

            assert_eq!(4, result_with_select_order.len());
            assert_eq!("datasource1", result_with_select_order[0].table_name);
            assert_eq!("field_d", result_with_select_order[0].col_name);
            assert_eq!("datasource2", result_with_select_order[1].table_name);
            assert_eq!("field_a", result_with_select_order[1].col_name);
            assert_eq!("datasource1", result_with_select_order[2].table_name);
            assert_eq!("field_c", result_with_select_order[2].col_name);
            assert_eq!("datasource2", result_with_select_order[3].table_name);
            assert_eq!("field_b", result_with_select_order[3].col_name);

            // No select order, sorted alphabetically
            let mut input_without_select_order = input.clone();
            input_without_select_order.select_order = None;

            let result_without_select_order = input_without_select_order
                .process_result_metadata("test_db", TypeMode::Standard, None)
                .unwrap();

            assert_eq!(4, result_without_select_order.len());
            assert_eq!("datasource1", result_without_select_order[0].table_name);
            assert_eq!("field_c", result_without_select_order[0].col_name);
            assert_eq!("datasource1", result_without_select_order[1].table_name);
            assert_eq!("field_d", result_without_select_order[1].col_name);
            assert_eq!("datasource2", result_without_select_order[2].table_name);
            assert_eq!("field_a", result_without_select_order[2].col_name);
            assert_eq!("datasource2", result_without_select_order[3].table_name);
            assert_eq!("field_b", result_without_select_order[3].col_name);

            // Empty selcect order should be treated as None
            let mut input_with_empty_select_order = input.clone();
            input_with_empty_select_order.select_order = Some(Vec::new());

            let result_with_empty_select_order = input_with_empty_select_order
                .process_result_metadata("test_db", TypeMode::Standard, None)
                .unwrap();

            assert_eq!(4, result_with_empty_select_order.len());
            assert_eq!("datasource1", result_with_empty_select_order[0].table_name);
            assert_eq!("field_c", result_with_empty_select_order[0].col_name);
            assert_eq!("datasource1", result_with_empty_select_order[1].table_name);
            assert_eq!("field_d", result_with_empty_select_order[1].col_name);
            assert_eq!("datasource2", result_with_empty_select_order[2].table_name);
            assert_eq!("field_a", result_with_empty_select_order[2].col_name);
            assert_eq!("datasource2", result_with_empty_select_order[3].table_name);
            assert_eq!("field_b", result_with_empty_select_order[3].col_name);
        }
```

jchemburkar (approved):

core/src/col_metadata.rs
nbagnard: @terakilobyte We are not filtering and `procesed_result_set_metadata` is short lived.
What about doing ' .map(|key| processed_result_set_metadata.get(key))` instead?
It achieves the same purpose and matches the logic.

terakilobyte: `filter_map` just removes None values, no? It just removes the need to call `unwrap`.

nbagnard: ah, I see, I didn't think about the possible case where a key is not present in the result set metadata.
Something would not be quite right if that was the case.
Let me see, we should gracefully error out, not panic, but also not just ignore the missing key

nbagnard:
> I think it would be good to capture our select order behavior in a unit test, perhaps something like the following:

@terakilobyte 
I added 2 more unit tests covered empty select list order and select list order containing a field not present in the result set metadata. The other 2 tests already existed.

terakilobyte (approved):
Nice cleanup here!

PR #325, opened 2025-04-18
SQL-2728: Update to use git ls-remote

terakilobyte:
Also requested a review from copilot to see what it does.


nbagnard (approved):
LGTM!

PR #329, opened 2025-05-22
SQL-2830: Enable building of eap

terakilobyte:
Due to changes in mongosql we find ourselves in this interim state. @jchemburkar you will probably want to file another ticket as soon as the rest of derivation is in and released, otherwise it will break again.

Clippy also highlighted that the Statement variant of a MongoHandle is quite large. I took a quick pass at it but there are memory management implications that need to be more thought out.


nbagnard (approved):
LGTM :)

jchemburkar (approved):

PR #334, opened 2025-06-09
SQL-2846: fix windows signing for standard build

jchemburkar:
The periodic builds succeed while the merge patches fail, because patch is not allowed to run the signing tasks


bucaojit (approved):
LGTM

nbagnard (approved):
LGTM!
And Oliver is probably right that it would be good to add to the sign-eap task too.
Unless the patch definition only cares about variants and tasks in the main Evergreen file (which would be a bummer)

PR #338, opened 2025-06-17
SQL-2868: update RELEASE_VERSION

jchemburkar:
The sbom release [task](https://spruce.mongodb.com/task/mongosql_odbc_driver_eap_code_quality_security_sbom__v2.0.0_beta_4_libv1.0.0_beta_5_25_06_17_23_10_15/logs?execution=0) failed. The implication of the error message (generated [here](https://github.com/mongodb/sql-engines-common-test-infra/blob/db60f9fbdcf8e1a9154ac647b94cdb81ecf1c431/evergreen/scripts/set_and_check_packages_version.sh#L23)) is that the `$release_version` variable is empty, and that it was successfully picked up as triggered by git tag. So I think this is just a result of variable case sensitivity, but let me know if there is a better idea. 

I didn't see obvious implications for other variables in a brief scan. I believe this is coming up because it is the first time releasing ODBC since switching over to the shared test infra, which defines the `set and check packages version` task.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -341,9 +341,9 @@ functions:
           ${prepare_shell}
           echo "Author = $AUTHOR"
           echo "Author email = $AUTHOR_EMAIL"
-          echo "Version = $RELEASE_VERSION"
-          SBOM_URL="https://translators-connectors-releases.s3.amazonaws.com/mongosql-odbc-driver/mongosql-odbc-${RELEASE_VERSION}.sbom.json"
-          SARIF_URL="https://translators-connectors-releases.s3.amazonaws.com/mongosql-odbc-driver/mongo-odbc-$RELEASE_VERSION.sast.sarif"
+          echo "Version = $release_version"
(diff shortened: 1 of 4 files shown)


mattChiaravalloti (approved):
Ah sorry I didn't catch this in the previous PR. Nathan, Oliver, and I discussed this in the initial commot-test-infra PR: I decided to go with lowercase names for all of the "arguments" and "env variables" in common-test-infra because they are all evergreen expansions. By convention (in our repos and evergreen's documentation), expansions are lowercase. Sorry it caused confusion.

I know it's weird that we're hanging in this transitional mid-state where tons of things are still defined as env variables while some are transitioned to just expansions. Arguably, in this PR we could have kept line 80 of `create-expansions.sh` exporting `RELEASE_VERSION` all-caps for the env variable but I'm glad you already switched it because eventually all `prepare_shell`s will go away in favor of expansions anyway (all of our `prepare_shell`s exclusively export env variables, which can/will be accomplished by `add_exansions_to_env: true` in all common-test-infra evergreen functions).

nbagnard (approved):
LGTM

PR #340, opened 2025-07-03
SQL-2877: Handle SQLEndTran satisfactorily 

pmeredit:


core/src/bson_type_info.rs
```diff
@@ -109,7 +109,28 @@ pub struct BsonTypeInfo {
     pub simple_type_info: Option<SimpleTypeInfo>,
 }
```
pmeredit, line 111: mew rust C doesn't like comparing function pointers.

core/src/cluster_type.rs
```diff
@@ -29,8 +29,7 @@ pub async fn determine_cluster_type(client: &Client) -> Result<MongoClusterType>
         Some(Bson::Int32(i)) if *i == 1 => {}
```
pmeredit, line 29: everything else in core is clippy updates.

odbc/src/api/functions.rs
```diff
@@ -1269,9 +1269,35 @@ pub unsafe extern "C" fn SQLDriversW(
 pub unsafe extern "C" fn SQLEndTran(
     _handle_type: HandleType,
     handle: Handle,
```
pmeredit, line 1271: Actual change is here.

core/src/bson_type_info.rs
pmeredit: It's a new warning from the newest version of rustc. Comparing function pointers was always UB. We should have removed the derived PartialEQ when we changed to function pointers.

odbc/src/api/functions.rs
```diff
@@ -1269,9 +1269,35 @@ pub unsafe extern "C" fn SQLDriversW(
...
+        // does not do anything.
+        trace_odbc!(
+            info,
+            mongo_handle,
+            "No transaction in progress, returning success for commit",
+            function_name!()
+        );
+        SqlReturn::SUCCESS
+    } else {
+        // the user is either requesting a rollback or some random value not in the spec, either
```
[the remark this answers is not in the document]
pmeredit: afaict it's HY012 in odbc 2 also, but the spec isn't clear. I think it's ok if we return that in odbc 2.

constants/src/lib.rs
```diff
@@ -186,6 +186,10 @@ pub const INVALID_CHARACTER_VALUE: OdbcState<'static> = OdbcState {
     odbc_2_state: "22005",
     odbc_3_state: "22018",
 };
+pub const INVALID_TRANSACTION_OPERATION_CODE: OdbcState<'static> = OdbcState {
+    odbc_2_state: "HY012",
```
nbagnard, line 190: Here is the mapping :)
https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/sqlstate-mappings?view=sql-server-ver17
ODBC 2 is `S1012`

odbc/src/api/functions.rs
```diff
@@ -1272,31 +1272,45 @@ pub unsafe extern "C" fn SQLEndTran(
...
+                info,
+                mongo_handle,
+                "No transaction in progress, returning error for rollback",
+                function_name!()
+            );
+            add_diag_info!(mongo_handle, ODBCError::RollbackNotSupported);
+            SqlReturn::ERROR
+        }
+        _ => {
+            // The user is requesting some random competion type not in the spec, we need to report
```
nbagnard, line 1300: [nit]
```suggestion
            // The user is requesting some random completion type not in the spec, we need to report
```

nbagnard (approved):
Just added the ODBC 2 code and found a minuscule typo.
Otherwise, good to go on my end!

constants/src/lib.rs
pmeredit: nice!

PR #341, opened 2025-07-08
SQL-2849: Set better default scopes for OIDC

pmeredit:
I was going to add a unit test for build_scopes, but it's actually rather hard to create a CoreProviderMetadata. I could change the interface to just take the supported_scopes, but that feels much worse.


bucaojit (approved):
LGTM

PR #344, opened 2025-07-22
SQL-2909: Fix sharp edge in ADF collection listing

pmeredit:
This change was not fast because it took a very long time to actually repro the error. The error could not be reproed on a local mongod because apparently local mongod does not return an Error in this case whereas ADF does. After I realized that, I needed to figure out how to actually create users in a local ADF... which took awhile.

The change itself took 2 seconds, and worked first try, of course :joy:

### DIFF
--- a/core/src/collections.rs
+++ b/core/src/collections.rs
@@ -102,6 +102,14 @@ pub struct MongoCollections {
     table_types_filter: Option<Vec<CollectionType>>,
 }
 
+fn database_filter(db_name: &str) -> bool {
+    !DISALLOWED_DB_NAMES.contains(&db_name)
+}
+
+fn collection_filter(collection_name: &str) -> bool {
+    !DISALLOWED_COLLECTION_NAMES.contains(&collection_name) && !collection_name.starts_with("system.")
+}
+
 // Statement related to a SQLColumns call.
 impl MongoCollections {
     // Create a new MongoStatement to list tables with the given database (catalogs) and collection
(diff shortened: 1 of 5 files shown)


nbagnard (changes requested):
This idea is not to silently ignore errors happening when trying to retrieve the list of collections for a given database.
As I explained in the ticket, we need to filter out empty database names.
You can do this by adding "" to DISALLOWED_DB_NAMES or adding `.filter(|&db_name| !db_name.is_empty())` in the filter list again.
I look at the Git blame history and this was done in [SQL-1440 PR](https://github.com/mongodb/mongo-odbc-driver/pull/136) but unfortunately broken in [SQL-2288 PR](https://github.com/mongodb/mongo-odbc-driver/pull/251).
This is a regression and we didn't catch it, which means that we also need to add a test with an ADF user having the minimal permission set.
Also, when I do a quick test, I see that run_command returns a Result : "called `Result::unwrap()` on an `Err` value: Error { kind: Command(CommandError { code: 17, code_name: "ProtocolError", message: "\"$db\" must not be empty, correlationID = 18548b06a8434559d5fe3ad3", topology_version: None }), labels: {}, wire_version: Some(25), source: None }".
We should find a way to get that information back to the user.

pmeredit:
> By the way, I have no idea how this "works in jdbc" since we can't use jdbc with Power BI and whatever was used to test it was likely doing different things. All I did was change the error handling here, but Power BI now works and correctly shows the available collection. This means that Power BI is probably doing some global listing and falling back to a local one when the global listing fails! So we really can't be sure that jdbc "works" here, since it's being used differently.

Though, I do suspect this wouldn't be a problem in jdbc, at the same time, since we don't just blindly unwrap possible errors in the jdbc driver at that spot :)

nbagnard:
> By the way, I have no idea how this "works in jdbc" since we can't use jdbc with Power BI and whatever was used to test it was likely doing different things. All I did was change the error handling here, but Power BI now works and correctly shows the available collection. This means that Power BI is probably doing some global listing and falling back to a local one when the global listing fails! So we really can't be sure that jdbc "works" here, since it's being used differently

No, it wasn't doing different things. I reproduced the issue on PowerBI, looked at the SQL trace and 
I had the explanation in the JIRA ticket : "the ODBC driver fails when using SQLTablesW to list all accessible collections" => So to test with JDBC, you simply call getTables(null,"%","%",new String[]{"TABLE, VIEW"}) on the JDBC side with the same restricted users.

The issue is not the unwrap itself, it is that we are passing in empty database names.

pmeredit:
> This idea is not to silently ignore errors happening when trying to retrieve the list of collections for a given database. As I explained in the ticket, we need to filter out empty database names. You can do this by adding "" to DISALLOWED_DB_NAMES or adding `.filter(|&db_name| !db_name.is_empty())` in the filter list again. I look at the Git blame history and this was done in [SQL-1440 PR](https://github.com/mongodb/mongo-odbc-driver/pull/136) but unfortunately broken in [SQL-2288 PR](https://github.com/mongodb/mongo-odbc-driver/pull/251). This is a regression and we didn't catch it, which means that we also need to add a test with an ADF user having the minimal permission set. Also, when I do a quick test, I see that run_command returns a Result : "called `Result::unwrap()` on an `Err` value: Error { kind: Command(CommandError { code: 17, code_name: "ProtocolError", message: ""$db" must not be empty, correlationID = 18548b06a8434559d5fe3ad3", topology_version: None }), labels: {}, wire_version: Some(25), source: None }". We should find a way to get that information back to the user.

This is specifically not "silently ignoring errors".

bucaojit (approved):
Updates look good to me.  

In the JDBC driver we have [this function](https://github.com/mongodb/mongo-jdbc-driver/blob/7dbaf15e0b93b55fab0c3f76a6d0ec32ada38b4c/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java#L409) that will filter out the empty database values.

nbagnard (approved):
LGTM! 👍

PR #345, opened 2025-07-25
SQL-2896: Filter out schema documents with no schema field

nbagnard:


core/src/query.rs
```diff
@@ -88,7 +88,9 @@ impl MongoQuery {
                     "_id": {
                         "$in": &collection_names
                         }
-                    }
+                    },
+                    // Only retrieve documents with a schema.
+                    "schema": { $exists: true }
```
pmeredit, line 93:
```suggestion
                    "schema": { "$exists": true }
```

PR #351, opened 2025-09-08
SQL-2805: update odbc repo to use new shared test infra funcs

jchemburkar:


pmeredit (approved):
SLGTM

nbagnard (changes requested):
This is probably a bug in the shared test infra code, but the SAST report is now at https://translators-connectors-releases.s3.amazonaws.com/mongosql-odbc-driver/artifacts/68c0e0fd71c4cc00081b12b6/ssdlc/mongo-odbc-driver.sast.sarif instead of https://translators-connectors-releases.s3.amazonaws.com/mongosql-odbc-driver/mongo-odbc-snapshot.sast.sarif.

The links in the compliance report are also not correct and point to non-existing files.

PR #372, opened 2026-03-13
SQL-3124: DNS resolver determination only happens once

nbagnard:
Currently, on Windows, the DNS resolver determination happens for every SQLDriverConnect calls because parseURI also tries to resolve host names and we added a fallback mechanism if Cloudflare is not usable.
This changes stashes the DNS resolver to use after the first connection and applies the related config to subsequent connections.

### DIFF
--- a/core/src/odbc_uri.rs
+++ b/core/src/odbc_uri.rs
@@ -78,7 +78,9 @@ lazy_static! {
 
 // Cache for the resolver configuration to use on Windows.
 // This is calculated once on first use and reused thereafter.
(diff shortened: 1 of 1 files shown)


bucaojit (approved):
Changes look good.  I had a few nits and one suggestion that can be added as a future task if it seems helpful.

EthanHardyMongo (approved):
I just had that one idea, but otherwise, LGTM!

bucaojit (approved):
Updates look good 👍

nbagnard:
evergreen retry

PR #373, opened 2026-03-20
SQL-3103: Fix issue with catalogs when user keeps fetching after receiving SQL_NO_DATA

nbagnard:
Guarding bounds to avoid panicking when users are not following ODBC flow and keep fetching after exhausting the resultset.


core/src/databases.rs
```diff
@@ -243,6 +243,10 @@ impl MongoStatement for MongoDatabases {
     // Increment current_db_index.
     // Return true if current_db_index index is <= for databases_names.length.
     fn next(&mut self, _: Option<&MongoConnection>) -> Result<(bool, Vec<Error>)> {
+        // Cursor was already exhausted
+        if self.current_db_index >= self.database_names.len() {
+            return Ok((false, vec![]));
+        }
         self.current_db_index += 1;
         Ok((self.current_db_index <= self.database_names.len(), vec![]))
```
EthanHardyMongo, line 252: I don't think this is adding any new behavior. If `self.current_db_index == self.database_names.len()` and then you add 1, the last part would return `Ok((false, vec![]))`.

core/src/mock_query.rs
EthanHardyMongo: Same thing here. We're already returning `Ok((false, vec![]))` if `self.current >= self.resultset.len()`

core/src/table_types.rs
EthanHardyMongo, line 41: you can remove this too.

core/src/type_info.rs
```diff
@@ -95,6 +95,10 @@ impl MongoStatement for MongoTypesInfo {
     // iterate through the list, searching for the next "valid" data type.
     // a type is valid if its sql type matches the desired sql type, or if we are getting all types.
     fn next(&mut self, _: Option<&MongoConnection>) -> Result<(bool, Vec<Error>)> {
+        // Cursor was already exhausted
+        if self.current_type_index > DATA_TYPES.len() {
+            return Ok((false, vec![]));
+        }
         loop {
```
EthanHardyMongo, line 103: you can remove this too

EthanHardyMongo (changes requested):
I'm pretty sure you can remove all the checks except the one added to `collections.rs`. Also, it looks like the original error happened in `collections.rs`:

```
ERROR: [Env_0x13bf5080][Conn_0x13bf5920][Stmt_0x13bf4220] SQLFetch:: [MongoDB][API] Caught panic: called Option::unwrap() on a None value

Ok(“in file ‘core\src\collections.rs’ at line 230”)
```

Edit: we need the extra checks to protect us from an overflow

core/src/databases.rs
nbagnard: I was torn about it too.
I went for adding the guard at the very beginning of the function because it seems useless to keep incrementing the index once we are past the boundary.
But I didn't find a more elegant way to do so without the repeated code.
The portion I added covers return "False" after "False" was already returned once.

EthanHardyMongo (approved):
Approving since we actually do need the extra checks

core/src/collections.rs
```diff
@@ -216,15 +216,18 @@ impl MongoStatement for MongoCollections {
...
-        if self.current_collection_index.is_none() {
-            self.current_collection_index = Some(0);
+        // Initialize database index on first call or retrieve current index
+        let db_index = *self.current_database_index.get_or_insert(0);
+        // Cursor was already exhausted
+        if db_index >= self.collections_for_db_list.len() {
+            return Ok((false, vec![]));
         }
+        // Initialize collection index on first call or retrieve current index
+        self.current_collection_index.get_or_insert(0);
```
jpowell-mongo, line 230: Question (non-blocking): Why do we reset this value a second time?

jpowell-mongo (approved):
Left a question but otherwise looks good.

core/src/collections.rs
nbagnard: @jpowell-mongo What do you mean by a second time, I see it being initialized only once?
We first initialize the DBs index, and then the collections index.

jpowell-mongo: Ah, you're right - I confused the variable names. I think my mind saw "current" at the beginning and "index" at the end and short-circuited - thinking it was the same variable

PR #375, opened 2026-03-27
SQL-3148: handle flaky odbc integration tests

jchemburkar:
There are occasional segfaults in the integration_test task of the periodic build of the driver. It seems that it could be because we are quickly disconnecting and freeing a handle, resulting in a race condition. This PR adds a proposed safety improvement for that operation.

The bigger issue, however, it it seems the Windows Error Reporting software (WER) operates in an order such that the rust test environment will catch the segfault before it returns a crash dump. This PR uses procdump + compiles the tests to an executable in order to capture at the error at the "first-chance" level, before the rust runtime has a chance to handle it. 

[Here](https://spruce.corp.mongodb.com/version/69c5d2e52fffbf0007c06454/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) is an evergreen patch where an integration test that intentionally causes a segfault fails and produces a .dmp file.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -964,7 +964,7 @@ functions:
           echo "----- Downloading Procdump ----"
           PROCDUMP_ZIP="$DUMP_PATH/procdump.zip"
           PROCDUMP_DIR="$DUMP_PATH/procdump"
-          curl -L -o "$PROCDUMP_ZIP" "https://download.sysinternals.com/files/Procdump.zip"
+          curl -f -L --retry 3 -o "$PROCDUMP_ZIP" "https://download.sysinternals.com/files/Procdump.zip"
           if [ $? -eq 0 ]; then
             mkdir -p "$PROCDUMP_DIR"
             unzip -o "$PROCDUMP_ZIP" -d "$PROCDUMP_DIR"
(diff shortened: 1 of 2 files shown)


evergreen.yml
```diff
@@ -955,13 +955,50 @@ functions:
...
+          mkdir -p $DUMP_PATH
           set +e
+
+          # Download and install procdump as JIT debugger for crash dump collection.
+          # Procdump captures first-chance exceptions, which is needed because Rust's
+          # vectored exception handler catches access violations before WER can see them.
+          echo "----- Downloading Procdump ----"
+          PROCDUMP_ZIP="$DUMP_PATH/procdump.zip"
+          PROCDUMP_DIR="$DUMP_PATH/procdump"
+          curl -L -o "$PROCDUMP_ZIP" "https://download.sysinternals.com/files/Procdump.zip"
```
bucaojit:
```suggestion
          curl -f -L --retry 3 -o "$PROCDUMP_ZIP" "https://download.sysinternals.com/files/Procdump.zip"
```
To have it exit non-zero if it can't access the file add the `-f`.  And it wouldn't hurt to put a retry.

bucaojit (approved):
Looks good.  Had one minor comment.

nbagnard (approved):
LGTM!
This got a lot more involved than what I thought it could be 😧

PR #376, opened 2026-04-03
SQL-3077: Add support for SQL_DATA_SOURCE_NAME attribute

bucaojit:
Adds SQL_DATA_SOURCE_NAME support to SQLGetInfo, returning the DSN name when connected via `DSN=` or an empty string when connected with `Driver=`.  

Stores the DSN value at connect time in ConnectionAttributes so it can be retrieved later.

Added unit test and tested manually with `pyodbc` on Windows to confirm it retrieves the data source name with `getinfo()`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2808,6 +2808,7 @@ dependencies = [
  "once_cell",
  "open",
  "openidconnect",
+ "rand 0.9.2",
  "regex",
  "reqwest",
  "rfc8252_http_server",
--- a/odbc/src/api/get_info_tests.rs
+++ b/odbc/src/api/get_info_tests.rs
@@ -780,4 +780,50 @@ mod unit {
         expected_value = u16::MAX,
         actual_value_modifier = modify_u16_value,
     );
+
+    test_get_info!(
+        data_source_name_no_dsn,
+        info_type = InfoType::SQL_DATA_SOURCE_NAME as u16,
+        expected_sql_return = SqlReturn::SUCCESS,
+        buffer_length = size_of::<WideChar>() as i16,
+        expected_length = 0,
+        expected_value = "",
+        actual_value_modifier = modify_string_value,
+    );
+
+    #[test]
+    fn data_source_name_with_dsn() {
+        unsafe {
+            let dsn_value = "MY_DSN";
+            let mut conn =
+                Connection::with_state(std::ptr::null_mut(), ConnectionState::Connected);
+            conn.attributes.write().unwrap().dsn = Some(dsn_value.to_string());
+            let mongo_handle: *mut _ = &mut MongoHandle::Connection(conn);
+
+            let value_ptr: *mut std::ffi::c_void =
+                Box::into_raw(Box::new([0u8; 900])) as *mut _;
+            let out_length: *mut SmallInt = &mut 0;
+            let buffer_length: SmallInt =
+                (dsn_value.len() + 1) as i16 * size_of::<WideChar>() as i16;
+
+            assert_eq!(
+                SqlReturn::SUCCESS,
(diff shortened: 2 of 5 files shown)


odbc/src/api/get_info_tests.rs
```diff
@@ -780,4 +780,51 @@ mod unit {
...
+        data_source_name_no_dsn,
+        info_type = InfoType::SQL_DATA_SOURCE_NAME as u16,
+        expected_sql_return = SqlReturn::SUCCESS,
+        buffer_length = size_of::<WideChar>() as i16,
+        expected_length = 0,
+        expected_value = "",
+        actual_value_modifier = modify_string_value,
+    );
+
+    #[test]
```
jchemburkar, line 794: I trust that this works, but maybe we can also add a unit test verifying that parsing hte odbc_uri correctly sets teh `dsn` on the `conn_attrs`, either here or in a separate unit test.

jchemburkar (approved):
lgtm -- just one though on tests

odbc/src/api/get_info_tests.rs
bucaojit: Sure, I added a test for that.

odbc/src/handles/definitions.rs
```diff
@@ -323,6 +323,9 @@ pub struct ConnectionAttributes {
     // to wait for any operation on a connection to timeout (other than
     // initial login).
     pub connection_timeout: Option<u32>,
+    // SQL_DATA_SOURCE_NAME: the DSN used to connect, if any.
+    // Empty string if the connection was made with DRIVER= instead of DSN=.
```
nicholascioli, line 327: Small nit: Is there any reason why these documentation comments are not using `///`? I see that the other structs in this file also just use regular comments.

Cargo.lock
nicholascioli, line 1: Was this file supposed to be updated?

odbc/src/api/get_info_tests.rs
```diff
@@ -780,4 +780,51 @@ mod unit {
...
+    fn data_source_name_with_dsn() {
```
nicholascioli, line 795: Nit: Is there a reason the entire test needs to be `unsafe`? It makes it hard to quickly see which parts are actually unsafe (and thus warrant more review)

nicholascioli (approved):
LGTM! I left a few nits since I'm still new to this codebase.

odbc/src/handles/definitions.rs
bucaojit: There's some guidelines on comments [here](https://wiki.corp.mongodb.com/spaces/DRIVERS/pages/260909864/SQL+Engines+Coding+Guildines#SQLEnginesCodingGuildines-Functionlevelcomments).  I don't think we've been very strict with `//` vs `///`.  I was matching the surrounding code.

Cargo.lock
bucaojit: Yeah, it generally gets updated with PRs you can see the history of the file.

odbc/src/api/get_info_tests.rs
bucaojit: It didn't seem necessary for this test but I'm open to hearing how it would add value here.  Are the unsafe sections requiring more review?  I did try adding more unsafe blocks but it looked like it was adding more noise and harder to review to me.  

The pattern for other tests are also to not separate out the specific unsafe parts.

nicholascioli: Yeah, in general it helps to try to scope the unsafe block as much as possible to make it clear to reviewers which sections need more care. This is not to say that only unsafe code goes in unsafe blocks, but that code which together is unsafe is marked as so.

PR #379, opened 2026-05-04
SQL-3221, SQL-3206: Inject dummy usernames/passwords into URI's that require them to prevent `must be URL encoded` error

EthanHardyMongo:
I manually tested the change, and it fixed the bug.

### DIFF
--- a/core/src/odbc_uri.rs
+++ b/core/src/odbc_uri.rs
@@ -649,12 +649,44 @@ mod unit {
         }
 
         #[test]
+        fn test_scram_sha1_specified_with_username_in_uri_is_unmodified() {
+            use crate::odbc_uri::ODBCUri;
+            let uri = "mongodb://foo:@localhost:27017/abc?authMechanism=SCRAM-SHA-1";
+            let mut odbc_uri = ODBCUri::new(format!("URI={uri};User=foo;PWD=bar")).unwrap();
+            assert_eq!(odbc_uri.construct_uri_for_parsing(uri).unwrap(), uri);
+        }
+
+        #[test]
+        fn test_scram_sha1_specified_with_username_and_password_in_uri_is_unmodified() {
+            use crate::odbc_uri::ODBCUri;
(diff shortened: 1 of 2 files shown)


core/src/odbc_uri.rs
```diff
@@ -502,15 +502,13 @@ impl ODBCUri {
...
-                uri.to_string()
-            },
-        )
+
+        // We inject a dummy username and password, so that this URI works with `ClientOptions::parse(uri)`.
+        // If there was no username and password in the URI, the parse function would fail. Additionally, we
+        // don't inject the actual username and password because that will be handled by `finalize_client_options` later.
+        // Also, if we inject the actual username and password here, we risk putting a reserved character in the URI, which
+        // will cause a `must be URL encoded` error.
+        Ok(format!("{protocol}dummy_username:dummy_password@{rest}"))
```
EthanHardyMongo, line 511: Natacha and I decided to use a dummy username and password to make it clear that this function is only adding things to the URI to prevent parse from failing. We could change all reserved characters to their URL encodings; however, we're going to set the username and password again in `finalize_client_options` anyway, so there's no benefit to doing that.

core/src/odbc_uri.rs
```diff
@@ -645,7 +643,7 @@ mod unit {
         fn test_scram_sha1_specified() {
```
nbagnard, line 643: todo:
Add tests showing that if there is a username or password no matter which auth mech is selected, the URI is unchanged.

nbagnard (approved):
I just have a small request for an extra test, otherwise it looks good to me 👍

jchemburkar (approved):
lgtm

EthanHardyMongo:
The failing SBOM is due to the rust driver. Until they release an updated version with the `hickory-proto` and `hickory-resolver` dependencies updated, the SBOM will fail. I saw in the rust channel that they are working on it in this thread: https://mongodb.slack.com/archives/CP893U9MX/p1777650307867869

PR #380, opened 2026-05-11
[SQL-3186][SQL-3198] Quality Of Life - ODBC Connection String Improvements

jpowell-mongo:
## Summary

A collection of small fixes to make the ODBC Driver more ergonomic. When we introduced the odbc-connectivity-tester during the troubleshooting epic, we found that we had to include username and password in the generated ODBC URL even when the auth mechanism did not require it.

**Fix 1:**  UID and PWD should not be required as part of the URL if the auth mechaism does not require it.
This is the case for OIDC and X509

```
DRIVER={MongoDB Atlas SQL ODBC Driver};URI=<clusterUrl>?tlsCertificateKeyFile=../../client.pem&tls=true&authSource=%24external&authMechanism=MONGODB-X509;DATABASE=test;UID=;PWD=;
```

And this for OIDC:
```
DRIVER={MongoDB Atlas SQL ODBC Driver};URI=<clusterUrl>/?directConnection=true&connectTimeoutMS=60000&tls=true&authMechanism=MONGODB-OIDC;DATABASE=airbnb;UID=jonathan.powell@mongodb.com;PWD=;
```

We've updated the code to ignore uid + pwd checks when these auth mechanisms are used because they validate this info externally.

**Fix 2:** Do not require username + password be included when using Kerberos

Kerberos Auth already requires user to specify username + password in the URL so we shouldn't require it to be included again.

## Testing

- Unit Tests in the ODBC Driver

**Connectivity Tester**
- [ ] Manual testing of OIDC to confirm UID + PWD are not required
- [ ] Manual testing of X509 to confirm UID + PWD are not required
- [ ] Manual testing of Kerberos to confirm UID + PWD are not required

### DIFF
--- a/core/src/odbc_uri.rs
+++ b/core/src/odbc_uri.rs
@@ -1039,10 +1039,7 @@ mod unit {
 
     #[cfg(test)]
     mod try_into_client_options {
-        use crate::odbc_uri::{DATABASE, PWD, UID, URI};
         use mongodb::options::ClientOptions;
-        use num_traits::One;
-        use openidconnect::http::Uri;
 
         #[tokio::test(flavor = "current_thread")]
         async fn username_required_when_auth_mechanism_environment_is_azure() {
--- a/core/src/odbc_uri.rs
+++ b/core/src/odbc_uri.rs
@@ -427,8 +427,8 @@ impl ODBCUri {
         // X509 and Kerberos (GSSAPI) do not require username and password as part of client options.
         // X509 does not specify username/pass in the URI because auth is done with a client cert
         // GSSAPI specifies username/pass in the uri, so it doesn't need to be specified in attributes.
-        if (auth_mechanism == &AuthMechanism::MongoDbX509
-            || auth_mechanism == &AuthMechanism::Gssapi)
+        if auth_mechanism == &AuthMechanism::MongoDbX509
+            || auth_mechanism == &AuthMechanism::Gssapi
         {
             return Ok(());
         }
@@ -619,15 +619,15 @@ impl ODBCUri {
                 self.remove(&[APPNAME]),
                 app_name,
             ]
-                .into_iter()
-                .flatten()
-                .fold(String::new(), |acc, x| {
-                    if acc.is_empty() {
-                        x
-                    } else {
-                        format!("{acc}|{x}")
-                    }
-                }),
+            .into_iter()
+            .flatten()
+            .fold(String::new(), |acc, x| {
+                if acc.is_empty() {
+                    x
+                } else {
+                    format!("{acc}|{x}")
+                }
+            }),
         )
     }
 }
@@ -1203,11 +1203,11 @@ mod unit {
             let opts = ODBCUri::new(
                 "USER=foo2;PWD=bar2;URI=mongodb://foo:bar@127.0.0.1:27017".to_string(),
             )
-                .unwrap()
-                .try_into_client_options()
-                .await
-                .unwrap()
-                .client_options;
+            .unwrap()
+            .try_into_client_options()
+            .await
+            .unwrap()
+            .client_options;
             let cred = opts.credential.unwrap();
             assert_eq!(expected_cred.username, cred.username);
             assert_eq!(expected_cred.password, cred.password);
@@ -1329,11 +1329,11 @@ mod unit {
             let opts = ODBCUri::new(
                 "SERVER=127.0.0.2:27017;URI=mongodb://foo:bar@127.0.0.1:27017".to_string(),
             )
-                .unwrap()
-                .try_into_client_options()
-                .await
-                .unwrap()
-                .client_options;
+            .unwrap()
+            .try_into_client_options()
+            .await
+            .unwrap()
+            .client_options;
             assert_eq!(expected_opts.hosts[0], opts.hosts[0]);
         }
 
@@ -1344,8 +1344,8 @@ mod unit {
             let expected_opts = ClientOptions::parse(
                 "mongodb://foo:bar@www.atlas.net:27017/?authSource=authDB&ssl=true",
             )
-                .await
-                .unwrap();
+            .await
+            .unwrap();
             let expected_cred = expected_opts.credential.unwrap();
             let opts = ODBCUri::new("UID=foo;PWD=bar;SERVER=www.atlas.net:27017;User=foo2;Password=bar2;uri=mongodb://localhost:29000/?authSource=authDB&ssl=true".to_string())
                 .unwrap()
@@ -1379,10 +1379,10 @@ mod unit {
                 "USER=foo;PWD=bar;SERVER=localhost:27017;APPNAME=powerbi-connector+1.0.0"
                     .to_string(),
             )
-                .unwrap()
-                .try_into_client_options()
-                .await
-                .unwrap();
+            .unwrap()
+            .try_into_client_options()
+            .await
+            .unwrap();
             assert_eq!(
                 format!("{}|powerbi-connector+1.0.0", *DEFAULT_APP_NAME),
                 uri_opts.client_options.app_name.unwrap()
@@ -1409,10 +1409,10 @@ mod unit {
                 "USER=foo;PWD=bar;URI=mongodb://localhost:27017/;APPNAME=powerbi-connector+1.0.0"
                     .to_string(),
             )
-                .unwrap()
-                .try_into_client_options()
-                .await
-                .unwrap();
+            .unwrap()
+            .try_into_client_options()
+            .await
+            .unwrap();
 
             assert_eq!(
                 format!("{}|powerbi-connector+1.0.0", *DEFAULT_APP_NAME),
@@ -1464,10 +1464,10 @@ mod unit {
             let uri_opts = ODBCUri::new(
                 "USER=foo;PWD=bar;URI=mongodb://localhost:27017/?appName=foo;".to_string(),
             )
-                .unwrap()
-                .try_into_client_options()
-                .await
-                .unwrap();
+            .unwrap()
+            .try_into_client_options()
+            .await
+            .unwrap();
 
             assert_eq!(
                 DRIVER_SHORT_NAME,
--- a/core/src/odbc_uri.rs
+++ b/core/src/odbc_uri.rs
@@ -414,6 +414,25 @@ impl ODBCUri {
     }
 
     fn check_client_opts_credentials(client_options: &ClientOptions) -> Result<()> {
+        // Kerberos (GSSAPI) Doesn't require user + pass as USER; PASS; if specified in the URI
+        // X509 should not require user or pass either
+        let auth_mechanism = client_options
+            .credential
+            .as_ref()
+            .unwrap()
+            .mechanism
+            .as_ref()
+            .unwrap();
+
+        // X509 and Kerberos (GSSAPI) do not require username and password as part of client options.
+        // X509 does not specify username/pass in the URI because auth is done with a client cert
+        // GSSAPI specifies username/pass in the uri, so it doesn't need to be specified in attributes.
+        if (auth_mechanism == &AuthMechanism::MongoDbX509
+            || auth_mechanism == &AuthMechanism::Gssapi)
+        {
+            return Ok(());
+        }
+
         if client_options
             .credential
             .as_ref()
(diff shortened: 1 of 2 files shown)


jpowell-mongo:
Moving to draft to debug tests

core/src/odbc_uri.rs
```diff
@@ -414,16 +415,45 @@ impl ODBCUri {
     }
 
     fn check_client_opts_credentials(client_options: &ClientOptions) -> Result<()> {
+        let auth_mechanism = client_options
+            .credential
+            .as_ref()
+            .unwrap()
```
bucaojit: I see this part repeated in the code and I'm thinking it could be separated like this.  And in lines 435, 444:
```
      let cred = client_options
          .credential
          .as_ref()
          .expect("credential must be set before validation");
      let auth_mechanism = cred.mechanism.as_ref().unwrap_or(&AuthMechanism::ScramSha256);
```

core/src/odbc_uri.rs
```diff
@@ -999,6 +1041,103 @@ mod unit {
...
+                .try_into_client_options()
+                .await
+                .is_ok());
+        }
+
+        #[tokio::test(flavor = "current_thread")]
+        async fn oidc_with_azure_environment_and_username_succeeds() {
+            use crate::odbc_uri::ODBCUri;
+            let odbc_uri = "DRIVER={MongoDB Atlas SQL ODBC Driver};\
+         URI=mongodb://myuser@cluster.example.net/?authMechanism=MONGODB-OIDC\
```
bucaojit: Is the username needed in both the URI and UID here?

bucaojit (changes requested):

core/src/odbc_uri.rs
jpowell-mongo: Based on the ticket description, the request was to require that the UID be included if the environment is Azure. 

I'll add an extra test case where we only include it via the URI. I'd have to confirm with Natacha on the behavior she wants.

core/src/odbc_uri.rs
jpowell-mongo: Good call! I'll update this shortly.

jpowell-mongo: Addressed here:
https://github.com/mongodb/mongo-odbc-driver/pull/380/changes/3564740d6fb433a5c970578c98be3348569e0ffc

core/src/odbc_uri.rs
jpowell-mongo: I added a unit test to show that you can pass the username in the URI and that will succeed:
https://github.com/mongodb/mongo-odbc-driver/pull/380/changes/a736cec8ed8d86200350e18c17033d4d2c67d7a3

core/src/odbc_uri.rs
```diff
@@ -414,27 +415,45 @@ impl ODBCUri {
...
+        // X509 does not specify username/pass in the URI because auth is done with a client cert
+        // GSSAPI specifies username/pass in the uri, so it doesn't need to be specified in attributes.
+        if auth_mechanism == &AuthMechanism::MongoDbX509 || auth_mechanism == &AuthMechanism::Gssapi
         {
-            return Err(Error::InvalidUriFormat(format!(
-                "One of {PWD_KWS:?} is required for a valid Mongo ODBC Uri"
-            )));
+            return Ok(());
+        }
+
```
bucaojit, line 434: nit: This variable can be declared in the following `if()` in line 451 since it's used there instead of outside of it.  
(This was supposed to be part of my previous comments but git didn't add it 🤷 , it's for the `is_oidc_and_azure_environment` variable)

jchemburkar (approved):
pending green waterfall, changes look good to me. Thanks for added comments!

core/src/odbc_uri.rs
```diff
@@ -414,27 +415,45 @@ impl ODBCUri {
...
+                .mechanism_properties
+                .as_ref()
+                .and_then(|props| props.get("ENVIRONMENT"))
+                .is_some_and(|v| v == &Bson::String("azure".to_string()));
+
+        if client_credentials.username.is_none() {
+            // OIDC only requires username when Azure is the environment.
+            if is_oidc_and_azure_environment || auth_mechanism != &AuthMechanism::MongoDbOidc {
+                return Err(Error::InvalidUriFormat(format!(
+                    "One of {USER_KWS:?} is required for a valid Mongo ODBC Uri"
```
nbagnard: todo:
Let's make the error clearer for the user.


```suggestion
                    "One of {USER_KWS:?} is required for a valid Mongo ODBC Uri when using an Azure managed identity. Set this to the client ID of the managed identity or the application ID of the service principal."
```

core/src/odbc_uri.rs
```diff
@@ -414,27 +415,45 @@ impl ODBCUri {
...
+                )));
+            }
+        }
+        if client_credentials.password.is_none() {
+            // OIDC doesn't require password either.
+            if auth_mechanism != &AuthMechanism::MongoDbOidc {
+                return Err(Error::InvalidUriFormat(format!(
+                    "One of {PWD_KWS:?} is required for a valid Mongo ODBC Uri"
```
nbagnard: suggestion:
Do you think we could also add the authentication mechanism in a user friendly form.


```suggestion
                    "One of {PWD_KWS:?} is required for a valid Mongo ODBC Uri when the authication mechanism is {auth_mechanism}"
```

core/src/odbc_uri.rs
```diff
@@ -695,6 +714,15 @@ mod unit {
             let mut odbc_uri = ODBCUri::new(format!("URI={uri};User=foo;PWD=bar")).unwrap();
             assert_eq!(odbc_uri.construct_uri_for_parsing(uri).unwrap(), uri);
         }
+
+        #[test]
+        fn test_gssapi_does_not_modify_uri() {
+            use crate::odbc_uri::ODBCUri;
+            let uri =
+                "mongodb://alice@CORP.EXAMPLE.COM:s3cr3t@mongo.corp.example.com/?authMechanism=GSSAPI&authSource=$external";
+            let mut odbc_uri = ODBCUri::new(format!("URI={uri}")).unwrap();
```
nbagnard: issue:
This test is a no-op. There is no UID/User or PWD/password in the odbc URI.
```suggestion
            let mut odbc_uri = ODBCUri::new(format!("URI={uri};User=foo;PWD=bar")).unwrap();
```

core/src/odbc_uri.rs
```diff
@@ -999,6 +1027,118 @@ mod unit {
...
+                    "{}",
+                    ODBCUri::new(odbc_uri.to_string())
+                        .unwrap()
+                        .try_into_client_options()
+                        .await
+                        .unwrap_err()
+                )
+            );
+        }
+        #[tokio::test(flavor = "current_thread")]
```
nbagnard, line 1079: todo:
Use a macro or helper function that just need the URI and remove the duplicated code ;)

core/src/odbc_uri.rs
```diff
@@ -414,27 +415,45 @@ impl ODBCUri {
...
+        let is_oidc_and_azure_environment = auth_mechanism == &AuthMechanism::MongoDbOidc
+            && client_credentials
+                .mechanism_properties
+                .as_ref()
+                .and_then(|props| props.get("ENVIRONMENT"))
+                .is_some_and(|v| v == &Bson::String("azure".to_string()));
+
+        if client_credentials.username.is_none() {
+            // OIDC only requires username when Azure is the environment.
+            if is_oidc_and_azure_environment || auth_mechanism != &AuthMechanism::MongoDbOidc {
```
nbagnard, line 444: question:
Surprisingly the Rust driver doesn't enforce this. Do you know if this is in the spec or just the public doc?
Can an OIDC Azure URI be valid without a username or is it for sure always required?

nbagnard (changes requested):
I noticed a few small things, but the change itself to make the connection handling clearer is looking good!

core/src/odbc_uri.rs
jpowell-mongo: Addressed in:
https://github.com/mongodb/mongo-odbc-driver/pull/380/changes/e11a926a152e3b69aa8af5d81a89dcd47b07e0de

I moved it inside the if-statement.

core/src/odbc_uri.rs
jpowell-mongo: Updated the comment here:
https://github.com/mongodb/mongo-odbc-driver/pull/380/changes/e11a926a152e3b69aa8af5d81a89dcd47b07e0de

core/src/odbc_uri.rs
jpowell-mongo: Updated the error message here to match your suggestion:
https://github.com/mongodb/mongo-odbc-driver/pull/380/changes/e11a926a152e3b69aa8af5d81a89dcd47b07e0de

core/src/odbc_uri.rs
jpowell-mongo: Reading the rust drive code, I didn't see any enforcement on Azure/Username either.

After reading the [spec in the Rust Driver Repo](https://github.com/mongodb/mongo-rust-driver/blob/main/spec/auth/mongodb-oidc.md#5-azure-tests), it seems that username is not always required for Azure. This is what their section on testing says:

```markdown
## (5) Azure Tests

Drivers MUST only run the Azure tests when testing on an Azure VM. See instructions in
[Drivers Evergreen Tools](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/auth_oidc/azure/README.md)
for test setup.

# 5.1 Azure With No Username

- Create an OIDC configured client with `ENVIRONMENT:azure` and a valid `TOKEN_RESOURCE` and no username.
- Perform a `find` operation that succeeds.
- Close the client.

# 5.2 Azure with Bad Username

- Create an OIDC configured client with `ENVIRONMENT:azure` and a valid `TOKEN_RESOURCE` and a username of `"bad"`.
- Perform a `find` operation that fails.
- Close the client.
```

I think the answer is that it's optional, but will need some confirmation. We might not need the additional oidc + azure check?

core/src/odbc_uri.rs
jpowell-mongo: Updated here:
https://github.com/mongodb/mongo-odbc-driver/pull/380/changes/2c8fc277d2cd6218858b58837acdb3efca61be52

jpowell-mongo:
I've addressed the comments, the outstanding question is about the logic for OIDC and Azure environment!

bucaojit (approved):
LGTM

core/src/odbc_uri.rs
nbagnard: If we don't need it or are unsure, let's not error on it. You can add a warning in the log which could help us track down error later on if necessary.

core/src/odbc_uri.rs
nbagnard: @jpowell-mongo I'll send you a patch via Slack. You'll see what I had in mind.

nbagnard (changes requested):
I don't think that we should restrict the connection and fail if there is no username for OIDC Azure until we have confirmation that this is indeed mandatory.
If it is possible to connect without a user name and this is the only conf a user can use, then we will need to remove this validation and re-release the driver to unblock the user.
A warning seems safer for now.

jpowell-mongo:
> I don't think that we should restrict the connection and fail if there is no username for OIDC Azure until we have confirmation that this is indeed mandatory. If it is possible to connect without a user name and this is the only conf a user can use, then we will need to remove this validation and re-release the driver to unblock the user. A warning seems safer for now.

I addressed your comments in these commits.

This one https://github.com/mongodb/mongo-odbc-driver/pull/380/changes/b33f4d65debe4e17323ac3e6e67fae1c3a442e82 added the assert_parse_ok, and assert_parse_err macros and refactored the tests to use them.

This one updated the username check to be a warning:
https://github.com/mongodb/mongo-odbc-driver/pull/380/changes/f0e5a77dd72c5844b6a8ec505895c9fac28bab7c

And this test removed unit tests related to mandatory username for OIDC + Azure since we relaxed that requirement to just a warning:
https://github.com/mongodb/mongo-odbc-driver/pull/380/changes/c93513626d7ccb596d61bdeb36c3f09697ad1bed

core/src/odbc_uri.rs
```diff
@@ -414,27 +415,45 @@ impl ODBCUri {
...
+            let is_oidc_and_azure_environment = auth_mechanism == &AuthMechanism::MongoDbOidc
+                && client_credentials
+                    .mechanism_properties
+                    .as_ref()
+                    .and_then(|props| props.get("ENVIRONMENT"))
+                    .is_some_and(|v| v == &Bson::String("azure".to_string()));
+
+            if is_oidc_and_azure_environment || auth_mechanism != &AuthMechanism::MongoDbOidc {
+                log::warn!(
+                    "One of {USER_KWS:?} is required for a valid Mongo ODBC Uri when using an Azure managed identity. Set this to the client ID of the managed identity or the application ID of the service principal."
```
nbagnard:
```suggestion
                    "No {USER_KWS:?} provided with an Azure managed identity. If you are encountering connection issue, you might need to set it to the client ID of the managed identity or the application ID of the service principal."
```

nbagnard (approved):
LGTM!

Please just apply the last suggestion I added. I think it will help make sure users don't mistake a warning with an error and we also still clearly tell them what to do

PR #386, opened 2026-06-18
SQL-3281: Add NULL handling for strings as pattern value behavior

nbagnard:
Currently, the driver crashes if a NPE is passed for the catalog name argument.
This is a bug. We missed this parameter when implementing [SQL-2496](https://jira.mongodb.org/browse/SQL-2496). Likely because for SQLColumns, the catalog is not a search pattern and we assumed the same was true for SQLTables.
We are now supporting all 4 parameters as expected.
See https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/arguments-in-catalog-functions?view=sql-server-ver17:
SQLTables : CatalogName SchemaName TableName TableType => PV PV PV VL

PV = [Pattern Value Arguments](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/pattern-value-arguments?view=sql-server-ver17)
VL = [Value List Arguments](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/value-list-arguments?view=sql-server-ver17)


jpowell-mongo (approved):

cstr/src/lib.rs
```diff
@@ -124,15 +124,29 @@ pub unsafe fn input_text_to_string_w(text: *const WideChar, len: isize) -> Strin
...
+/// that is, a null pointer and the search pattern % (any characters) are equivalent.
+///
+/// If `pattern_value` is false, a null pointer is converted to "".
 ///
 /// # Safety
 /// This converts raw C-pointers to rust Strings, which requires unsafe operations
 ///
 #[allow(clippy::uninit_vec)]
-pub unsafe fn input_text_to_string_w_allow_null(text: *const WideChar, len: isize) -> String {
+pub unsafe fn input_text_to_string_w_allow_null(
```
jchemburkar, line 139: this will also affect SQLColumns right? I was looking and couldnt find any coverage for null in those functions, maybe we should add a quick one there to make sure this still works?

jchemburkar (approved):
lgtm pending one question

cstr/src/lib.rs
nbagnard: correct! Looks like we don't have tests for SQLColumns at all the covering the null behavior, I'll add them.

PR #395, opened 2026-08-07
SQL-3408: Update the download center feed to point to downloads.mongodb.org

nbagnard:
Update the download center feed to point to downloads.mongodb.org and stop appending new versions

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -1796,18 +1796,6 @@ functions:
         permissions: public-read
         content_type: application/json
 
-  "set mongodb.org url for download center feed":
-    - command: s3.put
-      params:
-        aws_key: ${release_aws_key}
-        aws_secret: ${release_aws_secret}
-        local_file: mongosql-odbc-driver/resources/download-center/mongo-odbc-downloads_mongo_org.json
-        remote_file: mongosql-odbc-driver/mongo-odbc-downloads.json
-        bucket: translators-connectors-releases
-        permissions: public-read
-        content_type: application/json
-
-
 tasks:
   - name: make-odbc-docs
     depends_on:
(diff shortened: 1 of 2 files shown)


evergreen.yml
```diff
@@ -1800,35 +1785,29 @@ functions:
...
         aws_key: ${release_aws_key}
         aws_secret: ${release_aws_secret}
-        local_file: mongosql-odbc-driver/resources/download-center/mongo-odbc-downloads_updated.json
+        local_file: mongosql-odbc-driver/resources/download-center/mongo-odbc-downloads_template.json
+        remote_file: mongosql-odbc-driver/mongo-odbc-downloads.json
+        bucket: translators-connectors-releases
+        permissions: public-read
+        content_type: application/json
+
+  "set mongodb.org url for download center feed":
```
nicholascioli, line 1799: Question: where is this task used? I don't think I see it in use anywhere.

nicholascioli (commented):
Looks good! I left a question about a potentially unused task?

mattChiaravalloti:
It looks like this will inherently address [SQL-3393](https://jira.mongodb.org/browse/SQL-3393) since we are no longer including old releases in the download feed, only the latest.

nbagnard:
> It looks like this will inherently address [SQL-3393](https://jira.mongodb.org/browse/SQL-3393) since we are no longer including old releases in the download feed, only the latest.

@mattChiaravalloti 
That's right. I'll close SQL-3393 as a duplicate.

evergreen.yml
nbagnard: Fixed

mattChiaravalloti (approved):

jpowell-mongo (approved):

PR #396, opened 2026-08-10
SQL-3340: validate Atlas SQL interface entitlement JWT at connect

EthanHardyMongo:
This PR adds the JWT logic to the ODBC driver.


core/src/entitlement.rs
```diff
@@ -0,0 +1,586 @@
...
+//! The signing side of this contract lives in the `sql-interface` CNCP service
+//! (10gen/mms PR #178521, SQL-3325); the token shape here must stay in lock-step with it.
+//!
+//! Scope notes:
+//! * This gate applies **only to Atlas dedicated clusters**. On-prem / self-managed Enterprise
+//!   deployments have no marker and must still be allowed to connect, so we detect an Atlas
+//!   dedicated host from `hello.me` and skip the check otherwise.
+//! * Fetching the JWKS public key is a separate ticket (task 28). It is stubbed here behind
+//!   [`Ed25519KeyProvider`]; [`StubKeyProvider`] is wired in for now and unit tests inject a
+//!   known key to exercise the full verification path.
```
EthanHardyMongo, line 17: `task 28` refers to the 28th ticket in [AtlasSQLDirectCluster_tickets.md](https://github.com/10gen/engineering-documents/blob/main/teams/sql-engines/SQL-3026-atlas-sql-dedicated-cluster/AtlasSQLDirectCluster_tickets.md). There's no ticket created for that task, so it's referred to by it's number in the design.

evergreen.yml
```diff
@@ -1420,16 +1420,64 @@ functions:
           $COMMON_TEST_INFRA_DIR/test-environment/run_adf.sh stop
           exit $EXITCODE
 
-  "run ubuntu mongosqltranslate integration tests":
+  "run ubuntu on prem integration tests":
```
EthanHardyMongo, line 1423: I renamed everything to `on prem` since `mongosqltranslate` is no longer used in the ODBC driver.

integration_test/tests/on_prem_tests.rs
```diff
@@ -14,29 +14,29 @@ mod mongosqltranslate_tests {
...
+    //             crate::common::generate_srv_style_connection_string(Some("test".to_string()));
+    //         let result = connect_with_conn_string(env_handle, Some(conn_str), true);
+    //
+    //         assert!(
+    //             result.is_ok(),
+    //             "Expected successful connection, got error: {result:?}"
+    //         );
+    //
+    //         let _ = unsafe { Box::from_raw(env_handle) };
+    //     }
```
EthanHardyMongo, line 33: I commented this out for now so that evergreen could pass, but I'm not sure what the best way to handle this is. Is it easy to set up an srv style connection string for on-prem MongoDB? If so, we could do that. Otherwise, we could disable this test for now, eventually get Alexi's cluster set up with the Atlas Dedicated SQL Interface, and then reenable it.

evergreen/set_up_on_prem_int_tests.sh
```diff
@@ -0,0 +1,51 @@
...
+  --silent \
+  --fail \
+  --max-time 600 \
+  --retry 5 \
+  --retry-delay 0
+}
+
+# Download the sample_airbnb data and schemas
+download "https://mongosql-noexpire.s3.us-east-2.amazonaws.com/ODBC_driver_test_data/AlexiCluster0SampleAirbnb.archive.gz"
+
```
EthanHardyMongo, line 17: All the `on-prem` tests use the `sample_airbnb` data.

core/src/conn.rs
```diff
@@ -171,7 +171,19 @@ impl MongoConnection {
...
+                // dedicated clusters; verify_sql_interface_entitlement skips the check for on-prem
+                // / self-managed Enterprise hosts (which have no marker). ADF is unaffected.
+                runtime.block_on(async {
+                    crate::entitlement::verify_sql_interface_entitlement(
+                        &client,
+                        &crate::entitlement::StubKeyProvider,
+                    )
+                    .await
+                })?;
+            }
```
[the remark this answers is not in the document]
EthanHardyMongo: this is done on purpose for now.

evergreen/set_up_on_prem_int_tests.sh
```diff
@@ -0,0 +1,51 @@
...
+# Download and extract the mongodb-database-tools package
+download "https://fastdl.mongodb.org/tools/db/$MONGO_TOOLS_DOWNLOAD.tgz"
+tar zxvf "$MONGO_TOOLS_DOWNLOAD".tgz
+
+# Download and extract mongosh
+download "https://downloads.mongodb.com/compass/$MONGOSH_DOWNLOAD.tgz"
+tar zxvf "$MONGOSH_DOWNLOAD".tgz
+
+# Create on prem test user
+"$MONGOSH_DOWNLOAD"/bin/mongosh admin --port 28017 --eval "db.createUser({user: '${on_prem_test_user}', pwd: '${on_prem_test_pwd}', roles: ['readWrite']})"
```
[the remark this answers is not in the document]
EthanHardyMongo: This doesn't matter because this user only exists so that the test connection string can connect to the db. This user isn't reading or writing anything.

core/src/entitlement.rs
```diff
@@ -0,0 +1,614 @@
...
+                    "unable to read the local SQL interface public key file '{}': {e}",
+                    path.display()
+                ))
+            });
+        }
+        let resp = self.http.get(&self.jwks_url).send().map_err(|e| {
+            not_entitled(format!(
+                "unable to fetch the SQL interface JWKS from the registry: {e}"
+            ))
+        })?;
```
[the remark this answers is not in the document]
EthanHardyMongo: fixed.

core/src/conn.rs
```diff
@@ -171,7 +171,19 @@ impl MongoConnection {
 
         let type_of_cluster = runtime.block_on(async { determine_cluster_type(&client).await })?;
         match type_of_cluster {
-            MongoClusterType::AtlasDataFederation | MongoClusterType::Enterprise => {}
+            MongoClusterType::AtlasDataFederation => {}
+            MongoClusterType::Enterprise => {
+                // SQL-3340: Atlas SQL Direct Cluster entitlement gate. This only applies to Atlas
```
jchemburkar, line 176: nit: why is the ticket name in these comments? we typically only do this for TODO work, but I assume this comment is settled once the ticket is done / we aren't coming back

core/src/entitlement.rs
```diff
@@ -0,0 +1,914 @@
+//! SQL-3340: Atlas SQL Direct Cluster entitlement verification.
```
jchemburkar, line 1: similar -- I think we should remove the ticket. Also think the comment structure `//!` is different than our typical `///`?

core/src/entitlement.rs
```diff
@@ -0,0 +1,914 @@
...
+//!
+//! Atlas SQL "Direct Cluster" is a paid, per-cluster toggle. When it is enabled, the Atlas
```
jchemburkar, line 3: I'm also wondering if we should remove some of the detail here -- this repo is public, do we want to spell out exactly how this works? might be a slight security risk?

core/src/entitlement.rs
```diff
@@ -0,0 +1,914 @@
...
+        ))
+    })?;
+    serde_json::from_slice(&bytes).map_err(|_| {
+        not_entitled(format!(
+            "the Atlas SQL interface entitlement token {which} is not valid JSON"
+        ))
+    })
+}
+
+#[cfg(test)]
```
jchemburkar, line 464: lets make tests their own file

core/src/entitlement.rs
```diff
@@ -0,0 +1,914 @@
...
+pub trait Ed25519KeyProvider {
+    fn verifying_key(&self, kid: &str) -> Result<VerifyingKey>;
+}
+
+/// Placeholder key provider used until task 28 (ODBC public-key retrieval) lands.
+///
+/// It always fails, so any real Atlas dedicated cluster with the feature enabled will currently be
+/// rejected with a clear message. That is acceptable while the feature is pre-GA and gated; the
+/// happy path is exercised in tests via an injected [`Ed25519KeyProvider`].
+pub struct StubKeyProvider;
```
jchemburkar, line 66: I think we should remove / implement this for real now that we have nicks endpoint name, per slack thread

integration_test/tests/on_prem_tests.rs
jchemburkar: I think you would have to set up the dns record yourself, but I'm not too sure how difficult it is. I think its up to you, if you look into it and it seems intensive, maybe add a ticket and make this ignored (rather than commented out)?

jchemburkar (changes requested):
A few thoughts, I can rereview when the jwks changes are in

core/src/entitlement.rs
```diff
@@ -0,0 +1,914 @@
...
+    let verifying_key = key_provider.verifying_key(kid)?;
+    let signature_bytes = URL_SAFE_NO_PAD.decode(sig_b64).map_err(|_| {
+        not_entitled("the Atlas SQL interface entitlement token signature is not valid base64url")
+    })?;
+    let signature = Signature::from_slice(&signature_bytes).map_err(|_| {
+        not_entitled("the Atlas SQL interface entitlement token signature has an invalid length")
+    })?;
+    let signing_input = format!("{header_b64}.{payload_b64}");
+    verifying_key
+        .verify(signing_input.as_bytes(), &signature)
```
jchemburkar, line 396: On the MSM side (10gen/schema-manager-rs#1012) we use `verify_strict` here instead of `verify`. It's the same accept/reject for legitimate tokens, but `verify_strict` additionally rejects small-order / non-canonical public keys, which closes a known Ed25519 malleability footgun. Since the marker is a security gate and the cost is zero for real keys, it seemed worth pinning. Might be worth matching here so the two drivers verify identically.

core/src/entitlement.rs
```diff
@@ -0,0 +1,914 @@
...
+        // Atlas host). The entitlement gate does not apply; allow the connection.
+        None => return Ok(()),
+    };
+
+    // 2. Read the singleton entitlement marker from the reserved collection.
+    let marker_coll = client
+        .database(MARKER_DB)
+        .collection::<mongodb::bson::Document>(MARKER_COLLECTION);
+    let marker = marker_coll
+        .find_one(doc! { "_id": MARKER_ID })
```
jchemburkar, line 311: One difference from the MSM implementation (10gen/schema-manager-rs#1012): we pin both this marker read and the `hello` above (line ~295) to the primary read preference. The reason is the disable path — after a customer toggles the feature OFF, a read served by a lagging secondary can still return the old enabled marker and wrongly allow the connection. Pinning primary on both reads closes that window. Might be worth doing the same here.

core/src/entitlement.rs
```diff
@@ -0,0 +1,914 @@
...
+/// Issuer (`iss`) for normal markers signed by the CNCP SQL Interface service.
+pub const ISSUER_NORMAL: &str = "mongosql-service";
+/// Issuer (`iss`) for break-glass markers signed by the restricted support tool (SQL-3337).
+pub const ISSUER_EMERGENCY: &str = "mongosql-emergency";
+
+/// JWS algorithm the marker is signed with. We pin this on the verify side and never trust the
+/// token's self-declared `alg` for anything else (classic JWT alg-confusion mitigation).
+const EXPECTED_ALG: &str = "EdDSA";
+
+const MARKER_DB: &str = "__sql_interface";
```
jchemburkar, line 39: noting that the marker db is `__mdb_internal_sqlinterface`. should be updated when incorporating the new mongosql work.

integration_test/tests/on_prem_tests.rs
nbagnard, line 1: suggestion:
The integration tests are meant to validate that the driver works well for direct cluster connection.
This could be on-prem or Atlas. That's actually why they were named "mongosqltranslate".
Did you update the tests to use an on-prem deployment is because you couldn't connect to Alexi's cluster anymore?
If so, I think we can either enable it if Alexi is Ok with it or "manually" add a marker for it ourselves (to avoid billing cost).

integration_test/tests/on_prem_tests.rs
nbagnard: ah, this is what I was referring to above. We could enable the SQL interface on the existing Atlas replica set used by the integration and your troubles will be over (for this test).

evergreen.yml
nbagnard: suggestion:
I'd keep this as is.
The idea is that these tests are checking that the path using the embedded translation engine works.

evergreen.yml
```diff
@@ -1934,10 +1982,10 @@ tasks:
         variants: [windows-64, ubuntu2204, ubuntu2204-arm64]
       - func: "run ubuntu integration tests"
         variants: [ubuntu2204, ubuntu2204-arm64]
+      - func: "run ubuntu on prem integration tests"
+        variants: [ ubuntu2204, ubuntu2204-arm64 ]
       - func: "run ubuntu cluster type integration tests"
```
nbagnard, line 1940: todo:
Let's extend what we test here.
Alexi has all the clusters we need:
- Cluster0: Good candidate for enabling the SQL interface (since is our default cluster for integration tests already)
- Cluster1: Good candidate for a disabled SQL interface (checking explicit "disabled" marker)
- Cluster2: Good candidate for a never enabled SQL interface
- TestFree: Possible candidate to test that non dedicated cluster behave as expected too. Might not be required, because this is more or less the same than Cluster2 from the client point-of-view

core/src/entitlement.rs
nbagnard: Agreed, let's describe the behavior with high level description only.

core/src/entitlement.rs
nbagnard: You can remove the comment about non implemented behavior entirely for now.

core/src/entitlement.rs
```diff
@@ -0,0 +1,614 @@
...
+const MARKER_COLLECTION: &str = "__sql_status";
+const MARKER_ID: &str = "entitlement";
+
+// Atlas dedicated cluster hostnames always embed the cluster name before the first `-shard-`
+// segment and resolve under `.mongodb.net`, e.g.
+//   cluster0-shard-00-00.abc123.mongodb.net:27017        (replica set node)
+//   cluster0-shard-00-mongos-g0.abc123.mongodb.net:27017 (mongos)
+// The cluster name is the prefix before `-shard-`. On-prem / self-managed hosts do not match,
+// which is exactly how we scope the entitlement gate to Atlas dedicated clusters only.
```
nbagnard: nit:
Remove this. It doesn't provide any value.

core/src/entitlement.rs
nbagnard, line 1: todo:
Instead of re-writing the logic, let's use the implementation in `sql_interface_status.rs` from schema-manager and tweak the few small diff.
The closer both impl are, the easier it is to fix bugs.
I'll have a PR opened soon in schema manager though, because I noticed that you correctly only check Atlas clusters, while on schema manager, Jon's logic was also being applied to self hosted clusters.

core/src/odbc_uri.rs
```diff
@@ -33,6 +33,10 @@ pub const APPNAME: &str = "appname";
 pub const LOGLEVEL: &str = "loglevel";
 pub const SIMPLE_TYPES_ONLY: &str = "simple_types_only";
 pub const ENABLE_MAX_STRING_LENGTH: &str = "enable_max_string_length";
+// SQL-3340 / task 28: optional path to a local JWKS (jwks.json) used to verify the Atlas SQL
+// interface entitlement marker in air-gapped environments where the well-known JWKS endpoint is
+// unreachable. Consumed by entitlement::JwksKeyProvider once key retrieval (task 28) is wired in.
+pub const SQL_PUBLIC_KEY_PATH: &str = "sqlpublickeypath";
```
nbagnard: todo:
Remove for now

core/Cargo.toml
```diff
@@ -45,6 +45,11 @@ rfc8252_http_server = { path = "../rfc8252_http_server" }
 once_cell = { workspace = true }
 serde_json = { workspace = true }
 rand = "0.10"
+# SQL-3340: Atlas SQL Direct Cluster entitlement marker (Ed25519-signed JWS) verification.
+# The JWKS publishes raw OKP public keys, so we verify the compact JWS by hand with
+# ed25519-dalek rather than pulling in a full JWT library that expects DER/PEM keys.
+ed25519-dalek = "2"
```
nbagnard: todo:
Remove for now

core/Cargo.toml
```diff
@@ -45,6 +45,11 @@ rfc8252_http_server = { path = "../rfc8252_http_server" }
...
+base64 = "0.22"
```
nbagnard: todo:
Let's use the same dependency.
Schema manager is using `data-encoding`.

evergreen/create-expansions.sh
nbagnard, line 1: todo:
Let's try to remove this changes and enable the SQL interface for the cluster mapped to SRV_TEST_HOST.

nbagnard (changes requested):
The logic is good 👍 
And reviewing this, I noticed that we missed excluding self hosted clusters in schema manager, which would have been very problematic if released 😱 Thank you!!! 🙏
My comments are all mostly just related to aim for a closer match with schema manager and revert how testing was organized (keep direct cluster in mongosqltranslate_test.rs and different cluster /connection types in cluster_type_test.rs).

mattChiaravalloti:
In our other repos, we have a section at the top of the evergreen config to [ignore](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#project-level-file-ignoring) changes that only include markdown files ([mongosql](https://github.com/mongodb/mongosql/blob/bd8f89c8a50004ed32fd8894e29a9cab222f2689/evergreen.yml#L16-L18), [schema-manager-rs](https://github.com/10gen/schema-manager-rs/blob/dfdd6cafb7f8361774a387addd60412fd06448f3/evergreen.yml#L16-L18), [mongo-jdbc-driver](https://github.com/mongodb/mongo-jdbc-driver/blob/e36976018a193f59791b70ecaa55a0af0cb40453/.evg.yml#L4-L5), [sql-engines-common-test-infra](https://github.com/mongodb/sql-engines-common-test-infra/blob/28b8616fc95a2a5a614ad3b5e0dbf94e763a18d5/evergreen.yml#L15-L17), and soon [sqlproxy](https://github.com/10gen/sqlproxy/pull/572)). This repository is our last one* that does not include this filter. Would you mind adding it as a drive-by change in this PR since this seems to be the next most likely PR to merge into this repo?

See the links for examples of exactly what to include!

 (*) I said "last one" but actually the Tableau and PowerBI connector repos do not do this either, but those are not quite as relevant.

core/src/conn.rs
```diff
@@ -17,10 +17,14 @@ use tokio::runtime::Runtime;
 // we make from UserOptions to Client and Weak<Runtime> so that we do not hold around
 // Clients and Runtimes that are no longer in use. In most cases it won't matter, but for drivers that live in
 // memory for a long time, this could be a problem.
+// Each entry also caches the Atlas dedicated cluster name derived from `hello.me` when the Client
+// was created (SQL-3340). Resolving it is part of basic mongo connection setup, so it is done once
+// per Client rather than on every logical connection; the SQL interface status gate that consumes
+// it still runs per logical connection. `None` means the host is not an Atlas dedicated cluster.
```
nicholascioli, line 23: Nit: This is just a normal comment so it won't show up on hover for this struct type. Could we convert this to a docstring (e.g. by using the `///` for the comment)?

nicholascioli: Oh, I see that the reason for this is probably because of the distinction between `Weak` and `Arc` causing there to be duplicate types that would require different doc strings. 

A suggestion would be to either duplicate the docstring per type (with the comment about `Weak` being only on the non-garbage collected variant) or to make a new struct for the inner tuple type and then gate the specific member behind the feature flag.

Up to you as to which approach you take, but feel free to just file a ticket to fix this later so as to keep the scope of this PR constrained.

core/src/sql_interface_status.rs
```diff
@@ -0,0 +1,468 @@
...
+// segment and resolve under `.mongodb.net`, e.g.
+//   cluster0-shard-00-00.abc123.mongodb.net:27017        (replica set node)
+//   cluster0-shard-00-mongos-g0.abc123.mongodb.net:27017 (mongos)
+// The cluster name is the prefix before `-shard-`. On-prem / self-managed hosts do not match,
+// which is exactly how we scope the status gate to Atlas dedicated clusters only. We require the
+// `.mongodb.net` suffix rather than just looking for `-shard-`, so that a self-managed host that
+// happens to be named `<something>-shard-0.<domain>` is not mistaken for an Atlas cluster and
+// wrongly gated.
+static ATLAS_DEDICATED_HOST: LazyLock<regex::Regex> =
+    LazyLock::new(|| regex::Regex::new(r"^([^.]+)-shard-\d+.*\.mongodb\.net(:\d+)?$").unwrap());
```
nicholascioli: Nit (non-blocking): Please use expect here with a message so that if this pattern ever breaks it'll be easy to see why / where.


```suggestion
    LazyLock::new(|| regex::Regex::new(r"^([^.]+)-shard-\d+.*\.mongodb\.net(:\d+)?$").expect("atlas dedicated host regex pattern should have parsed"));
```

core/src/sql_interface_status.rs
```diff
@@ -0,0 +1,468 @@
...
+/// Extracts the Atlas dedicated cluster name from a `hello.me` value, if the host is an Atlas
+/// dedicated cluster. Returns `None` for any other host (on-prem, self-managed, ADF, …), which
+/// signals that the status gate does not apply.
+///
+/// The name is lowercased to match the producer's convention for the marker's `sub` claim. Atlas
+/// hostnames are already lowercase; this is defensive normalization.
+pub fn atlas_dedicated_cluster_name(me: &str) -> Option<String> {
+    ATLAS_DEDICATED_HOST
+        .captures(me)
+        .map(|caps| caps[1].to_lowercase())
```
nicholascioli: Suggestion (non-blocking): The `caps[1]` here can panic if the regex ever changes to not have a capture group. It might be worth using `get` followed by `expect` here so that we catch this if it happens in the wild.


```suggestion
        .map(|caps| caps.get(1).expect("matched atlas dedicated host should have had a capture group").to_lowercase())
```

nicholascioli: Another option would be to name the capture group and try to grab it that way.


```suggestion
        .map(|caps| caps.name("cluster").expect("matched atlas dedicated host should have had a named capture group for the cluster").to_lowercase())
```

core/src/sql_interface_status.rs
```diff
@@ -0,0 +1,468 @@
...
+
+/// Reads the marker token from `__sql_status` using a Primary read, so the gate never observes a
+/// stale replica and always sees the latest toggle state. A read failure is
+/// [`Error::SqlInterfaceStatusReadFailed`]; a missing document is
+/// [`Error::SqlInterfaceUnavailable`]; a present document with an empty or non-string token is
+/// [`Error::SqlInterfaceStatusInvalid`].
+async fn read_marker_token(client: &Client) -> Result<String> {
+    let marker = client
+        .database(MARKER_DB)
+        .collection::<Document>(MARKER_COLLECTION)
```
nicholascioli, line 130: Suggestion (non-blocking): Might be worth having a deserializable struct for the collection instead of the generic `Document`. Then it would handle ensuring that `token` is present and a string.

core/src/sql_interface_status.rs
```diff
@@ -0,0 +1,468 @@
...
+            log::warn!("SQL interface status marker has no readable token field");
+            Err(Error::SqlInterfaceStatusInvalid)
+        }
+    }
+}
+
+/// Extracts the status-bearing claims from a compact JWS by base64url-decoding its payload (the
+/// second segment), per RFC 4648 §5 (no padding). The signature is neither required nor validated,
+/// per the status-only gate.
+fn decode_marker_claims(token: &str) -> Option<MarkerClaims> {
```
nicholascioli, line 166: Suggestion (non-blocking): While this works, it might be worth using a dedicated library for raw JWT logic, since it can be easy to get wrong.

core/src/sql_interface_status.rs
```diff
@@ -0,0 +1,468 @@
...
+        Some(issuer @ (ISSUER_NORMAL | ISSUER_EMERGENCY)) => issuer,
+        other => {
+            log::warn!("unrecognized SQL interface marker issuer {other:?}");
+            return Err(Error::SqlInterfaceStatusInvalid);
+        }
+    };
+
+    // `sub` binds the marker to this cluster, preventing a valid marker from being copied from
+    // another cluster.
+    if claims.sub.as_deref() != Some(cluster_name) {
```
nicholascioli, line 201: Nit (non-blocking): It might be worth always doing a case insensitive comparison here so as to not have to re-release the clients if the way the marker is generated introduces either non-lowercasable letters (in the event of UTF-8) or uppercase letters.

core/src/sql_interface_status.rs
```diff
@@ -0,0 +1,468 @@
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use serde_json::json;
+
+    /// Builds a compact JWS with an arbitrary header/signature and the given claims as the
+    /// payload. The header alg and the signature are deliberately junk: the status-only gate must
+    /// never look at either.
+    fn token_with_claims(claims: &serde_json::Value) -> String {
+        let header = URL_SAFE_NO_PAD.encode(br#"{"alg":"Ed25519","typ":"JWT"}"#);
```
nicholascioli: Question (non-blocking): I'm not sure exactly how the markers are generated, but the header might actually be something like `{"alg": "EdDSA", "crv": "Ed25519", ...}`.

I don't think this affects your tests, but it will whenever we enable signature fingerprinting.

core/src/sql_interface_status.rs
```diff
@@ -0,0 +1,468 @@
...
+    /// Epoch seconds offset from now, for exercising emergency-marker expiry.
+    fn epoch_from_now(offset_secs: i64) -> i64 {
+        now_epoch_seconds().expect("system clock is after the Unix epoch") + offset_secs
+    }
+
+    #[test]
+    fn decode_marker_claims_decodes_external_reference_payload() {
+        // The payload segment was produced independently by Python's base64.urlsafe_b64encode
+        // (padding stripped), proving we decode a real externally-encoded token, not just our own
+        // encoder. Header/signature segments are irrelevant to the status-only gate.
```
nicholascioli: Nit (non-blocking): Could you paste what that payload looks like here for reference?

core/src/sql_interface_status.rs
```diff
@@ -0,0 +1,468 @@
...
+        // A middle segment that decodes but is not a JSON object.
+        assert!(
+            decode_marker_claims(&format!("aaa.{}.bbb", URL_SAFE_NO_PAD.encode(b"{"))).is_none()
+        );
+    }
+
+    #[test]
+    fn no_signature_validation_is_attempted() {
+        // The signature segment is garbage and the header alg is non-standard, yet an enabled
+        // marker with a matching issuer and sub still passes: the gate never inspects either.
```
nicholascioli: Question: Do we want this to be the case? While we might not care right now that the signature is by us, shouldn't we still care that it is a _valid_ signature that matches the header + payload?

evergreen/set_up_on_prem_int_tests.sh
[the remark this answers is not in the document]
nicholascioli: Doesn't the user need to be able to read the `__mdb_internal_sqlinterface:__sql_status` collection in order to fetch the marker?

nicholascioli (approved):
Looks good! Left a few nits / questions.

integration_test/tests/on_prem_tests.rs
EthanHardyMongo: > Did you update the tests to use an on-prem deployment is because you couldn't connect to Alexi's cluster anymore?

Yeah. 

> If so, I think we can either enable it if Alexi is Ok with it or "manually" add a marker for it ourselves (to avoid billing cost).

Yeah that works!

evergreen/set_up_on_prem_int_tests.sh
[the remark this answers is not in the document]
EthanHardyMongo: There's no marker for on-prem, but I'm reverting this change anyway.

core/src/sql_interface_status.rs
EthanHardyMongo: That's a good point, but this follows what was done in the schema manager, so I guess we don't care about that for now.

core/src/sql_interface_status.rs
EthanHardyMongo: It was also done this way in the schema manager, so I think we do want to be case sensitive here.

core/src/sql_interface_status.rs
EthanHardyMongo: I think we should add a dedicated library when we add fingerprinting, but since we're not doing the hard parts of the JWT logic yet (we're just base64url-decoding the payload segment and reading three claims), I think this is fine for now. This also matches what the schema manager does.

core/src/sql_interface_status.rs
EthanHardyMongo: I added a `MarkerDoc` struct but kept find_one returning a Document and deserialized into a `MarkerDoc` struct afterwards. Typing the collection directly would run serde inside find_one, so a malformed marker would come back as a `mongodb::error::Error` and get reported as a read failure rather than an invalid marker, and we'd lose the distinction between "couldn't reach the DB", "no marker", and "bad marker". This way we get the type-checking and keep the error mapping.

core/src/sql_interface_status.rs
EthanHardyMongo: good call! I added this: 
```
      ATLAS_DEDICATED_HOST.captures(me).map(|caps| {                                                                                                                                                             
          // A missing group here would mean the pattern above lost its capture group, not that this                                                                                                             
          // host is non-Atlas. Panicking is deliberate: it is caught at the ODBC boundary and fails                                                                                                             
          // the connection, whereas returning `None` would be read as "not an Atlas cluster" and                                                                                                                
          // would silently disable the gate for every Atlas cluster.                                                                                                                                            
          caps.get(1)                                                                                                                                                                                            
              .expect("a matched Atlas dedicated host always has a cluster-name capture group")                                                                                                                  
              .as_str()                                                                                                                                                                                          
              .to_lowercase()                                                                                                                                                                                    
      })  
```

core/src/conn.rs
EthanHardyMongo: Yeah I'll file a ticket for this.

core/src/sql_interface_status.rs
EthanHardyMongo: I did make it a little stricter though: now the token must have exactly three segments.

core/src/sql_interface_status.rs
EthanHardyMongo: Yeah, I added it.

core/src/sql_interface_status.rs
EthanHardyMongo: I changed it so that the header now mirrors what a real marker carries.

integration_test/tests/direct_cluster_tests.rs
```diff
@@ -1,10 +1,9 @@
 mod common;
 
-mod mongosqltranslate_tests {
+mod direct_cluster_tests {
     use crate::common::{
```
EthanHardyMongo, line 4: I renamed this test file and the tests to make it clearer what this is. However, I can change it back if that's preferable. I wanted to change it since we have a mongosqltranslate library that's no longer used in the ODBC driver.

integration_test/tests/connection_tests.rs
```diff
@@ -44,6 +44,85 @@ mod integration {
...
     }
 
+    // Connects to a real Atlas cluster that has the SQL interface enabled, so the status marker is
+    // present and reports `enabled: true`. This exercises the SQL interface status gate against a
+    // live marker rather than a fixture. It doubles as coverage that SRV style URI connection
+    // strings connect successfully.
+    #[test]
+    fn test_atlas_cluster_with_sql_interface_enabled_succeeds() {
+        let env_handle = allocate_env(AttrOdbcVersion::SQL_OV_ODBC3);
+        let conn_str = crate::common::generate_srv_style_connection_string("test", "cluster0");
```
EthanHardyMongo, line 54: The atlas cluster type tests using `cluster0` and `cluster1` won't work until we have them set up with the SQL interface. Additionally, the direct cluster tests won't work until `cluster0` is set up.

integration_test/tests/on_prem_cluster_type_test.rs
```diff
@@ -1,7 +1,7 @@
 mod common;
 
-#[cfg(feature = "cluster_type_tests")]
-mod cluster_type {
+#[cfg(feature = "on_prem_cluster_type_tests")]
+mod on_prem_cluster_type {
```
EthanHardyMongo, line 4: I renamed this file because it is only for testing on-prem cluster types because they have to be handled differently in our testing. I put the atlas cluster type tests in `connection_tests.rs` because they can be run like any other connection test.

integration_test/tests/common/mod.rs
```diff
@@ -101,22 +101,17 @@ pub fn generate_connection_str(user: Option<String>, database: Option<String>) -
...
-        env::var("SRV_TEST_DB").expect("SRV_TEST_DB is not set")
-    };
-
     let auth_db = env::var("SRV_TEST_AUTH_DB").expect("SRV_TEST_AUTH_DB is not set");
     let host = env::var("SRV_TEST_HOST").expect("SRV_TEST_HOST is not set");
     let username = env::var("SRV_TEST_USER").expect("SRV_TEST_USER is not set");
     let password = env::var("SRV_TEST_PWD").expect("SRV_TEST_PWD is not set");
 
-    let mongodb_uri = format!("mongodb+srv://{username}:{password}@{host}/?authSource={auth_db}");
+    let mongodb_uri =
```
nbagnard, line 113: issue:
SRV_TEST_HOST contains the cluster name right now in our secret manager

EthanHardyMongo: I updated the evergreen `srv_test_host` variable to `rotpc.mongodb.net`.

nbagnard (approved):
LGTM

jchemburkar (approved):

nbagnard:
@EthanHardyMongo The Macos variant is still using MacOS 13. I thought we did a round to fix this on all our projects, but it looks like we missed this one. It's not running on the waterfall and we don't release a mac driver, so I am changing the patch and commit check to allow the PR to go through.

nbagnard:
evergreen retry

